PR #16, opened 2015-01-16
Change platform check to /linux/

### LINKED ISSUE #15: Platform check not sufficiently permissive
@Sirupsen @csfrancis I was trying to use semian on an x64 Ubuntu 14.10 installation and I got `Semian is not supported on x86_64-linux-gnu - all operations will no-op`

The current check is just `end_with?('-linux')` which clearly isn't sufficient, though I don't know enough about the format of `RUBY_PLATFORM` to say what else is needed.

csfrancis:
@Sirupsen, @eapache 

Fixes #15.

### DIFF
--- a/lib/semian/platform.rb
+++ b/lib/semian/platform.rb
@@ -1,6 +1,6 @@
 class Semian
   # Determines if Semian supported on the current platform.
   def self.supported_platform?
-    RUBY_PLATFORM.end_with?('-linux')
+    /linux/.matches(RUBY_PLATFORM)
   end
 end
--- a/lib/semian/platform.rb
+++ b/lib/semian/platform.rb
@@ -1,6 +1,6 @@
 class Semian
   # Determines if Semian supported on the current platform.
   def self.supported_platform?
-    /linux/.matches(RUBY_PLATFORM)
+    /linux/.match(RUBY_PLATFORM)
   end
 end


eapache:
:+1:

csfrancis:
Travis seems to be :sleeping:

eapache:
I have noticed that on Sarama too - they must be very under-provisioned for free accounts

Sirupsen:
We actually have a paid account, don't know how we can attach this to it @j-mutter ?

j-mutter:
No, it's :hankey: - only private repos get access to our paid account :(

csfrancis:
:poop:

Sirupsen:
@joshk :(

joshk:
what have I done?

Sirupsen:
@joshk is it possible to attach public repositories to a paid account to get prioritized builds?

joshk:
Not at the moment, we are working on that (merging .org and .com together)




That said, using the Docker setup means we have uber capacity. Any delays in builds today would be due to a queuing component failing which we recently fixed :)

Sirupsen:
We're actually [on that](https://github.com/Shopify/semian/blob/master/.travis.yml#L3), you changed it at DockerCon :-)

joshk:
Yep, I saw, thats why I was telling you about the downtime, because there should be no need for prioritised builds as we have ALLLL the capacity :)

PR #18, opened 2015-02-06
Wrap semian resources with a circuit breaker

byroot:
@csfrancis @Sirupsen @fw42 for review please.

(It is missing documentation, but I prefer to do it once we settle on the API etc.)

### DIFF
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -73,15 +73,14 @@ module Semian
   #
   # Returns the registered resource.
   def register(name, tickets: 0, permissions: 0660, timeout: 1, success_threshold: 1, error_threshold: 5, error_timeout: 20, exceptions: [])
-    resource = Resource.new(name, tickets, permissions, timeout)
     circuit_breaker = CircuitBreaker.new(
-      resource,
       success_threshold: success_threshold,
       error_threshold: error_threshold,
       error_timeout: error_timeout,
-      exceptions: exceptions,
+      exceptions: Array(exceptions) + [::Semian::BaseError],
     )
-    resources[name] = circuit_breaker
+    resource = Resource.new(name, tickets, permissions, timeout)
+    resources[name] = ProtectedResource.new(resource, circuit_breaker)
   end
 
   # Retrieves a resource by name.
@@ -97,6 +96,7 @@ end
 
 require 'semian/resource'
 require 'semian/circuit_breaker'
+require 'semian/protected_resource'
 require 'semian/platform'
 if Semian.supported_platform?
   require 'semian/semian'
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -1,3 +1,4 @@
+require 'logger'
 #
 # === Overview
 #
@@ -52,38 +53,50 @@
 #
 # This is the same as the previous example, but overrides the timeout from the default value of 500 milliseconds to 1 second.
 #
-class Semian
-  class << self
-    # Registers a resource.
-    #
-    # +name+: Name of the resource - this can be either a string or symbol.
-    #
-    # +tickets+: Number of tickets. If this value is 0, the ticket count will not be set,
-    # but the resource must have been previously registered otherwise an error will be raised.
-    #
-    # +permissions+: Octal permissions of the resource.
-    #
-    # +timeout+: Default timeout in seconds.
-    #
-    # Returns the registered resource.
-    def register(name, tickets: 0, permissions: 0660, timeout: 1)
-      resource = Resource.new(name, tickets, permissions, timeout)
-      resources[name] = resource
-    end
+module Semian
+  extend self
 
-    # Retrieves a resource by name.
-    def [](name)
-      resources[name]
-    end
+  attr_accessor :logger
 
-    # Retrieves a hash of all registered resources.
-    def resources
-      @resources ||= {}
-    end
+  self.logger = Logger.new(nil)
+
+  # Registers a resource.
+  #
+  # +name+: Name of the resource - this can be either a string or symbol.
+  #
+  # +tickets+: Number of tickets. If this value is 0, the ticket count will not be set,
+  # but the resource must have been previously registered otherwise an error will be raised.
+  #
+  # +permissions+: Octal permissions of the resource.
+  #
+  # +timeout+: Default timeout in seconds.
+  #
+  # Returns the registered resource.
+  def register(name, tickets: 0, permissions: 0660, timeout: 1, success_threshold: 1, error_threshold: 5, error_timeout: 20, exceptions: [])
+    resource = Resource.new(name, tickets, permissions, timeout)
+    circuit_breaker = CircuitBreaker.new(
+      resource,
+      success_threshold: success_threshold,
+      error_threshold: error_threshold,
+      error_timeout: error_timeout,
+      exceptions: exceptions,
+    )
+    resources[name] = circuit_breaker
+  end
+
+  # Retrieves a resource by name.
+  def [](name)
+    resources[name]
+  end
+
+  # Retrieves a hash of all registered resources.
+  def resources
+    @resources ||= {}
   end
 end
 
 require 'semian/resource'
+require 'semian/circuit_breaker'
 require 'semian/platform'
 if Semian.supported_platform?
   require 'semian/semian'
--- a/lib/semian/circuit_breaker.rb
+++ b/lib/semian/circuit_breaker.rb
@@ -1,16 +1,12 @@
 module Semian
   class CircuitBreaker
-    extend Forwardable
     attr_reader :state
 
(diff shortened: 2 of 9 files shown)


lib/semian.rb
```diff
@@ -52,38 +53,52 @@
...
-    # +permissions+: Octal permissions of the resource.
-    #
-    # +timeout+: Default timeout in seconds.
-    #
-    # Returns the registered resource.
-    def register(name, tickets: 0, permissions: 0660, timeout: 1)
-      resource = Resource.new(name, tickets, permissions, timeout)
-      resources[name] = resource
-    end
+    attr_accessor :logger
```
Sirupsen: Why both `extend self` and `class << self` ?

lib/semian.rb
```diff
@@ -52,38 +53,52 @@
...
+  end
+
+  self.logger = Logger.new(nil)
 
-    # Retrieves a resource by name.
-    def [](name)
-      resources[name]
-    end
+  # Registers a resource.
```
Sirupsen: This is a little vague since we both create a Semian resource and Circuit Breaker

lib/semian.rb
```diff
@@ -52,38 +53,52 @@
...
+  #
+  # +tickets+: Number of tickets. If this value is 0, the ticket count will not be set,
+  # but the resource must have been previously registered otherwise an error will be raised.
+  #
+  # +permissions+: Octal permissions of the resource.
+  #
+  # +timeout+: Default timeout in seconds.
+  #
+  # Returns the registered resource.
+  def register(name, tickets: 0, permissions: 0660, timeout: 1, success_threshold: 1, error_threshold: 5, error_timeout: 20, exceptions: [])
```
Sirupsen: What's `error_timeout`? It's not documented.

The thresholds and exceptions aren't documented either.

Sirupsen: Maybe call it `circuit_timeout`?

lib/semian/circuit_breaker.rb
```diff
@@ -0,0 +1,131 @@
+module Semian
+  class CircuitBreaker
```
Sirupsen: You should add some documentation to this class.

lib/semian/circuit_breaker.rb
```diff
@@ -0,0 +1,131 @@
...
+      @error_last_at && (@error_last_at + @error_timeout < Time.now)
+    end
+
+    def log_state_transition(new_state)
+      return if @state.nil? || new_state == @state
+
+      str = "[#{self.class.name}] State transition from #{@state} to #{new_state}."
+      str << " success_count=#{@success_count} error_count=#{@error_count}"
+      str << " success_count_threshold=#{@success_count_threshold} error_count_threshold=#{@error_count_threshold}"
+      str << " error_timeout=#{@error_timeout} error_last_at=\"#{@error_last_at}\""
```
Sirupsen: Could we have a helper that converts a hash to nice fields like this?

test/test_circuit_breaker.rb
```diff
@@ -0,0 +1,101 @@
+require 'test/unit'
+require 'semian'
+require 'timecop'
+
+class TestSemian < Test::Unit::TestCase
```
Sirupsen: `TestCircuitBreaker`

Sirupsen:
Couple of comments but nothing really stands out. :heart: that you're doing this @byroot

lib/semian.rb
byroot: Hum, right. Brain fart I guess.

lib/semian.rb
byroot: Again. I think the documentation will have to fix totally revamped. But I prefer to do so once all the API concerns have been decided.

test/test_circuit_breaker.rb
byroot: oops

lib/semian/circuit_breaker.rb
byroot: So I just tried in local:

```ruby
    def log_state_transition(new_state)
      return if @state.nil? || new_state == @state

      Semian.logger.info "[#{self.class.name}] State transition from #{@state} to #{new_state}. " << log_variables(
        :success_count,
        :error_count,
        :success_count_threshold,
        :error_count_threshold,
        :error_timeout,
        :error_last_at,
      )
    end

    def log_variables(*variables)
      variables.map do |variable_name|
        value = instance_variable_get("@#{variable_name}")
        "#{v}=#{value.inspect}"
      end.join(' ')
    end
```

Not sure if it's really better...

Sirupsen: Not a big deal. If we see this pattern again we can look into it.

byroot:
Another thing I forgot to mention. The circuit breaker in his current form is not thread safe at all. So it's not suitable for people using puma or Sidekia for example.

Sirupsen:
That's cool @byroot, I don't think we need to worry about that for now as long as we document it.

Thread safe code is slow code... and it's definitely more complicated code

csfrancis:
Like you said in the PR description, we'll have to update the docs once this ships, but overall it looks good to me. :+1:

lib/semian/circuit_breaker.rb
```diff
@@ -0,0 +1,131 @@
+module Semian
+  class CircuitBreaker
+    extend Forwardable
+    attr_reader :state
+
+    def_delegators :@resource, :acquire, :destroy, :count, :semid
```
fw42: This stuff seems specific to semaphores. I don't feel like it belongs in this class. After all we have lots of use-cases for CircuitBreaker that have nothing to do with semaphores.

lib/semian/circuit_breaker.rb
```diff
@@ -0,0 +1,131 @@
...
+
+    def initialize(resource, exceptions:, success_threshold:, error_threshold:, error_timeout:)
+      @resource = resource
+      @success_count_threshold = success_threshold
+      @error_count_threshold = error_threshold
+      @error_timeout = error_timeout
+      @exceptions = Array(exceptions) + [::Semian::BaseError]
```
fw42: same here. Should the Semian::BaseError really be in here?

lib/semian/circuit_breaker.rb
fw42: how are we going to plug statsd into this?

lib/semian/circuit_breaker.rb
```diff
@@ -0,0 +1,131 @@
...
+      reset
+    end
+
+    def with_fallback(fallback, &block)
```
fw42: I know we talked about it in person, but now that I see it, I think the class design was cleaner the way it was before (in the Shopify codebase) with the Resiliency::Runner. A CircuitBreaker is a concept that in itself doesn't (shouldn't) know anything about "running code" or "fallbacks" or "exceptions". This class should encapsulate the internal state of the CircuitBreaker and nothing else, imho.

lib/semian/circuit_breaker.rb
byroot: I'll add statsd support in a followup. I'm not totally sure how I'll do it yet, but I'll try to make it non obtrusive.

lib/semian/circuit_breaker.rb
byroot: I see you point. I did this to not change the current Semian API, but since I bout many people are using it yet, we can find a better way. Let's chat a bit.

byroot:
@fw42 I just refactored the code, `Semian::CircuitBreaker` now doesn't know anything about resources, and can be used as a standalone.

fw42:
Cool, this looks a bit better to me. Not sure about the class name, but I will let @Sirupsen and @csfrancis fight over that :-) I still kinda think that the "running code with a circuit breaker" part (i.e. the exceptions and the with_fallback stuff) should also be somewhere else. But no huge deal for now.

byroot:
> Not sure about the class name

My reasoning was to continue with the circuit breaker methapor: 

>> A circuit breaker is an automatically operated electrical switch designed to protect an electrical circuit from damage caused by overload or short circuit

So a resource wrapped with a circuit breaker is a "ProtectedResource" `¯\_(ツ)_/¯`

I'm totally open to other names though.

> I still kinda think that the "running code with a circuit breaker" part [...] should also be somewhere else

You mean it should be handled by the Redis/MySQL/Whatever client?

My idea here is that the acquire or with_fallback shouldn't really be used directly, but since most clients will just have to specify a ticket count an a list of exceptions we might as well factorize it.

But I guess we can still change that if it do not feel right when I start to implement the clients wrappers.

@Sirupsen @csfrancis still good for you?

Sirupsen:
@byroot LGTM.

Sirupsen:
And I agree let's wait with documentation

PR #20, opened 2015-02-09
Initial attempt at Semian::Mysql2

byroot:
For the fun I integrated it in Shipit: https://github.com/Shopify/shipit2/commit/10ee2cc8d0eec39f7c882b47b6e8693adcac998c

If must be manually enabled via:

```ruby
Semian::Mysql2.enable!([optional global settings])
```

And then individual connections can have their settings defined when the client is instanciated:

```ruby
Mysql2::Client.new(semian: {timeout: 42, ...})
```

Which mean that for rails app, it can be done from the `database.yml`

```ruby
production:
  ...
  semian:
    timeout: 42
```

A couple things:

  - Again no more StatsD. I believe it's the integration application responsibility. However I'll see if Semian can provide a couple hooks to make it easier.
  - I simply ported our custom code, but I think `connect` and `query` should be using the circuit breaker. The fallback would be to raise a `Mysql2::SemianError`.

@fw42 @Sirupsen @csfrancis thoughts? (I know it's an holiday for you so feel free to pospone the review until tomorrow)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,5 +2,10 @@ language: ruby
 
 sudo: false
 
+before_install:
+  - wget -O toxiproxy-1.0.0.deb https://github.com/Shopify/toxiproxy/releases/download/v1.0.0/toxiproxy_1.0.0_amd64.deb
+  - sudo dpkg -i toxiproxy-1.0.0.deb
+  - sudo service start toxiproxy
+
 rvm:
   - 2.1.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,9 @@
 language: ruby
 
-sudo: false
+sudo: true
 
 before_install:
-  - wget -O toxiproxy-1.0.0.deb https://github.com/Shopify/toxiproxy/releases/download/v1.0.0/toxiproxy_1.0.0_amd64.deb
-  - sudo dpkg -i toxiproxy-1.0.0.deb
-  - sudo service start toxiproxy
+  - scripts/install_toxiproxy.sh
 
 rvm:
   - 2.1.1
--- a/Rakefile
+++ b/Rakefile
@@ -29,6 +29,11 @@ else
   task :build do; end
 end
 
+task :populate_proxy do
+  require 'toxiproxy'
+  Toxiproxy.populate(File.expand_path('../test/fixtures/toxiproxy.json', __FILE__))
+end
+
 # ==========================================================
 # Testing
 # ==========================================================
@@ -42,7 +47,7 @@ Rake::TestTask.new 'test' do |t|
     t.test_files = all_files - FileList['test/test_semian.rb']
   end
 end
-task :test => :build
+task :test => [:build, :populate_proxy]
 
 # ==========================================================
 # Documentation
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -56,6 +56,11 @@ require 'logger'
 module Semian
   extend self
 
+  BaseError = Class.new(StandardError)
+  SyscallError = Class.new(BaseError)
+  TimeoutError = Class.new(BaseError)
+  InternalError = Class.new(BaseError)
+
   attr_accessor :logger
 
   self.logger = Logger.new(nil)
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -105,5 +105,3 @@ else
   $stderr.puts "Semian is not supported on #{RUBY_PLATFORM} - all operations will no-op"
 end
 require 'semian/version'
-
-require 'semian/mysql2' if defined?(Mysql2)
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -105,3 +105,5 @@ else
   $stderr.puts "Semian is not supported on #{RUBY_PLATFORM} - all operations will no-op"
 end
 require 'semian/version'
+
+require 'semian/mysql2' if defined?(Mysql2)
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -83,9 +83,19 @@ module Semian
     resources[name] = ProtectedResource.new(resource, circuit_breaker)
   end
 
+  def retrieve_or_register(name, **args)
+    self[name] || register(name, **args)
+  end
+
   # Retrieves a resource by name.
   def [](name)
     resources[name]
+  end
+
+  def destroy(name)
+    if resource = resources.delete(name)
+      resource.destroy
+    end
   end
 
   # Retrieves a hash of all registered resources.
--- a/lib/semian/mysql2.rb
+++ b/lib/semian/mysql2.rb
@@ -37,7 +37,7 @@ module Semian
     end
 
     def semian_resource
-      @semian_resource ||= ::Semian.register(semian_identifier, **semian_options)
+      @semian_resource ||= ::Semian.retrieve_or_register(semian_identifier, **semian_options)
     end
 
     def semian_options
--- a/lib/semian/mysql2.rb
+++ b/lib/semian/mysql2.rb
@@ -23,7 +23,7 @@ module Semian
     end
 
     def query(*)
-      semian_resource.acquire { super }
+      semian_resource.with_fallback(-> { raise ::Semian::BaseError }) { super }
     rescue ::Semian::BaseError => error
       raise ::Mysql2::SemianError.new(semian_identifier, error)
     end
@@ -31,7 +31,7 @@ module Semian
     private
 
     def connect(*)
-      semian_resource.acquire { super }
+      semian_resource.with_fallback(-> { raise ::Semian::BaseError }) { super }
     rescue Semian::BaseError => error
       raise ::Mysql2::SemianError.new(semian_identifier, error)
     end
--- a/lib/semian/mysql2.rb
+++ b/lib/semian/mysql2.rb
@@ -1,3 +1,6 @@
+require 'semian'
+require 'mysql2'
+
 class Mysql2::SemianError < Mysql2::Error
   def initialize(semian_identifier, *args)
     super(*args)
@@ -11,18 +14,9 @@ end
 
 module Semian
   module Mysql2
-    class << self
-      attr_accessor :default_options
-
-      def enable!(default_options = {})
-        self.default_options = default_options
-        ::Mysql2::Client.prepend(self)
-      end
-    end
-
     def semian_identifier
       @semian_identifier ||= begin
-        name = query_options[:semian_resource]
+        name = query_options[:semian] && query_options[:semian][:name]
         name ||= [query_options[:host] || 'localhost', query_options[:port] || 3306].join(':')
         :"mysql_#{name}"
       end
@@ -47,11 +41,12 @@ module Semian
     end
 
     def semian_options
-      options = ::Semian::Mysql2.default_options.dup
-      if query_options.key?(:semian)
-        options.merge!(query_options[:semian].map { |k, v| [k.to_sym, v] }.to_h)
-      end
+      options = query_options[:semian] || {}
+      options = options.map { |k, v| [k.to_sym, v] }.to_h
+      options.delete(:name)
       options
     end
   end
 end
+
+::Mysql2::Client.prepend(Semian::Mysql2)
--- a/lib/semian/mysql2.rb
+++ b/lib/semian/mysql2.rb
@@ -1,39 +1,54 @@
 require 'semian'
 require 'mysql2'
 
-class Mysql2::SemianError < Mysql2::Error
-  def initialize(semian_identifier, *args)
-    super(*args)
-    @semian_identifier = semian_identifier
+module Mysql2
+  class SemianError < Mysql2::Error
+    def initialize(semian_identifier, *args)
+      super(*args)
+      @semian_identifier = semian_identifier
+    end
+
+    def to_s
+      "[#{@semian_identifier}] #{super}"
+    end
   end
 
-  def to_s
-    "[#{@semian_identifier}] #{super}"
-  end
+  ResourceOccupiedError = Class.new(SemianError)
+  CircuitOpenError = Class.new(SemianError)
 end
 
 module Semian
   module Mysql2
+    DEFAULT_HOST = 'localhost'
+    DEFAULT_PORT = 3306
+
     def semian_identifier
       @semian_identifier ||= begin
-        name = query_options[:semian] && query_options[:semian][:name]
-        name ||= [query_options[:host] || 'localhost', query_options[:port] || 3306].join(':')
+        unless name = query_options[:semian] && query_options[:semian][:name]
+          host = query_options[:host] || DEFAULT_HOST
+          port = query_options[:port] || DEFAULT_PORT
+          name = "#{host}:#{port}"
+        end
         :"mysql_#{name}"
       end
     end
 
     def query(*)
-      semian_resource.with_fallback(-> { raise ::Semian::BaseError }) { super }
+      semian_resource.acquire { super }
+    rescue ::Semian::CircuitBreaker::OpenCircuitError => error
+      raise ::Mysql2::CircuitOpenError.new(semian_identifier, error)
     rescue ::Semian::BaseError => error
-      raise ::Mysql2::SemianError.new(semian_identifier, error)
+      raise ::Mysql2::ResourceOccupiedError.new(semian_identifier, error)
     end
 
     private
 
     def connect(*)
-      semian_resource.with_fallback(-> { raise ::Semian::BaseError }) { super }
-    rescue Semian::BaseError => error
-      raise ::Mysql2::SemianError.new(semian_identifier, error)
+      semian_resource.acquire { super }
+    rescue ::Semian::CircuitBreaker::OpenCircuitError => error
+      raise ::Mysql2::CircuitOpenError.new(semian_identifier, error)
+    rescue ::Semian::BaseError => error
+      raise ::Mysql2::ResourceOccupiedError.new(semian_identifier, error)
     end
 
     def semian_resource
--- a/lib/semian/mysql2.rb
+++ b/lib/semian/mysql2.rb
@@ -1,0 +1,57 @@
+class Mysql2::SemianError < Mysql2::Error
+  def initialize(semian_identifier, *args)
+    super(*args)
+    @semian_identifier = semian_identifier
+  end
+
+  def to_s
+    "[#{@semian_identifier}] #{super}"
+  end
+end
+
+module Semian
+  module Mysql2
+    class << self
+      attr_accessor :default_options
+
+      def enable!(default_options = {})
+        self.default_options = default_options
+        ::Mysql2::Client.prepend(self)
+      end
+    end
+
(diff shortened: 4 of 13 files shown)


fw42:
@byroot Holiday is next week :)

byroot:
Oh that explain why so many people are working during a holiday.

lib/semian/mysql2.rb
```diff
@@ -0,0 +1,57 @@
...
+      def enable!(default_options = {})
+        self.default_options = default_options
+        ::Mysql2::Client.prepend(self)
+      end
+    end
+
+    def semian_identifier
+      @semian_identifier ||= begin
+        name = query_options[:semian_resource]
+        name ||= [query_options[:host] || 'localhost', query_options[:port] || 3306].join(':')
```
fw42: kinda meh about hardcoding localhost and 3306 here :/

byroot: Well. That's MySQL defaults. They won't change any time soon.

lib/semian/mysql2.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    def semian_resource
+      @semian_resource ||= ::Semian.register(semian_identifier, **semian_options)
+    end
+
+    def semian_options
+      options = ::Semian::Mysql2.default_options.dup
+      if query_options.key?(:semian)
+        options.merge!(query_options[:semian].map { |k, v| [k.to_sym, v] }.to_h)
```
fw42: is this the non-AS version of symbolize_keys?

byroot: Pretty much yes.

fw42: How about this:
```ruby
query_options[:semian].each{ |k, v| options[k.to_sym] = v }
```
I think that's a bit simpler.

byroot: Sure.

lib/semian/mysql2.rb
```diff
@@ -0,0 +1,57 @@
...
+end
+
+module Semian
+  module Mysql2
+    class << self
+      attr_accessor :default_options
+
+      def enable!(default_options = {})
+        self.default_options = default_options
+        ::Mysql2::Client.prepend(self)
```
Sirupsen: :fr:

lib/semian/mysql2.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+    end
+
+    def query(*)
+      semian_resource.acquire { super }
+    rescue ::Semian::BaseError => error
+      raise ::Mysql2::SemianError.new(semian_identifier, error)
+    end
+
+    private
```
Sirupsen: This should be after `#connect`

lib/semian/mysql2.rb
```diff
@@ -0,0 +1,57 @@
...
+    rescue Semian::BaseError => error
+      raise ::Mysql2::SemianError.new(semian_identifier, error)
+    end
+
+    def semian_resource
+      @semian_resource ||= ::Semian.register(semian_identifier, **semian_options)
+    end
+
+    def semian_options
+      options = ::Semian::Mysql2.default_options.dup
```
Sirupsen: If the `default_options` are never set, what would happen?

lib/semian/mysql2.rb
byroot: I though so too, but no:

```ruby
>> require 'mysql2'
=> false
>> Mysql2::Client.new.connect
NoMethodError: private method `connect' called for #<Mysql2::Client:0x007f9d2501a7d8>
	from (irb):5
	from /Users/byroot/.rvm/rubies/ruby-2.1.3/bin/irb:11:in `<main>'
```

Sirupsen:
I'm not sure I like the `enable!`, I'd prefer if they just weren't `require`d by default, and then you `require` the drivers you want.

lib/semian/mysql2.rb
Sirupsen: Should we warn to the log when `:semian` isn't present?

byroot: Hum, if you don't set the global options, and don't set the `:semian` in the `database.yml` either, you'll end up with `0` tickets. Meaning it will raise.

I start to wonder if the global options actually make sense. I think I'll get rid of it, and warn is no options are present.

byroot:
I just figured out I totally forgot to write a single test :facepalm:

lib/semian/mysql2.rb
Sirupsen: Yeah, I don't like that too much. I'm :+1: being explicit and having an explicit raise as well.

byroot:
So:

  - `Semian::Mysql2.enable!` is now `require 'semian/mysql2'`
  - No more default config. It have to be passed to the constructor.
  - It now protect `query` and `connect` with `with_fallback`, not just `acquire`
  - I added a few tests.

@Sirupsen mind doing a second round of review?

lib/semian/mysql2.rb
```diff
@@ -0,0 +1,52 @@
...
+  def to_s
+    "[#{@semian_identifier}] #{super}"
+  end
+end
+
+module Semian
+  module Mysql2
+    def semian_identifier
+      @semian_identifier ||= begin
+        name = query_options[:semian] && query_options[:semian][:name]
```
Sirupsen: Raise an exception defined at the top level if the options aren't passed?

```
raise UnconfiguredAdapter, "Semian configuration options not passed to MySQL, see docs" unless ..
```

lib/semian/mysql2.rb
```diff
@@ -0,0 +1,52 @@
...
+        name ||= [query_options[:host] || 'localhost', query_options[:port] || 3306].join(':')
+        :"mysql_#{name}"
+      end
+    end
+
+    def query(*)
+      semian_resource.with_fallback(-> { raise ::Semian::BaseError }) { super }
```
Sirupsen: `with_fallback` because it has a circuit?

test/test_mysql2.rb
```diff
@@ -0,0 +1,39 @@
+require 'test/unit'
+require 'semian/mysql2'
+
+class TestMysql2 < Test::Unit::TestCase
```
Sirupsen: I'd like to see some more tests, particularly for:

* Toxiproxy slowness tests
* Circuit breaker
* Recovery after circuit break
* Possibly multiple processes

I know that this is painful. The majority of time developing this library will be testing.

lib/semian/mysql2.rb
byroot: Hum... I don't now. `Semian.register` have default values. If we want these to be always explicitly defined, then we might as well remove the default values no?

lib/semian/mysql2.rb
byroot: Yes.

test/test_mysql2.rb
byroot: Will do.

lib/semian/mysql2.rb
Sirupsen: Yeah, I'm not a fan of defaults. We wouldn't use them in Shopify.

byroot:
@Sirupsen I think you'll like the tests better now.

byroot:
As for the defaults, I agree that we shouldn't have any, but if you are ok with it, I will remove them in a followup.

Rakefile
```diff
@@ -29,6 +29,11 @@ else
   task :build do; end
 end
 
+task :populate_proxy do
+  require 'toxiproxy'
+  Toxiproxy.populate(File.expand_path('../test/fixtures/toxiproxy.json', __FILE__))
+end
```
Sirupsen: Interesting, I've always done this in the test helper. This is not a bad idea, but the problem is that running a single test file (which I for example do often) will fail with this approach. Since Semian will have a bunch of latency tests that will be prone to being slow, it's likely we'll end up in a situation where developers will run single tests a lot—where this won't work.

lib/semian.rb
```diff
@@ -83,11 +83,21 @@ def register(name, tickets: 0, permissions: 0660, timeout: 1, success_threshold:
     resources[name] = ProtectedResource.new(resource, circuit_breaker)
   end
 
+  def retrieve_or_register(name, **args)
+    self[name] || register(name, **args)
+  end
```
Sirupsen: Nice, I like this!

test/test_mysql2.rb
```diff
@@ -0,0 +1,148 @@
...
+      tickets: 1,
+      timeout: 0.05,
+      error_threshold: 1,
+      success_threshold: 2,
+      error_timeout: 5,
+    }.merge(semian_options)
+    Mysql2::Client.new(host: '127.0.0.1', port: '43306', semian: options)
+  end
+
+  class FakeMysql < Mysql2::Client
```
Sirupsen: Perhaps we should opt in to use Mocha from the get go? Thoughts?

test/test_mysql2.rb
```diff
@@ -0,0 +1,148 @@
...
+  def connect_to_mysql!(semian_options = {})
+    options = {
+      name: :testing,
+      tickets: 1,
+      timeout: 0.05,
+      error_threshold: 1,
+      success_threshold: 2,
+      error_timeout: 5,
+    }.merge(semian_options)
+    Mysql2::Client.new(host: '127.0.0.1', port: '43306', semian: options)
```
Sirupsen: This port is in the ephemeral port range (`32,768` and onwards), which means another connection could've taken it, causing strange intermittent test failures. It's unlikely in this test case because the number of connections is small, but we've seen this in Shopify. Pick a port < `32,768`.

test/test_mysql2.rb
```diff
@@ -0,0 +1,148 @@
...
+      assert_equal 2, client.query('SELECT 1 + 1 as sum;').to_a.first['sum']
+    end
+  end
+
+  private
+
+  def background(&block)
+    thread = Thread.new(&block)
+    threads << thread
+    thread.join(0.1)
```
Sirupsen: Why the argument to `#join`?

Rakefile
byroot: Yeah it's just that there is no test helper yet. I need to fix that. 

Maybe in a followup if you don't mind.

test/test_mysql2.rb
```diff
@@ -0,0 +1,148 @@
...
+
+  def connect_to_mysql!(semian_options = {})
+    options = {
+      name: :testing,
+      tickets: 1,
+      timeout: 0.05,
+      error_threshold: 1,
+      success_threshold: 2,
+      error_timeout: 5,
+    }.merge(semian_options)
```
Sirupsen: Not a big deal, but I'd consider making this a constant at the top to make it absolutely clear that this is the configuration we're testing throughout. Especially the ticket count is important.

test/test_mysql2.rb
```diff
@@ -0,0 +1,148 @@
...
+    Semian.destroy(:mysql_testing)
+  end
+
+  def teardown
+    threads.each { |t| t.kill }
+    @threads = []
+  end
+
+  def test_semian_identifier
+    assert_equal :mysql_foo, FakeMysql.new(semian: {name: 'foo'}).semian_identifier
```
Sirupsen: Any major reason you didn't want to use the MySQL client? I don't see a reason with connecting N times here.

test/test_mysql2.rb
byroot: Yes maybe. I hate mocha though. I'll see if there is a better mocking library for minitest. Unless you have objections.

test/test_mysql2.rb
```diff
@@ -0,0 +1,148 @@
...
+    assert_equal :'mysql_127.0.0.1:3306', FakeMysql.new(host: '127.0.0.1').semian_identifier
+    assert_equal :'mysql_example.com:42', FakeMysql.new(host: 'example.com', port: 42).semian_identifier
+  end
+
+  def test_resource_acquisition_for_connect
+    client = connect_to_mysql!
+
+    Semian[:mysql_testing].acquire do
+      assert_raises Mysql2::SemianError do
+        Mysql2::Client.new(semian: {name: :testing, tickets: 1, timeout: 0})
```
Sirupsen: Why doesn't this use `connect_to_mysql!`?

test/test_mysql2.rb
```diff
@@ -0,0 +1,148 @@
...
+
+  def yield_to_background
+    threads.each(&:join)
+  end
+
+  def connect_to_mysql!(semian_options = {})
+    options = {
+      name: :testing,
+      tickets: 1,
+      timeout: 0.05,
```
Sirupsen: Since we're not asserting anything on the timeout itself (.. which is flaky), you should consider just making this 0.

test/test_mysql2.rb
```diff
@@ -0,0 +1,148 @@
...
+        connect_to_mysql!
+      end
+    end
+  end
+
+  def test_circuit_breaker_on_connect
+    @proxy.downstream(:latency, latency: 500).apply do
+      background { connect_to_mysql! }
+
+      3.times do
```
Sirupsen: Use the constant from the Semian options.

test/test_mysql2.rb
```diff
@@ -0,0 +1,148 @@
...
+      end
+    end
+
+    yield_to_background
+
+    assert_raises Mysql2::SemianError do
+      connect_to_mysql!
+    end
+
+    Timecop.travel(10) do
```
Sirupsen: Ditto.

test/test_mysql2.rb
```diff
@@ -0,0 +1,148 @@
...
+      3.times do
+        assert_raises Mysql2::SemianError do
+          connect_to_mysql!
+        end
+      end
+    end
+
+    yield_to_background
+
+    assert_raises Mysql2::SemianError do
```
Sirupsen: Is there a more specific type for circuit breaker errors? We should embed this information into this error.

test/test_mysql2.rb
byroot: No sure what you mean. `Mysql2::Client` automatically attempt to connect when instantiated. Since the default is to use a socket, that might just fail.

I just want to test the resource naming here.

test/test_mysql2.rb
byroot: It should that's a mistake.

test/test_mysql2.rb
```diff
@@ -0,0 +1,148 @@
...
+
+class TestMysql2 < Test::Unit::TestCase
+  def setup
+    @proxy = Toxiproxy[:semian_test_mysql]
+    Semian.destroy(:mysql_testing)
+  end
+
+  def teardown
+    threads.each { |t| t.kill }
+    @threads = []
```
Sirupsen: I'd use an `attr_accessor` here. It's awkward to use instance variable + method.

test/test_mysql2.rb
```diff
@@ -0,0 +1,148 @@
...
+  end
+
+  def test_circuit_breaker_on_query
+    client = connect_to_mysql!
+    client2 = connect_to_mysql!
+
+    @proxy.downstream(:latency, latency: 1000).apply do
+      background { client2.query('SELECT 1 + 1;') }
+
+      3.times do
```
Sirupsen: Same comment on constants.

test/test_mysql2.rb
```diff
@@ -0,0 +1,148 @@
...
+      end
+    end
+
+    yield_to_background
+
+    assert_raises Mysql2::SemianError do
+      client.query('SELECT 1 + 1;')
+    end
+
+    Timecop.travel(10) do
```
Sirupsen: Ditto.

.travis.yml
```diff
@@ -2,5 +2,10 @@ language: ruby
 
 sudo: false
 
+before_install:
+  - wget -O toxiproxy-1.0.0.deb https://github.com/Shopify/toxiproxy/releases/download/v1.0.0/toxiproxy_1.0.0_amd64.deb
+  - sudo dpkg -i toxiproxy-1.0.0.deb
+  - sudo service start toxiproxy
```
Sirupsen: Provide a `script/install-toxiproxy-debian` for this so third-party developers can set this up as well, and include instructions for getting started in the README (run `script/setup` which runs `script/install-toxiproxy-debian`).

test/test_mysql2.rb
byroot: Indeed.

Sirupsen:
Really high quality work @byroot, love the tests :heart_eyes:

Pick and choose from the comments and then this is ready to ship.

test/test_mysql2.rb
Sirupsen: Completely up to you. I've used `rr` in the past, but don't really have an opinion.

Sirupsen:
OK with removing defaults in a follow-up.

test/test_mysql2.rb
byroot: Good idea. Like a 2 subclass of `Mysql2::SemianError`?  `Mysql2::ResourceOccupiedError` and `Mysql2::ResourceMarkedDeadError`.

(might now be the best names ever, but I guess you get the idea.)

test/test_mysql2.rb
byroot: Not sure what you mean. Are you ok with using timecop to test error timeout?

byroot: Nvm. Constants.

test/test_mysql2.rb
Sirupsen: Exactly!

byroot:
So:

  - The client now use `ProtectedResource#acquire` that delegate to `CircuitBreaker#acquire` and `Resource#acquire`
  - `CircuitBreaker#acquire` do not take a fallback, but preemptively raise a `CircuitOpenError` is the circuit is open.
  - The client now raises either a `Mysql2::ResourceOccupiedError` or a `Mysql2::CircuitOpenError`. Both inherits from `Mysql2::Error`.

Also:

  - `with_fallback` might actually be useless now. Maybe we should get rid of it.
  - About the ghetto concurrency, I can't use a cond signal because I can't plug mysql inside the semaphore. But that way seems fairly reliable, I haven't had any weird failure. Anyway we can look how to improve it in a followup.

I'll :sheep: after lunch unless any of you guys have concern until then.

Rakefile
Sirupsen: K

lib/semian/protected_resource.rb
```diff
@@ -1,8 +1,10 @@
+require 'forwardable'
+
 module Semian
   class ProtectedResource
```
Sirupsen: Just to recap, a protected resource is a resource that's both behind Semian and a Circuit Break?

byroot: Yes!

.travis.yml
```diff
@@ -1,6 +1,9 @@
 language: ruby
 
-sudo: false
+sudo: true
```
Sirupsen: :+1: should give faster builds as well.

Sirupsen:
:heart: :gift_heart: LGTM

fw42:
:+1:

PR #21, opened 2015-02-12
Monitoring api

byroot:
Here a very simple monitoring API for semian resources:

Usage example:
```ruby
Semian[:mysql_master].listen(:occupied) do |resource|
  StatsD.increment('Shopify.mysql.semian.timeout', 1, sample_rate: 1, tags: ["resource:#{resource.name}", "total_tickets:#{resource.tickets}"])
end
```

The 3 events are:
  - `:occupied`
  - `:circuit_open`
  - `:success`

@Sirupsen @fw42 @csfrancis for review please.

### DIFF
--- a/ext/semian/semian.c
+++ b/ext/semian/semian.c
@@ -489,7 +489,7 @@ void Init_semian()
   eInternal = rb_const_get(cSemian, rb_intern("InternalError"));
 
   rb_define_alloc_func(cResource, semian_resource_alloc);
-  rb_define_method(cResource, "initialize", semian_resource_initialize, 4);
+  rb_define_method(cResource, "_initialize", semian_resource_initialize, 4);
   rb_define_method(cResource, "acquire", semian_resource_acquire, -1);
   rb_define_method(cResource, "count", semian_resource_count, 0);
   rb_define_method(cResource, "semid", semian_resource_id, 0);
--- a/lib/semian/protected_resource.rb
+++ b/lib/semian/protected_resource.rb
@@ -4,7 +4,7 @@ module Semian
   class ProtectedResource
     extend Forwardable
 
-    def_delegators :@resource, :destroy, :count, :semid
+    def_delegators :@resource, :destroy, :count, :semid, :tickets
     def_delegators :@circuit_breaker, :reset
 
     def initialize(resource, circuit_breaker)
@@ -13,7 +13,20 @@ module Semian
     end
 
     def acquire(*args, &block)
-      @circuit_breaker.acquire { @resource.acquire(*args, &block) }
+      @circuit_breaker.acquire do
+        begin
+          @resource.acquire(*args) do
+            Semian.notify(:success, self)
+            yield self
+          end
+        rescue ::Semian::TimeoutError
+          Semian.notify(:occupied, self)
+          raise
+        end
+      end
+    rescue ::Semian::OpenCircuitError
+      Semian.notify(:circuit_open, self)
+      raise
     end
 
     def with_fallback(fallback, &block)
--- a/lib/semian/resource.rb
+++ b/lib/semian/resource.rb
@@ -1,6 +1,11 @@
 module Semian
   class Resource #:nodoc:
+    attr_reader :tickets, :name
+
     def initialize(name, tickets, permissions, timeout)
+      _initialize(name, tickets, permissions, timeout)
+      @name = name
+      @tickets = tickets
     end
 
     def destroy
--- a/test/test_semian.rb
+++ b/test/test_semian.rb
@@ -1,9 +1,9 @@
-require 'test/unit'
+require 'minitest/autorun'
 require 'semian'
 require 'tempfile'
 require 'fileutils'
 
-class TestSemian < Test::Unit::TestCase
+class TestSemian < MiniTest::Unit::TestCase
   def setup
     Semian.destroy(:testing) rescue nil
   end
@@ -33,7 +33,7 @@ class TestSemian < Test::Unit::TestCase
 
   def test_register_with_no_tickets_raises
     assert_raises Semian::SyscallError do
-      Semian.register :testing
+      Semian.register :testing, tickets: 0
     end
   end
 
@@ -160,6 +160,7 @@ class TestSemian < Test::Unit::TestCase
     Semian[:testing].acquire do
       acquired = true
       assert_equal 1, Semian[:testing].count
+      assert_equal 2, Semian[:testing].tickets
     end
 
     assert acquired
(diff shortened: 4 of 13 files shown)


byroot:
Thinking about it, I wonder if a global API wouldn't be better:

```
Semian.listen(:occupied) do |resource|
  StatsD.increment("Shopify.semian.#{resource.name}.occupied", 1, sample_rate: 1, tags: ["resource:#{resource.name}", "total_tickets:#{resource.tickets}"])
end
```

byroot:
Or even:

```ruby
Semian.listen do |event, resource|
  StatsD.increment("Shopify.semian.#{resource.name}.#{event}", 1, sample_rate: 1, tags: ["resource:#{resource.name}", "total_tickets:#{resource.tickets}"])
end
```

lib/semian/resource.rb
```diff
@@ -1,6 +1,11 @@
 module Semian
   class Resource #:nodoc:
+    attr_reader :tickets, :name
+
     def initialize(name, tickets, permissions, timeout)
+      @name = name
+      @tickets = tickets
+      _initialize(name, tickets, permissions, timeout)
```
fw42: this seems hacky

ext/semian/semian.c
```diff
@@ -489,7 +491,7 @@ void Init_semian()
   eInternal = rb_const_get(cSemian, rb_intern("InternalError"));
 
   rb_define_alloc_func(cResource, semian_resource_alloc);
-  rb_define_method(cResource, "initialize", semian_resource_initialize, 4);
+  rb_define_method(cResource, "_initialize", semian_resource_initialize, 4);
```
Sirupsen: ?

byroot:
So finally I went with the 3 solution for one main reason:

  - It allow to register the handler before the Resource is even registered, which will simplify things a lot.

lib/semian/protected_resource.rb
```diff
@@ -13,7 +13,22 @@ def initialize(resource, circuit_breaker)
...
+          @resource.acquire(*args, &block)
+        rescue ::Semian::TimeoutError
+          success = false
+          Semian.notify(:occupied, self)
+          raise
+        end
+      end
+    rescue ::Semian::OpenCircuitError
+      success = false
+      Semian.notify(:circuit_open, self)
```
Sirupsen: Should we do this in the circuit breaker instead of rescueing?

lib/semian/protected_resource.rb
```diff
@@ -13,7 +13,22 @@ def initialize(resource, circuit_breaker)
     end
 
     def acquire(*args, &block)
-      @circuit_breaker.acquire { @resource.acquire(*args, &block) }
+      success = true
+      @circuit_breaker.acquire do
+        begin
+          @resource.acquire(*args, &block)
+        rescue ::Semian::TimeoutError
```
Sirupsen: Same comment as to CB.

ext/semian/semian.c
byroot: See @fw42's comment. That allow me to have override the C constructor with Ruby code.

It's a fairly common practice for C extensions.

test/test_semian.rb
```diff
@@ -160,11 +160,91 @@ def test_count
     Semian[:testing].acquire do
       acquired = true
       assert_equal 1, Semian[:testing].count
+      assert_equal 2, Semian[:testing].tickets
     end
 
     assert acquired
   end
 
+  def test_occupied_instrumentation
```
Sirupsen: I'd make this a new test file.

lib/semian/protected_resource.rb
```diff
@@ -13,7 +13,22 @@ def initialize(resource, circuit_breaker)
...
+          Semian.notify(:occupied, self)
+          raise
+        end
+      end
+    rescue ::Semian::OpenCircuitError
+      success = false
+      Semian.notify(:circuit_open, self)
+      raise
+    ensure
+      Semian.notify(:success, self) if success
```
fw42: if `acquire` raises with a different exception than you expect here, this will still notify for success... is that intended?

lib/semian/protected_resource.rb
byroot: I wish, but I need to rescue anyway so that I can toggle the `success` flag.

lib/semian/protected_resource.rb
byroot: It is yes.

test/test_semian.rb
```diff
@@ -160,11 +160,91 @@ def test_count
...
+      if event == :occupied
+        assert_equal Semian[:testing], resource
+        notified = true
+      end
+    end
+
+    Semian[:testing].acquire do
+      begin
+        Semian[:testing].acquire {}
+      rescue Semian::TimeoutError
```
Sirupsen: `assert_raise`?

test/test_semian.rb
```diff
@@ -160,11 +160,91 @@ def test_count
...
+      if event == :circuit_open
+        assert_equal Semian[:testing], resource
+        notified = true
+      end
+    end
+
+    Semian[:testing].acquire do
+      begin
+        Semian[:testing].acquire {}
+      rescue Semian::TimeoutError
```
Sirupsen: ditto

test/test_semian.rb
```diff
@@ -160,11 +160,91 @@ def test_count
...
+      end
+    end
+
+    begin
+      Semian[:testing].acquire {}
+    rescue Semian::OpenCircuitError
```
Sirupsen: assert

test/test_semian.rb
```diff
@@ -160,11 +160,91 @@ def test_count
...
+    Semian.listen do |event, resource|
+      if event == :success
+        assert_equal Semian[:testing], resource
+        notified = true
+      end
+    end
+
+    begin
+      Semian[:testing].acquire { raise "Some error" }
+    rescue
```
Sirupsen: assertion

Sirupsen:
I like the simplicity of this. :+1:

fw42:
You are asserting that the expected notifications are triggered, but you are not asserting that no unexpected ones are triggered as well. Might make sense to add that.

Also I don't like the name ```#listen```. How about ```#notify``` or ```#subscribe``` (like [ActiveSupport](http://api.rubyonrails.org/classes/ActiveSupport/Notifications.html) calls it)?

Besides those two comments: :+1:

Sirupsen:
:+1: on consistency with AR

byroot:
> but you are not asserting that no unexpected ones are triggered as well. Might make sense to add that.

Right. Will do.

> Also I don't like the name #listen. How about #notify or #subscribe

Lol, I just renamed `subscribe` in `listen` 2 minutes ago. But ok, fine by me.

Sirupsen:
:+1:

PR #22, opened 2015-02-12
Alias original Mysql2 methods for easier monitoring

byroot:
So. https://github.com/Shopify/semian/pull/21 added a convenient API to monitor Semian itself.

But in many case you also want to instrument how long the original operation take, e.g. Mysql query time, Redis query time, etc.

So for monitoring puprose you need an easy access to the original method so you can do the usual patching mess.

That PR does exactly that, by renaming the original methods that are protected, so that they kind be patched for monitoring purpose.

Long story short: That will allow us to use `StatsD::Instrument` again.

@Sirupsen @fw42 for review please.

### DIFF
--- a/lib/semian/mysql2.rb
+++ b/lib/semian/mysql2.rb
@@ -22,6 +22,15 @@ module Semian
     DEFAULT_HOST = 'localhost'
     DEFAULT_PORT = 3306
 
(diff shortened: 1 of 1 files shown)


Sirupsen:
Good point. Add a comment above the included method. Then LGTM

PR #23, opened 2015-02-13
No more default values

byroot:
So as we talked before, this remove most of the default values because they can hardly make sense in the abstract.

@Sirupsen for review please.


Sirupsen:
Cool LGMT

PR #25, opened 2015-02-16
Allow not to configure semian for Mysql2

byroot:
I initially wrote that because I though we needed it, but actually we dont.

But maybe it still make sense.

@Sirupsen thoughts?

### DIFF
--- a/lib/semian/mysql2.rb
+++ b/lib/semian/mysql2.rb
@@ -32,8 +32,7 @@ module Semian
 
     def semian_identifier
       @semian_identifier ||= begin
-        semian_options = query_options[:semian] || {}
-        unless name = semian_options['name'.freeze] || semian_options[:name]
+        unless name = semian_options && semian_options[:name]
           host = query_options[:host] || DEFAULT_HOST
           port = query_options[:port] || DEFAULT_PORT
           name = "#{host}:#{port}"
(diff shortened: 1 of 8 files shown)


Sirupsen:
I don't think we should encourage this behaviour if we can avoid it.

lib/semian/mysql2.rb
```diff
@@ -62,7 +62,15 @@ def connect(*args)
     end
 
     def semian_resource
-      @semian_resource ||= ::Semian.retrieve_or_register(semian_identifier, **semian_options)
+      @semian_resource ||= case query_options[:semian]
+      when false
+        UnprotectedResource.new(semian_identifier)
```
Sirupsen: What about for Redis etc?

Sirupsen: Also do you have any ideas on how we don't have to implement this in every driver? That kind of sucks. I don't have any off the bat though.

byroot:
@Sirupsen and @fw42 I rebased that PR if you guys want to take a look.

lib/semian/unprotected_resource.rb
```diff
@@ -0,0 +1,31 @@
+module Semian
+  class UnprotectedResource
```
Sirupsen: Consider a short comment describing the intention of this class.

lib/semian/unprotected_resource.rb
```diff
@@ -0,0 +1,31 @@
...
+    attr_reader :name
+
+    def initialize(name)
+      @name = name
+    end
+
+    def tickets
+      0
```
Sirupsen: `-1` perhaps? Just to distinguish completely cleanly in monitoring. `0` is possible in the semaphore, `-1` is completely impossible.

Sirupsen:
Two minor, optional, comments. LGTM!

Sirupsen:
Re my previous comment: @byroot convinced me we need this.

PR #26, opened 2015-02-16
Redis adapter

byroot:
Semian integration for redis-rb

@Sirupsen @fw42 for review please (can totally wait tomorrow)

### DIFF
--- a/lib/semian/adapter.rb
+++ b/lib/semian/adapter.rb
@@ -1,0 +1,40 @@
+module Semian
+  module Adapter
+    def semian_identifier
+      raise NotImplementedError.new("Semian adapters must implement a `semian_identifier` method")
+    end
+
+    private
+
+    def acquire_semian_resource(scope:, &block)
+      return yield if resource_already_acquired?
+      semian_resource.acquire(scope: scope) do
+        mark_resource_as_acquired(&block)
+      end
+    rescue ::Semian::OpenCircuitError => error
+      raise self.class::CircuitOpenError.new(semian_identifier, error)
+    rescue ::Semian::BaseError => error
+      raise self.class::ResourceOccupiedError.new(semian_identifier, error)
+    end
+
+    def semian_resource
+      @semian_resource ||= ::Semian.retrieve_or_register(semian_identifier, **semian_options)
+    end
+
+    def semian_options
+      raise NotImplementedError.new("Semian adapters must implement a `semian_options` method")
+    end
+
+    def resource_already_acquired?
+      @resource_acquired
+    end
+
+    def mark_resource_as_acquired
+      previous = @resource_acquired
+      @resource_acquired = true
+      yield
+    ensure
+      @resource_acquired = previous
+    end
+  end
+end
--- a/lib/semian/mysql2.rb
+++ b/lib/semian/mysql2.rb
@@ -1,16 +1,10 @@
 require 'semian'
+require 'semian/adapter'
 require 'mysql2'
 
 module Mysql2
   class SemianError < Mysql2::Error
-    def initialize(semian_identifier, *args)
-      super(*args)
-      @semian_identifier = semian_identifier
-    end
-
-    def to_s
-      "[#{@semian_identifier}] #{super}"
-    end
+    include ::Semian::AdapterError
   end
 
   ResourceOccupiedError = Class.new(SemianError)
@@ -19,6 +13,11 @@ end
 
 module Semian
   module Mysql2
+    include Semian::Adapter
+
+    ResourceOccupiedError = ::Mysql2::ResourceOccupiedError
+    CircuitOpenError = ::Mysql2::CircuitOpenError
+
     DEFAULT_HOST = 'localhost'
     DEFAULT_PORT = 3306
 
@@ -44,11 +43,7 @@ module Semian
     end
 
     def query(*args)
-      semian_resource.acquire(scope: :query) { raw_query(*args) }
-    rescue ::Semian::OpenCircuitError => error
-      raise ::Mysql2::CircuitOpenError.new(semian_identifier, error)
-    rescue ::Semian::BaseError => error
-      raise ::Mysql2::ResourceOccupiedError.new(semian_identifier, error)
+      acquire_semian_resource(scope: :query) { raw_query(*args) }
     end
 
     private
--- a/lib/semian/redis.rb
+++ b/lib/semian/redis.rb
@@ -1,0 +1,58 @@
+require 'semian'
+require 'semian/adapter'
+require 'redis'
+
+class Redis
+  class SemianError < Redis::BaseConnectionError
+    include ::Semian::AdapterError
+  end
+
+  ResourceOccupiedError = Class.new(SemianError)
+  CircuitOpenError = Class.new(SemianError)
+end
+
+module Semian
+  module Redis
+    include Semian::Adapter
+
+    ResourceOccupiedError = ::Redis::ResourceOccupiedError
+    CircuitOpenError = ::Redis::CircuitOpenError
+
+    # The naked methods are exposed as `raw_query` and `raw_connect` for instrumentation purpose
+    def self.included(base)
+      base.send(:alias_method, :raw_io, :io)
+      base.send(:remove_method, :io)
+
+      base.send(:alias_method, :raw_connect, :connect)
+      base.send(:remove_method, :connect)
+    end
+
+    def semian_identifier
+      @semian_identifier ||= begin
+        opts = options[:semian] || options['semian'.freeze] || {}
+        name = opts[:name] || opts['name'.freeze]
+        name ||= "#{location}/#{db}"
+        :"redis_#{name}"
+      end
+    end
+
+    def io(&block)
+      acquire_semian_resource(scope: :query) { raw_io(&block) }
+    end
+
+    def connect
+      acquire_semian_resource(scope: :connection) { raw_connect }
+    end
+
+    private
+
+    def semian_options
+      opts = options[:semian] || options['semian'.freeze] || {}
+      opts = opts.map { |k, v| [k.to_sym, v] }.to_h
+      opts.delete(:name)
+      opts
+    end
+  end
+end
+
+::Redis::Client.include(Semian::Redis)
--- a/test/redis_test.rb
+++ b/test/redis_test.rb
@@ -1,0 +1,160 @@
+require 'test_helper'
+
+class TestRedis < MiniTest::Unit::TestCase
+  ERROR_TIMEOUT = 5
+  ERROR_THRESHOLD = 1
+  SEMIAN_OPTIONS = {
+    name: :testing,
+    tickets: 1,
+    timeout: 0,
+    error_threshold: ERROR_THRESHOLD,
+    success_threshold: 2,
+    error_timeout: ERROR_TIMEOUT,
+  }
+
+  attr_writer :threads
+  def setup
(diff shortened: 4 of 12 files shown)


lib/semian/adapter.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    def semian_resource
+      @semian_resource ||= ::Semian.retrieve_or_register(semian_identifier, **semian_options)
+    end
+
+    def semian_options
+      raise NotImplementedError.new("Semian adapters must implement a `semian_options` method")
+    end
+  end
+end
```
byroot: I extracted the common patterns here. Maybe that's too much, you guys tell me.

lib/semian/adapter.rb
```diff
@@ -0,0 +1,34 @@
+module Semian
+  module Adapter
+    def semian_identifier
+      raise NotImplementedError.new("Semian adapters must implement a `semian_identifier` method")
+    end
+
+    private
+
+    def acquire_semian_resource(scope:)
+      return yield if @resource_acquired # Prevent a single adapter from acquiring the resource twice by nesting calls
```
byroot: Redis basically does:

```ruby
def get
  ensure_connected do
    actuall_get
  end
end
```

Which ended up doing a 2 nested acquires. I believe we better prevent that.

byroot:
Also since we have a couple redis clients not using  semian yet, we might have to merge https://github.com/Shopify/semian/pull/25 too.

lib/semian/adapter.rb
Sirupsen: Nice catch.

Sirupsen: However what if you do something like:

```ruby
@redis1.pipelined do
  @redis1.set "omg", "hi"
  @redis1.set "lol", "hi"
  @redis2.set "omg", "sup"
end
```

Does `pipeline` actually end up acquiring a connection as well? I'm not completely familiar with how the client handles this situation.

Sirupsen: Or even if you could every do something like:

```ruby
@redis1.set "omg", -> { @redis2.set("i-hope", "nobody does this") }
```

lib/semian/adapter.rb
Sirupsen: I think that makes sense.

lib/semian/redis.rb
```diff
@@ -0,0 +1,58 @@
...
+      @semian_identifier ||= begin
+        opts = options[:semian] || options['semian'.freeze] || {}
+        name = opts[:name] || opts['name'.freeze]
+        name ||= "#{location}/#{db}"
+        :"redis_#{name}"
+      end
+    end
+
+    def io(&block)
+      acquire_semian_resource(scope: :io) { raw_io(&block) }
```
Sirupsen: Isn't the scope more something like `command` or `query`? A little more human readable when this is used primarily for monitoring.

Sirupsen: We should keep this consistent with MySQL unless it absolutely doesn't make sense.

lib/semian/redis.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      base.send(:alias_method, :raw_connect, :connect)
+      base.send(:remove_method, :connect)
+    end
+
+    def semian_identifier
+      @semian_identifier ||= begin
+        opts = options[:semian] || options['semian'.freeze] || {}
+        name = opts[:name] || opts['name'.freeze]
+        name ||= "#{location}/#{db}"
```
Sirupsen: Where does `location` and `db` come from? What are they usually?

test/redis_test.rb
```diff
@@ -0,0 +1,180 @@
...
+  end
+
+  private
+
+  def background(&block)
+    thread = Thread.new(&block)
+    threads << thread
+    thread.join(0.1)
+    thread
+  end
```
Sirupsen: Perhaps it makes sense to extract those to the test helper at this point?

lib/semian/redis.rb
byroot: They are implemented by `Redis::Client` https://github.com/redis/redis-rb/blob/be4326e7f2d399069831ba05ef53ba51c9a76c7e/lib/redis/client.rb#L107-L109

`location = "#{host}:#{port}"`. They both have proper defaults.

test/redis_test.rb
```diff
@@ -0,0 +1,180 @@
...
+    assert_raises Redis::CircuitOpenError do
+      client.get('foo')
+    end
+
+    Timecop.travel(ERROR_TIMEOUT + 1) do
+      assert_equal '2', client.get('foo')
+    end
+  end
+
+  private
```
Sirupsen: Needs a test on the nested acquire (although I'm still not sure what that'll look like).

lib/semian/redis.rb
byroot: Yeah, you're right. `connection` and `query` are probably the 2 scopes most adapters will have. I'll update Mysql2 accordingly.

lib/semian/adapter.rb
byroot: > Does pipeline actually end up acquiring a connection as well?

Pipelined is not a subclass of `Redis::Client`. Also the pipepline are first evaluated and then the client send the commands to the server, e.g:

```ruby
def pipelined
  yield commands_collector
  ensure_connected do
    perform commands_collector.commands
  end
end
```
(This is pseudo code but the idea is here)

> Or even if you could every do something like:

That would be fine too since the flag is an instance variable of the adapter, in this case each redis has his own flag.

test/redis_test.rb
byroot: It's already tested since all redis commands end up doing a nested acquire.

test/redis_test.rb
byroot: Indeed, I just copypasted mysql test suite like a lazy bastard. I'll factorize.

lib/semian/redis.rb
```diff
@@ -0,0 +1,58 @@
...
+    include ::Semian::AdapterError
+  end
+
+  ResourceOccupiedError = Class.new(SemianError)
+  CircuitOpenError = Class.new(SemianError)
+end
+
+module Semian
+  module Redis
+    include Adapter
```
fw42: Can we ```include Semian::Adapter``` for clarity?

lib/semian/adapter.rb
fw42: do we really need the comment here? Do we have a test that captures this possible bug?

lib/semian/mysql2.rb
```diff
@@ -19,6 +13,11 @@ def to_s
 
 module Semian
   module Mysql2
+    include Adapter
+
+    ResourceOccupiedError = ::Mysql2::ResourceOccupiedError
+    CircuitOpenError = ::Mysql2::CircuitOpenError
```
fw42: oO

fw42: why do we need this?

lib/semian/redis.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    ResourceOccupiedError = ::Redis::ResourceOccupiedError
+    CircuitOpenError = ::Redis::CircuitOpenError
+
+    # The naked methods are exposed as `raw_query` and `raw_connect` for instrumentation purpose
```
fw42: comment really necessary?

test/fixtures/toxiproxy.json
```diff
@@ -3,5 +3,10 @@
     "name": "semian_test_mysql",
     "upstream": "localhost:3306",
     "listen": "localhost:13306"
+  },
```
fw42: hm this file isn't really a fixture, is it? should this be in a different directory?

test/redis_test.rb
```diff
@@ -0,0 +1,180 @@
...
+    subscriber = Semian.subscribe do |event, resource, scope|
+      notified = true
+      assert_equal :success, event
+      assert_equal Semian[:redis_testing], resource
+      assert_equal :connect, scope
+    end
+
+    connect_to_redis!
+
+    assert notified, 'No notification have been emitted'
```
fw42: s/have/has/ or s/notification/notifications/

test/redis_test.rb
```diff
@@ -0,0 +1,180 @@
...
+    subscriber = Semian.subscribe do |event, resource, scope|
+      notified = true
+      assert_equal :success, event
+      assert_equal Semian[:redis_testing], resource
+      assert_equal :io, scope
+    end
+
+    client.get('foo')
+
+    assert notified, 'No notification have been emitted'
```
fw42: same

lib/semian/adapter.rb
Sirupsen: Why not have a comment here @fw42 ? It was helpful for me during review.

lib/semian/redis.rb
Sirupsen: I asked for this comment for the MySQL adapter. I think this comment is very helpful.

lib/semian/adapter.rb
fw42: because comments are often considered a code smell and easily get out of sync with the code. but if you think that it really improves the code here, then ok... I would prefer to have a helper method with a more descriptive name rather than a comment.. something like
```ruby
return yield if same_adapter_already_acquired_resource?
```

fw42: back to my other question: are any tests going to fail if this line is removed?

Sirupsen: In this case I think you're right :+1: in the other case I think it's still very helpful

byroot: > back to my other question: are any tests going to fail if this line is removed?

Yes all of the `redis_test.rb` suite will fail.

lib/semian/mysql2.rb
byroot: So that `Semian::Adapter` can raise them.

fw42: seems kinda hacky, but ok

byroot:
I just:

  - Refactored the double acquisition handling, with a meaningful method name to get rid of the comment
  - Refactored the test helpers
  - Move the toxiproxy config outside of the fixtures direcotry
  - `include Semian::Adapter` for clarity
  - Normalized scope names with `query` and `connection`

test/test_helper.rb
```diff
@@ -1,10 +1,17 @@
...
 require 'toxiproxy'
 require 'timecop'
 require 'tempfile'
 require 'fileutils'
 
+require 'helpers/background_helper'
+
 Semian.logger = Logger.new(nil)
-Toxiproxy.populate(File.expand_path('../fixtures/toxiproxy.json', __FILE__))
+Toxiproxy.populate(File.expand_path('../helpers/toxiproxy.json', __FILE__))
```
Sirupsen: Off-side comment: I think `Toxiproxy.populate` should take a Hash, I may have gotten a little aggro in forcing people to store it in a JSON file. I still think that should be done in most use cases, but what do you think about making it a Hash? Would make the introduction simpler too.

Sirupsen:
LGTM

Sirupsen:
It'll take some extra work in Shopify because it now means we have to protect all Redises, but that needed to be done anyway.

test/test_helper.rb
byroot: I think the JSON file is handy for project that have many proxies, but yes in that case a literal hash would be simpler.

Sirupsen: Definitely. But at that point people can abstract it themselves easily.

PR #32, opened 2015-02-20
Rename :occupied to :busy

csfrancis:
@byroot, @Sirupsen 

I was discussing this with @Sirupsen, and I find `occupied` to be kind of an awkward word to use when there are no tickets available. This renames `:occupied` to `:busy`.


byroot:
If you say so I believe you. You're speaking :us: better than me.

byroot:
Damn I forgot but we should rename the exceptions too.

Sirupsen:
@csfrancis ^

PR #34, opened 2015-02-20
Adds a disable env var

camilo:
Some processes that run offline actually don't benefit from semian, for instance processes that do maintenance by just moving a ton of resources around concurrently, they pretty much occupy  all the available tickets by design.

For those cases  I'm providing a way of disabling semian.

### DIFF
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -77,8 +77,8 @@ module Semian
   InternalError = Class.new(BaseError)
   OpenCircuitError = Class.new(BaseError)
 
-  def enabled?
-    !ENV['SEMIAN_DISABLED']
+  def semaphores_enabled?
+    !ENV['SEMIAN_SEMAPHORES_DISABLED']
   end
 
   module AdapterError
@@ -153,11 +153,11 @@ require 'semian/circuit_breaker'
 require 'semian/protected_resource'
 require 'semian/unprotected_resource'
 require 'semian/platform'
-if Semian.sysv_semaphores_supported? && Semian.enabled?
+if Semian.sysv_semaphores_supported? && Semian.semaphores_enabled?
   require 'semian/semian'
 else
   Semian::MAX_TICKETS = 0
   Semian.logger.info("Semian is not supported on #{RUBY_PLATFORM} - all operations will no-op") unless Semian.sysv_semaphores_supported?
-  Semian.logger.info("Semian is disabled, is this what you really want? - all operations will no-op") unless Semian.enabled?
+  Semian.logger.info("Semian is disabled, is this what you really want? - all operations will no-op") unless Semian.semaphores_enabled?
 end
 require 'semian/version'
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -157,7 +157,7 @@ if Semian.sysv_semaphores_supported? && Semian.semaphores_enabled?
   require 'semian/semian'
 else
   Semian::MAX_TICKETS = 0
-  Semian.logger.info("Semian is not supported on #{RUBY_PLATFORM} - all operations will no-op") unless Semian.sysv_semaphores_supported?
-  Semian.logger.info("Semian is disabled, is this what you really want? - all operations will no-op") unless Semian.semaphores_enabled?
+  Semian.logger.info("Semian sysv semaphores are not supported on #{RUBY_PLATFORM} - all operations will no-op") unless Semian.sysv_semaphores_supported?
+  Semian.logger.info("Semian semaphores are disabled, is this what you really want? - all operations will no-op") unless Semian.semaphores_enabled?
 end
 require 'semian/version'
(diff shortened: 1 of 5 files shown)


camilo:
@csfrancis @Sirupsen @byroot

byroot:
Hum. I didn't like our previous way of opt-in. But I can't find any argument against an opt-out mechanism.

I guess :+1:

byroot:
> they pretty much occupy all the available tickets by design.

That being said, semian is not just about the semaphore. The resource is also protected by a circuit breaker that can be handy even for such scripts and workers, so that they don't hammer a resource used by the rest of the app.

lib/semian.rb
```diff
@@ -149,10 +153,11 @@ def resources
 require 'semian/protected_resource'
 require 'semian/unprotected_resource'
 require 'semian/platform'
-if Semian.supported_platform?
+if Semian.supported_platform? && !Semian.enabled?
```
csfrancis: Shouldn't this be ` && Semian.enabled?`

lib/semian.rb
```diff
@@ -149,10 +153,11 @@ def resources
...
   require 'semian/semian'
 else
   Semian::MAX_TICKETS = 0
-  Semian.logger.info("Semian is not supported on #{RUBY_PLATFORM} - all operations will no-op")
+  Semian.logger.info("Semian is not supported on #{RUBY_PLATFORM} - all operations will no-op") if !Semian.supported_platform?
+  Semian.logger.info("Semian is disabled, is this what you really want? - all operations will no-op") if !Semian.enabled?
```
Sirupsen: unless

lib/semian.rb
Sirupsen: :+1:

fw42:
> That being said, semian is not just about the semaphore. 

Agree. The flag should just disable the semaphore instrumentation, but not prevent all the code from being loaded. That could cause lots of NameError issues and stuff like that in the application code. Opt-in code like non-semaphore circuit breakers should still work.

camilo:
> Agree. The flag should just disable the semaphore instrumentation, but not prevent all the code from being loaded. That could cause lots of NameError issues and stuff like that in the application code.

It should not tho, if that were the case exactly the same would happen on unsupported platforms.

camilo:
>  Opt-in code like non-semaphore circuit breakers should still work.

Also that is kinda bad, if we have an enabled/disabled it should disable all the things or none, if we want that feature control  then we need some config DLS ( :-1: ) or many env vars per feature ( :-1: :-1: :-1: ).

The only thing I'm trying to achieve is to opt-out in situations where semian does not matter (admin scripts and the like), and some features might get in the way, also could be useful for testing as in disable and see that is what we hypothesize would happen without circuits and semaphores actually happens.

csfrancis:
I think I lean towards what Flo is saying here. What @camilo is trying to do is make the semaphore operations no-op, which isn't exactly the same as "disabling Semian" (because yes, Semian is more than just semaphores now). Making semaphores no-op is also what we do in the case where we're running on an unsupported platform.

Perhaps this should be called `SEMIAN_SEMAPHORE_NOOP`, or something else that clearly states that this is only disabling semaphores.

camilo:
I guess unsuported platforms only unsupport semaphores in that case is this what is happening here?

```ruby
if Semian.supported_platform? && Semian.enabled?
  require 'semian/semian'
else
```

Is not obvious by looking at that code that it ONLY disables semaphores given that the name of the gem itself is semian.

byroot:
Ok, I'd say go for it. No semaphore will be acquired, but the circuit breaker will still be there so the behaviour shouldn't change overall.

Please cleanup the `if ! / unless` though.

csfrancis:
> I guess unsuported platforms only unsupport semaphores in that case is this what is happening here?

Yeah, that's what it is. Maybe clean this up so the check is something like `sysv_semaphores_supported?` instead of `supported_platform?`.

camilo:
`if ! / unless` does it even matter? aimma just make if `if !!!`

byroot:
Good joke

lib/semian.rb
```diff
@@ -149,10 +153,11 @@ def resources
 require 'semian/protected_resource'
 require 'semian/unprotected_resource'
 require 'semian/platform'
-if Semian.supported_platform?
+if Semian.supported_platform? && Semian.enabled?
   require 'semian/semian'
 else
   Semian::MAX_TICKETS = 0
-  Semian.logger.info("Semian is not supported on #{RUBY_PLATFORM} - all operations will no-op")
+  Semian.logger.info("Semian is not supported on #{RUBY_PLATFORM} - all operations will no-op") if !Semian.supported_platform?
```
fw42: :trollface:

camilo:
ok now I added  `if !!!!!!!!!` and changed the names both for the env var and method.

camilo:
hmrr iamma change `enabled?` too to `semaphores_enabled?`

camilo:
iamma merge unless someone objects.

csfrancis:
LGTM.

byroot:
Sure merge away. I have an open PR to update semian. I'll update it once you merge.

PR #36, opened 2015-02-23
Get rid of with_fallback

byroot:
@fw42 @Sirupsen for review please.

I don't think it belongs to Semian. Semian is for protecting datastore drivers etc, with_fallback can only be used at an higher level.

### DIFF
--- a/lib/semian/circuit_breaker.rb
+++ b/lib/semian/circuit_breaker.rb
@@ -25,14 +25,6 @@ module Semian
       result
     end
 
-    def with_fallback(fallback, &block)
-      acquire(&block)
-    rescue *@exceptions
-      evaluate_fallback(fallback)
-    rescue OpenCircuitError
-      evaluate_fallback(fallback)
-    end
-
     def request_allowed?
       return true if closed?
       half_open if error_timeout_expired?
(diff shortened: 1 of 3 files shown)


Sirupsen:
LGTM

Semian is for failing fast. Fallbacks and that kind of thing should be done with decorators in the application. However, we should document how we recommend doing this.

fw42:
> Semian is for protecting datastore drivers

why can't it be both? This pattern has proved incredibly useful for Shopify. Why not share it?

Sirupsen:
I thought we had removed `with_fallback` in Shopify in favour of just plain rescues?

byroot:
First Semian's `with_fallback` is not identical to the one we use, and can't be used the same way. I made a mistake I should never had implemented it that way.

The current implementation acquire the resource directly, which mean that if you use `with_fallback` too high in the stack you'll hold the resource for much longer that you need, which is bad.

Secondly our implementation is not perfect yet, it can close Redis1's circuit breaker if it catches exceptions from Redis2. I aim to fix that and then I'll reconsider it's inclusion in Semian.

Is that ok with you @fw42?

fw42:
sure, if you think this was broken to begin with.. then remove it

PR #37, opened 2015-02-23
Consider Redis::BaseConnectionError in the circuit breaker errors count

byroot:
@Sirupsen @fw42 

I feel bad about that one. I should have done it way earlier. Connection and query timeout should increment the circuit breaker count.


fw42:
:+1:

PR #38, opened 2015-02-23
Expose the semian identifier in AdapterExceptions

byroot:
@fw42 @Sirupsen for review please.

The idea is to be able to discriminate semian errors based on which resource fired it.

### DIFF
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -78,6 +78,8 @@ module Semian
   OpenCircuitError = Class.new(BaseError)
 
   module AdapterError
+    attr_reader :semian_identifier
+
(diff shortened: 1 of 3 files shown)


Sirupsen:
LGTM

I'm assuming a follow up is fixing the case you described in #36?

byroot:
Indeed. But for that It also requires to be able to identify the adapter native errors e.g. `Redis::BaseConnectionError`.

I'm not sure how I'll do it yet, but this pose the basics.

Sirupsen:
K, sounds good.

fw42:
:+1:

PR #43, opened 2015-02-25
Also rescue Errno::EINVAL in Redis adapter because of an upstream bug

byroot:
@Sirupsen @fw42 for review please.

Not sure if it's worth testing, it only appear after several reconnection attempts.

### DIFF
--- a/lib/semian/redis.rb
+++ b/lib/semian/redis.rb
@@ -4,6 +4,7 @@ require 'redis'
 
 class Redis
   Redis::BaseConnectionError.class_exec { attr_accessor :semian_identifier }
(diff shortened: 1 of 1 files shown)


Sirupsen:
How did you find this? Can we have a test case for it?

byroot:
It appeared in Shopify's test suite, in a file where we test a lot of behaviour with all redises down in a row. That triggered that bug. I can try to add a repro in the test suite.

byroot:
As discussed offline, it appear that big appear mostly on OS X. I tried to add a test for it but was unsuccessful.

I'll merge anyway and monitor the upstream ticket for resolution.

Ref: https://github.com/redis/hiredis-rb/issues/21

fw42:
wow, this sucks

PR #51, opened 2015-09-17
OS X support

charliesome:
This pull request adds OS X support to Semian.

The biggest potentially controversial change here is that I've removed the ticket acquisition timeout entirely. This is because OS X doesn't have `semtimedop`. I asked @Sirupsen about this and he backgrounded me on the original motivation for this timeout and why it isn't strictly necessary anymore.

Because we no longer block on semaphore operations, I was also able to remove the GVL release code.

cc @Sirupsen @csfrancis @byroot


byroot:
> this timeout and why it isn't strictly necessary anymore.

Yes. We recently removed all usage of it. So it's fine. We'll have to do a major version bump though.

Flagging this for later review even though @csfrancis is the authoritative person here.

Also you have a CI failure that seems legit.

charliesome:
The other thing I discussed with @Sirupsen is the use of `SHA1` to generate Semaphore keys from strings. There's a slight potential for this to generate the same key for multiple resources, or even between totally different applications that both happen to use SysV semaphores.

Is there any reason why Semian uses `SHA1` for this and not `ftok`? `ftok` generates a key from a file path and guarantees uniqueness this way, rather than SHA1's probabilistic uniqueness.

charliesome:
@byroot Pushed up f7372fb which fixes that failure. Turns out that test was implicitly relying on timeouts (as it tried to acquire the resource immediately after sending `SIGKILL` to the child process that had already acquired the resource). I've added a call to `Process.waitpid` after `Process.kill` so we only try to acquire the resource after the child process has died.

Sirupsen:
The argument I have for not requiring a ticket count I tried to explain to Charlie last night is that the equation for a ticket count's effectiveness effectively looks like this: `p^t = q`, where `p` is probability of sampling a worker talking to the resource that has ticket count `t`, an `q` is the accepted false positive rate for sampling `t` workers talking to the resource. When `t` grows beyond 6-8 because of this exponential relationship a timeout has a very questionable value, and makes it harder to argue about an effective ticket count. A `t` of, say, 8 is as appropriate for 15 workers as it is for 80 workers (this means that it's better, for Semian, to have more workers on the same box as you reduce capacity less during a failure in that case). The reason we introduced the timeout was that we ran SysV namespaces with only 4 workers, which meant that we couldn't go beyond a ticket count of 2 and thus needed another mechanism to tune. However, we no longer do this and I don't anticipate us doing this again—and this is why I think it's absolutely fine to remove `semtimedop` in favour of never having a timeout.

Sirupsen:
I've put it on my list to test this PR in production. I'll attempt to do that next week. Nothing really jumps out, thanks @charliesome :)

ext/semian/extconf.rb
```diff
@@ -23,7 +23,7 @@
 have_func 'rb_thread_blocking_region'
 have_func 'rb_thread_call_without_gvl'
 
-$CFLAGS = "-D_GNU_SOURCE -Werror -Wall "
+$CFLAGS = "-D_GNU_SOURCE -Wall "
```
csfrancis: Why did you have to get rid of `-Werror`?

ext/semian/semian.c
```diff
@@ -167,13 +134,11 @@ static VALUE
...
 {
   short delta;
-  struct timespec ts = { 0 };
-  ts.tv_sec = kInternalTimeout;
 
   if (get_max_tickets(tc->sem_id) != tc->tickets) {
     delta = tc->tickets - get_max_tickets(tc->sem_id);
 
-    if (perform_semop(tc->sem_id, kIndexTickets, delta, 0, &ts) == -1) {
+    if (perform_semop(tc->sem_id, kIndexTickets, delta, 0) == -1) {
```
csfrancis: This is a *little* racier now because there's no longer a timeout when trying to adjust the count. You could conceivably have exhausted all the tickets and then try to lower the overall ticket count, which could cause this to fail now that there's no timeout. But that seems pretty unlikely.

ext/semian/extconf.rb
charliesome: `SHA1` is deprecated in the OS X system OpenSSL headers and causes a warning.

fw42: Lets switch to something else then rather than disabling the warning

Sirupsen: @fw42 Charlie already suggested following up with ftok

byroot:
I tested this branch with shopify in dev and test. Seems all green. We can try it out on production servers once the last concerns are settled.

ext/semian/extconf.rb
charliesome: Switching to `ftok` will likely require changes in the external Semian API. This might not be a big deal since this branch will already require a major version bump, but using `ftok` properly will not be as simple as just replacing the `SHA1` call.

We could pass `-Wno-deprecated-declarations` to disable this *specific* warning, but I'm not sure how that will affect compilers that don't support that option.

ext/semian/semian.c
charliesome: @csfrancis Good catch. Would it be acceptable if we changed this call to not pass `IPC_NOWAIT` (so it'll block until there are enough tickets to lower the max count)?

charliesome:
Since this pull request will already require a major version bump, I'd like to discuss another breaking change that we could make while we're at it.

IMO we should use `ftok` rather than `SHA1` to generate semaphore keys. `ftok` returns a key which is generated from a file's identity on disk rather than just being a simple digest. This guarantees that there won't be clashes between two semaphores. While clashes are fairly unlikely with SHA1, (1 in 4 billion with a 32 bit `key_t`), a guarantee that a clash won't happen is even better :grin: 

Using `ftok` requires that we have a file on disk per resource that is visible to all workers. We could just use `/tmp/semian.#{resource_name}.sem` or something, but this feels a little sketchy and makes more assumptions about environments that Semian will be used in than I'm comfortable making - it assumes that `/tmp` exists and is writable (in this brave new world of containerising everything, this may not be a reasonable assumption). It also assumes that the all processes sharing an IPC namespace *also* share the same root directories and mount namespaces - again, not necessarily a reasonable assumption these days.

I think the ideal solution here is to add a parameter to `Semian::Resource` (and expose that up through `Semian.register`) that allows the user to customise the filename used for `ftok`. That could set to something in `/tmp` by default, or it could be a required parameter.

byroot:
> Using ftok requires that we have a file on disk per resource that is visible to all workers.

This (unless I'm missing something) is I'm afraid impossible for us since we're running our processes in multiple docker containers. We do share the IPC namespace between containers, but not the filesytem.

ext/semian/semian.c
csfrancis: I'm not really sure if I like that any better. If we omit `IPC_NOWAIT` then we could risk the process blocking indefinitely *if* there were processes that decided to hold onto their tickets (also seems very unlikely). It's kind of gross, but what about `sleep`ing on a failure and retrying a given number of times (basically emulating what we were doing before with the timeout)?

Sirupsen:
Yeah, @byroot has a valid point for our deployment. I really don't want to do a shared mount between all the containers just for this :/

csfrancis:
> While clashes are fairly unlikely with SHA1, (1 in 4 billion with a 32 bit key_t), a guarantee that a clash won't happen is even better

`ftok` isn't bulletproof either, unfortunately (from the BSD docs on OSX):
> BUGS
The returned key is computed based on the device minor number and inode of the specified path, in combination with the lower 8 bits of the given id.  Thus, it is quite possible for the routine to return duplicate keys.

csfrancis:
Perhaps we could do something like store a mapping of all active `resource_id` -> `sem_id`s in a hash to detect collisions? It seems unlikely that we would encounter one, but if we did, the user would definitely want to know about it.

byroot:
I think you are going to far guys. We're talking about hashing human identifiers. Even MD5 would do. The risk of accidental collisions is extremely low.

charliesome:
>`ftok` isn't bulletproof either, unfortunately

Ugh, I missed that. That's a shame. I guess SHA1 is good enough in that case then

ext/semian/semian.c
Sirupsen: Agree with @csfrancis, we need to have retries and a timeout here. This is also needed in https://github.com/Shopify/semian/pull/54 for acquiring the lock around mutating the shared memory. They can share a fallback for `semtimedop` though.

kyewei: Is one of the issues preventing this from going forward the lack of a `semtimedop` shim with timeouts for OS X? I looked into this. We'd need to bring back the nogvl lock.

Initially, I found @byroot 's shim here: https://github.com/Shopify/semian/pull/7#issuecomment-59564579 , but I found after running it on a live OS X install that using signals like `SIGALRM` won't work in a multi-threaded environment - a process can only schedule one alarm at a time, and alarm handlers don't easily distinguish between different threads, which are both big problems.

The way around this is to use some form of sleep. I wrote one that works (it can of course be improved on):
```c
#include <unistd.h> // for usleep

// timeval_subtract is adapted from 
// https://opensource.apple.com/source/distcc/distcc-31.0.81/src/timeval.c
int
timeval_subtract(struct timeval *result,
                 const struct timeval *x,
                 const struct timeval *y)
{
  struct timeval _x = *x;
  struct timeval _y = *y;

  /* Perform the carry for the later subtraction by updating Y. */
  if (_x.tv_usec < _y.tv_usec) {
    int nsec = (_y.tv_usec - _x.tv_usec) / 1000000 + 1;
    _y.tv_usec -= 1000000 * nsec;
    _y.tv_sec += nsec;
  }
  if (_x.tv_usec - _y.tv_usec > 1000000) {
    int nsec = (_x.tv_usec - _y.tv_usec) / 1000000;
    _y.tv_usec += 1000000 * nsec;
    _y.tv_sec -= nsec;
  }

  /* Compute the time remaining to wait.
     `tv_usec' is certainly positive. */
  result->tv_sec = _x.tv_sec - _y.tv_sec;
  result->tv_usec = _x.tv_usec - _y.tv_usec;

  /* Return 1 if result is negative. */
  return _x.tv_sec < _y.tv_sec;
}

int 
__attribute__((weak)) // GCC provides this, can also use an #ifdef
semtimedop(int semid, struct sembuf *sops, size_t nsops, const struct timespec *timeout)
{
  struct sembuf _sops = *sops;
  _sops.sem_flg |= IPC_NOWAIT;
  int code = semop(semid, &_sops, nsops);

  struct timeval timeout_date, current_date;
  gettimeofday(&timeout_date, NULL);
  timeout_date.tv_usec += timeout->tv_nsec/1000;
  while (timeout_date.tv_usec > 1000000) {
    timeout_date.tv_usec -= 1000000;
    timeout_date.tv_sec += 1;
  }
  timeout_date.tv_sec += timeout->tv_sec;

  int timed_out = 0;
  while (!timed_out && -1 == code && EAGAIN == errno) {
    errno = 0;
    code = 0;
    usleep(100); // Can be changed
    code = semop(semid, &_sops, nsops);
    gettimeofday(&current_date, NULL); // Get new time
    timed_out = timeval_subtract(&current_date, &timeout_date, &current_date); // See if elapsed time > timeout
  }
  return code;
}
```

byroot: > Initially, I found @byroot 's shim here

OMG don't use that! :p

Sirupsen:
I'd question again why we even need this. Why can't we just disable this in test and dev? @charliesome

byroot:
@Sirupsen because then you can't run your integration tests on OSX. :sad_panda:

Sirupsen:
Yeah, but those don't `fork` etc., so you should be able to use the in-memory adapter, no?

byroot:
Oh. Well, if we end up having an in-memory semaphore, then yes.

Soleone:
any updates on this by chance? this came up to test certain semian details in development mode in osx

bf4:
@sirupsen @charliesome This looks like an interesting PR but is about a year and a half old... any reason it's still open? looks like it's targeted for only Linux in prod, so it could be closed with 'make sure your dev environment is linux; we don't expect support for BSD/OSX'?

sirupsen:
@bf4 we're still interested in this work upstream, someone just needs to own it :)

PR #54, opened 2015-09-28
Initial commit implementing a per-host circuit breaker state using shared memory and semaphores

kyewei:
@Sirupsen @csfrancis for review, suggest refactoring if needed!

Implementation details:
Shared memory stores the space of `int`+`int`+`double[]`, corresponding to `@successes`, `@errors.count`, and the `@errors` array itself (stored as doubles). The array has a max size of `@errors.count` since the Ruby code ensures the array never grows above that size.

* Most of semaphore initialization, permissions, etc were borrowed from `semian.c`. 
* A regular array (as opposed to a linked list) was chosen because of the small array length and simplicity.
* Shared memory is not explicitly marked for deletion from the C side since (1) it's on the order of ten's of bytes, (2) it won't increase if the same key/id is used to create the memory. This is the same(?) rationale as why in `semian.c`, `semian_resource_free()` doesn't clear the semaphores, instead leaving them to be cleared on reboot.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 /.bundle/
-/lib/semian/*.so
-/lib/semian/*.bundle
+/lib/semian*/*.so
+/lib/semian*/*.bundle
 /tmp/*
 *.gem
 /html/
--- a/Rakefile
+++ b/Rakefile
@@ -24,6 +24,10 @@ if Semian.sysv_semaphores_supported?
     ext.ext_dir = 'ext/semian'
     ext.lib_dir = 'lib/semian'
   end
+  Rake::ExtensionTask.new('semian_cb_data', GEMSPEC) do |ext|
+    ext.ext_dir = 'ext/semian_cb_data'
+    ext.lib_dir = 'lib/semian_cb_data'
+  end
   task :build => :compile
 else
   task :build do; end
@@ -45,5 +49,5 @@ task :test => :build
 # ==========================================================
 require 'rdoc/task'
 RDoc::Task.new do |rdoc|
-  rdoc.rdoc_files.include("lib/*.rb", "ext/semian/*.c")
+  rdoc.rdoc_files.include("lib/*.rb", "ext/semian/*.c",  "ext/semian_cb_data/*.c")
 end
--- a/ext/semian_cb_data/semian_cb_data.c
+++ b/ext/semian_cb_data/semian_cb_data.c
@@ -1,0 +1,900 @@
+#include <ruby.h>
+#include <stdio.h>
+
+#include <errno.h>
+#include <sys/types.h>
+#include <sys/ipc.h>
+#include <sys/sem.h>
+#include <sys/shm.h>
+#include <sys/time.h>
+
+#include <openssl/sha.h>
+#include <unistd.h>
+#include <stdbool.h>
+
+// needed for semctl
+union semun {
+  int              val;    /* Value for SETVAL */
+  struct semid_ds *buf;    /* Buffer for IPC_STAT, IPC_SET */
+  unsigned short  *array;  /* Array for GETALL, SETALL */
+  struct seminfo  *__buf;  /* Buffer for IPC_INFO
+                             (Linux-specific) */
+};
+
+#if defined(HAVE_RB_THREAD_CALL_WITHOUT_GVL) && defined(HAVE_RUBY_THREAD_H)
+// 2.0
+#include <ruby/thread.h>
+#define WITHOUT_GVL(fn,a,ubf,b) rb_thread_call_without_gvl((fn),(a),(ubf),(b))
+#elif defined(HAVE_RB_THREAD_BLOCKING_REGION)
+ // 1.9
+typedef VALUE (*my_blocking_fn_t)(void*);
+#define WITHOUT_GVL(fn,a,ubf,b) rb_thread_blocking_region((my_blocking_fn_t)(fn),(a),(ubf),(b))
+#endif
+
+// struct sembuf { // found in sys/sem.h
+//   unsigned short sem_num; /* semaphore number */
+//   short sem_op; /* semaphore operation */
+//   short sem_flg; /* operation flags */
+// };
+
+typedef struct {
+  int successes;
+  int arr_length;
+  double errors[];
+} shared_cb_data;
+
+typedef struct {
+  //semaphore, shared memory data and pointer
+  key_t key;
+  size_t arr_max_size;
+  bool lock_triggered;
+  int semid;
+  int shmid;
+  shared_cb_data *shm_address;
+} semian_cb_data;
+
+static int system_max_semaphore_count;
+static const int kCBSemaphoreCount = 1; // # semaphores to be acquired
+static const int kCBTicketMax = 1;
+static const int kCBInitializeWaitTimeout = 5; /* seconds */
+static const int kCBIndexTicketLock = 0;
+static const int kCBInternalTimeout = 5; /* seconds */
+
+static struct sembuf decrement; // = { kCBIndexTicketLock, -1, SEM_UNDO};
+static struct sembuf increment; // = { kCBIndexTicketLock, 1, SEM_UNDO};
+
+static VALUE eInternal, eSyscall, eTimeout; // Semian errors
+
+static void semian_cb_data_mark(void *ptr);
+static void semian_cb_data_free(void *ptr);
+static size_t semian_cb_data_memsize(const void *ptr);
+static VALUE semian_cb_data_alloc(VALUE klass);
+static VALUE semian_cb_data_init(VALUE self, VALUE name, VALUE size, VALUE permissions);
+static VALUE semian_cb_data_clean(VALUE self);
+static void set_semaphore_permissions(int sem_id, int permissions);
+static void configure_tickets(int sem_id, int tickets, int should_initialize);
+static int create_semaphore(int key, int permissions, int *created);
+static VALUE semian_cb_data_acquire_semaphore (VALUE self, VALUE permissions);
+static VALUE semian_cb_data_delete_semaphore(VALUE self);
+static VALUE semian_cb_data_lock(VALUE self);
+static VALUE semian_cb_data_unlock(VALUE self);
+static void *semian_cb_data_lock_without_gvl(void *self);
+static void *semian_cb_data_unlock_without_gvl(void *self);
+static VALUE semian_cb_data_acquire_memory(VALUE self, VALUE permissions);
+static void semian_cb_data_delete_memory_inner (semian_cb_data *ptr);
+static VALUE semian_cb_data_delete_memory (VALUE self);
+static VALUE semian_cb_data_get_successes(VALUE self);
+static VALUE semian_cb_data_set_successes(VALUE self, VALUE num);
+static VALUE semian_cb_data_semid(VALUE self);
+static VALUE semian_cb_data_shmid(VALUE self);
+static VALUE semian_cb_data_array_at_index(VALUE self, VALUE idx);
+static VALUE semian_cb_data_array_set_index(VALUE self, VALUE idx, VALUE val);
+static VALUE semian_cb_data_array_length(VALUE self);
+static VALUE semian_cb_data_set_push_back(VALUE self, VALUE num);
+static VALUE semian_cb_data_set_pop_back(VALUE self);
+static VALUE semian_cb_data_set_push_front(VALUE self, VALUE num);
+static VALUE semian_cb_data_set_pop_front(VALUE self);
+
+// needed for TypedData_Make_Struct && TypedData_Get_Struct
+static const rb_data_type_t
+semian_cb_data_type = {
+  "semian_cb_data",
+  {
+    semian_cb_data_mark,
+    semian_cb_data_free,
+    semian_cb_data_memsize
+  },
+  NULL, NULL, RUBY_TYPED_FREE_IMMEDIATELY
+};
+
+/*
+ * Generate key
+ */
+static key_t
+generate_key(const char *name)
+{
+  union {
+    unsigned char str[SHA_DIGEST_LENGTH];
+    key_t key;
+  } digest;
+  SHA1((const unsigned char *) name, strlen(name), digest.str);
+  /* TODO: compile-time assertion that sizeof(key_t) > SHA_DIGEST_LENGTH */
+  return digest.key;
+}
+
+/*
+ * Log errors
+ */
+static void
+raise_semian_syscall_error(const char *syscall, int error_num)
+{
+  rb_raise(eSyscall, "%s failed, errno %d (%s)", syscall, error_num, strerror(error_num));
+}
+
+/*
+ * Functions that handle type and memory
+*/
+static void
+semian_cb_data_mark(void *ptr)
+{
+  /* noop */
+}
+
+static void
+semian_cb_data_free(void *ptr)
+{
+  semian_cb_data *data = (semian_cb_data *) ptr;
+
+
+  // Under normal circumstances, memory use should be in the order of bytes,
+  //   and shouldn't increase if the same key/id is used
+  //   so there is no need to call this unless certain all other semian processes are stopped
+  //   (also raises concurrency errors: "object allocation during garbage collection phase")
+
+  //semian_cb_data_delete_memory_inner (data);
+
+  xfree(data);
+}
+
+static size_t
+semian_cb_data_memsize(const void *ptr)
+{
+  return sizeof(semian_cb_data);
+}
+
+static VALUE
+semian_cb_data_alloc(VALUE klass)
+{
+  VALUE obj;
+  semian_cb_data *ptr;
+
+  obj = TypedData_Make_Struct(klass, semian_cb_data, &semian_cb_data_type, ptr);
+  return obj;
+}
+
+
+
+
+
+
+/*
+ * Init function exposed as ._initialize() that is delegated by .initialize()
+ */
+static VALUE
+semian_cb_data_init(VALUE self, VALUE id, VALUE size, VALUE permissions)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+
+  if (TYPE(id) != T_SYMBOL && TYPE(id) != T_STRING)
+    rb_raise(rb_eTypeError, "id must be a symbol or string");
+  if (TYPE(size) != T_FIXNUM /*|| TYPE(size) != T_BIGNUM*/)
+    rb_raise(rb_eTypeError, "expected integer for arr_max_size");
+  if (TYPE(permissions) != T_FIXNUM)
+    rb_raise(rb_eTypeError, "expected integer for permissions");
+
+  if (NUM2SIZET(size) <= 0)
+    rb_raise(rb_eArgError, "arr_max_size must be larger than 0");
+
+  const char *id_str = NULL;
+  if (TYPE(id) == T_SYMBOL) {
+    id_str = rb_id2name(rb_to_id(id));
+  } else if (TYPE(id) == T_STRING) {
+    id_str = RSTRING_PTR(id);
+  }
+  ptr->key = generate_key(id_str);
+  //rb_warn("converted name %s to key %d", id_str, ptr->key);
+
+  // Guarantee arr_max_size >=1 or error thrown
+  ptr->arr_max_size = NUM2SIZET(size);
+
+  // id's default to -1
+  ptr->semid = -1;
+  ptr->shmid = -1;
+  // addresses default to NULL
+  ptr->shm_address = 0;
+  ptr->lock_triggered = false;
+
+  semian_cb_data_acquire_semaphore(self, permissions);
+  semian_cb_data_acquire_memory(self, permissions);
+
+  return self;
+}
+
+static VALUE
+semian_cb_data_clean(VALUE self)
+{
+  semian_cb_data_delete_memory(self);
+  semian_cb_data_delete_semaphore(self);
+  return self;
+}
+
+
+
+/*
+ * Functions set_semaphore_permissions, configure_tickets, create_semaphore
+ * are taken from semian.c with extra code removed
+ */
+static void
+set_semaphore_permissions(int sem_id, int permissions)
+{
+  union semun sem_opts;
+  struct semid_ds stat_buf;
+
+  sem_opts.buf = &stat_buf;
+  semctl(sem_id, 0, IPC_STAT, sem_opts);
+  if ((stat_buf.sem_perm.mode & 0xfff) != permissions) {
+    stat_buf.sem_perm.mode &= ~0xfff;
+    stat_buf.sem_perm.mode |= permissions;
+    semctl(sem_id, 0, IPC_SET, sem_opts);
+  }
+}
+
+
+static void
+configure_tickets(int sem_id, int tickets, int should_initialize)
+{
+  struct timeval start_time, cur_time;
+
+  if (should_initialize) {
+    if (-1 == semctl(sem_id, 0, SETVAL, kCBTicketMax)) {
+      rb_warn("semctl: failed to set semaphore with semid %d, position 0 to %d", sem_id, 1);
+      raise_semian_syscall_error("semctl()", errno);
+    } else {
+      //rb_warn("semctl: set semaphore with semid %d, position 0 to %d", sem_id, 1);
+    }
+  } else if (tickets > 0) {
+    // it's possible that we haven't actually initialized the
+    // semaphore structure yet - wait a bit in that case
+    int ret;
+    if (0 == (ret = semctl(sem_id, 0, GETVAL))) {
+      gettimeofday(&start_time, NULL);
+      while (0 == (ret = semctl(sem_id, 0, GETVAL))) { // loop while value == 0
+        usleep(10000); /* 10ms */
+        gettimeofday(&cur_time, NULL);
+        if ((cur_time.tv_sec - start_time.tv_sec) > kCBInitializeWaitTimeout) {
+          rb_raise(eInternal, "timeout waiting for semaphore initialization");
+        }
+      }
+      if (-1 == ret) {
+        rb_raise(eInternal, "error getting max ticket count, errno: %d (%s)", errno, strerror(errno));
+      }
+    }
+
+    // Rest of the function (originally from semian.c) was removed since it isn't needed
+  }
+}
+
+static int
+create_semaphore(int key, int permissions, int *created)
+{
+  int semid = 0;
+  int flags = 0;
+
+  *created = 0;
+  flags = IPC_EXCL | IPC_CREAT | permissions;
+
+  semid = semget(key, kCBSemaphoreCount, flags);
+  if (semid >= 0) {
+    *created = 1;
+    //rb_warn("semget: received %d semaphore(s) with key %d, semid %d", kCBSemaphoreCount, key, semid);
+  } else if (semid == -1 && errno == EEXIST) {
+    flags &= ~IPC_EXCL;
+    semid = semget(key, kCBSemaphoreCount, flags);
+    //rb_warn("semget: retrieved existing semaphore with key %d, semid %d", key, semid);
+  }
+  return semid;
+}
+
+
+
+
+/*
+ * Create or acquire previously made semaphore
+ */
+
+static VALUE
+semian_cb_data_acquire_semaphore (VALUE self, VALUE permissions)
+{
+  // Function flow, semaphore creation methods are
+  //   borrowed from semian.c since they have been previously tested
+
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+
+  if (TYPE(permissions) != T_FIXNUM)
+    rb_raise(rb_eTypeError, "expected integer for permissions");
+
+  // bool for initializing (configure_tickets) or not
+  int created = 0;
+  key_t key = ptr->key;
+  int semid = create_semaphore(key, FIX2LONG(permissions), &created);
+  if (-1 == semid) {
+    raise_semian_syscall_error("semget()", errno);
+  }
+  ptr->semid = semid;
+
+  // initialize to 1 and set permissions
+  configure_tickets(ptr->semid, kCBTicketMax, created);
+  set_semaphore_permissions(ptr->semid, FIX2LONG(permissions));
+
+  return self;
+}
+
+
+static VALUE
+semian_cb_data_delete_semaphore(VALUE self)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+  if (-1 == ptr->semid) // do nothing if semaphore not acquired
+    return self;
+
+  if (-1 == semctl(ptr->semid, 0, IPC_RMID)) {
+    if (EIDRM == errno) {
+      rb_warn("semctl: failed to delete semaphore set with semid %d: already removed", ptr->semid);
+      ptr->semid = -1;
+    } else {
+      rb_warn("semctl: failed to remove semaphore with semid %d, errno %d (%s)",ptr->semid, errno, strerror(errno));
+    }
+  } else {
+    //rb_warn("semctl: semaphore set with semid %d deleted", ptr->semid);
+    ptr->semid = -1;
+  }
+  return self;
+}
+
+
+
+
+/*
+ * semian_cb_data_lock/unlock and associated functions decrement/increment semaphore
+ */
+
+static VALUE
+semian_cb_data_lock(VALUE self)
+{
+  return (VALUE) WITHOUT_GVL(semian_cb_data_lock_without_gvl, (void *)self, RUBY_UBF_IO, NULL);
+}
+
+static void *
+semian_cb_data_lock_without_gvl(void *self)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct((VALUE)self, semian_cb_data, &semian_cb_data_type, ptr);
+  if (ptr->lock_triggered)
+    return (void *)Qtrue;
+  if (-1 == ptr->semid){
+    rb_raise(eInternal, "semid not set, errno %d: (%s)", errno, strerror(errno));
+    return (void *)Qfalse;
+  }
+  VALUE retval;
+
+  struct timespec ts = { 0 };
+  ts.tv_sec = kCBInternalTimeout;
+
+  if (-1 == semtimedop(ptr->semid,&decrement,1, &ts)) {
+    rb_raise(eInternal, "error with semop locking,  %d: (%s)", errno, strerror(errno));
+    retval=Qfalse;
+  } else
+    retval=Qtrue;
+
+  ptr->lock_triggered = true;
+  //rb_warn("semop: lock success");
+  return (void *)retval;
+}
+
+static VALUE
+semian_cb_data_unlock(VALUE self)
+{
+  return (VALUE) WITHOUT_GVL(semian_cb_data_unlock_without_gvl, (void *)self, RUBY_UBF_IO, NULL);
+}
+
+static void *
+semian_cb_data_unlock_without_gvl(void *self)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct((VALUE)self, semian_cb_data, &semian_cb_data_type, ptr);
+  if (!(ptr->lock_triggered))
+    return (void *)Qtrue;
+  if (-1 == ptr->semid){
+    rb_raise(eInternal, "semid not set, errno %d: (%s)", errno, strerror(errno));
+    return (void *)Qfalse;
+  }
+  VALUE retval;
+
+  struct timespec ts = { 0 };
+  ts.tv_sec = kCBInternalTimeout;
+
+  if (-1 == semtimedop(ptr->semid,&increment,1 , &ts)) {
+    rb_raise(eInternal, "error with semop unlocking, errno: %d (%s)", errno, strerror(errno));
+    retval=Qfalse;
+  } else
+    retval=Qtrue;
+
+  ptr->lock_triggered = false;
+  //rb_warn("semop unlock success");
+  return (void *)retval;
+}
+
+
+
+
+/*
+  Acquire memory by getting shmid, and then attaching it to a memory location,
+    requires semaphore for locking/unlocking to be setup
+*/
+static VALUE
+semian_cb_data_acquire_memory(VALUE self, VALUE permissions)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+
+  if (-1 == ptr->semid){
+    rb_raise(eInternal, "semid not set, errno %d: (%s)", errno, strerror(errno));
+    return self;
+  }
+  if (TYPE(permissions) != T_FIXNUM)
+    rb_raise(rb_eTypeError, "expected integer for permissions");
+
+  if (!semian_cb_data_lock(self))
+    return self;
+
+  key_t key = ptr->key;
+  if (-1 == (ptr->shmid = shmget( key,
+                                  2*sizeof(int) + ptr->arr_max_size * sizeof(double),
+                                  IPC_CREAT | IPC_EXCL | FIX2LONG(permissions)))) {
+    if (errno == EEXIST)
+      ptr->shmid = shmget(key, ptr->arr_max_size, IPC_CREAT);
+  }
+  if (-1 == ptr->shmid) {
+    rb_raise(eSyscall, "shmget() failed to acquire a memory shmid with key %d, size %zu, errno %d (%s)", key, ptr->arr_max_size, errno, strerror(errno));
+  } else {
+    //rb_warn("shmget: successfully got memory id with key %d, shmid %d, size %zu", key, ptr->shmid, ptr->arr_max_size);
+  }
+
+  if (0 == ptr->shm_address) {
+    ptr->shm_address = shmat(ptr->shmid, (void *)0, 0);
+    if (((void*)-1) == ptr->shm_address) {
+      rb_raise(eSyscall, "shmat() failed to attach memory with shmid %d, size %zu, errno %d (%s)", ptr->shmid, ptr->arr_max_size, errno, strerror(errno));
+      ptr->shm_address = 0;
+    } else {
+      //rb_warn("shmat: successfully attached shmid %d to %p", ptr->shmid, ptr->shm_address);
+      shared_cb_data *data = ptr->shm_address;
+      data->successes = 0;
+      data->arr_length = 0;
+      int i=0;
+      for (; i< data->arr_length; ++i)
+        data->errors[i]=0;
+
+    }
+  }
+
+  semian_cb_data_unlock(self);
+  return self;
+}
+
+static void
+semian_cb_data_delete_memory_inner (semian_cb_data *ptr)
+{
+  if (0 != ptr->shm_address){
+    if (-1 == shmdt(ptr->shm_address)) {
+      rb_raise(eSyscall,"shmdt: no attached memory at %p, errno %d (%s)", ptr->shm_address, errno, strerror(errno));
+    } else {
+      rb_warn("shmdt: successfully detached memory at %p", ptr->shm_address);
+    }
+    ptr->shm_address = 0;
+  }
+
+  if (-1 != ptr->shmid) {
+    // Once IPC_RMID is set, no new attaches can be made
+    if (-1 == shmctl(ptr->shmid, IPC_RMID, 0)) {
+      if (errno == EINVAL) {
+        ptr->shmid = -1;
+      } {
+        rb_raise(eSyscall,"shmctl: error removing memory with shmid %d, errno %d (%s)", ptr->shmid, errno, strerror(errno));
+      }
+    } else {
+      ptr->shmid = -1;
+    }
+  }
+}
+
+
+static VALUE
+semian_cb_data_delete_memory (VALUE self)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+
+  if (!semian_cb_data_lock(self))
+    return self;
+
+  semian_cb_data_delete_memory_inner(ptr);
+
+  semian_cb_data_unlock(self);
+  return self;
+}
+
+
+
+
+
+/*
+ * Below are methods for successes, semid, shmid, and array pop, push, peek at front and back
+ *  and clear, length
+ */
+
+static VALUE
+semian_cb_data_get_successes(VALUE self)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+
+  // check shared memory for NULL
+  if (0 == ptr->shm_address)
+    return Qnil;
+
+
+  int successes = ptr->shm_address->successes;
+
+  semian_cb_data_unlock(self);
+  return INT2NUM(successes);
+}
+
+static VALUE
+semian_cb_data_set_successes(VALUE self, VALUE num)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+
+  if (0 == ptr->shm_address)
+    return Qnil;
+
+  if (!semian_cb_data_lock(self))
+    return Qnil;
+
+  if (TYPE(num) != T_FIXNUM && TYPE(num) != T_FLOAT/*|| TYPE(size) != T_BIGNUM*/)
+    return Qnil;
+
+  ptr->shm_address->successes = NUM2INT(num);
+
+  semian_cb_data_unlock(self);
+  return num;
+}
+
+
+static VALUE
+semian_cb_data_semid(VALUE self)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+  return INT2NUM(ptr->semid);
+}
+static VALUE
+semian_cb_data_shmid(VALUE self)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+  return INT2NUM(ptr->shmid);
+}
+
+static VALUE
+semian_cb_data_array_at_index(VALUE self, VALUE idx)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+
+  if (0 == ptr->shm_address)
+    return Qnil;
+
+  if (TYPE(idx) != T_FIXNUM && TYPE(idx) != T_FLOAT/*|| TYPE(size) != T_BIGNUM*/)
+    return Qnil;
+
+  int index = NUM2INT(idx);
+
+  if (index <0 || index >= ptr->arr_max_size) {
+    return Qnil;
+  }
+
+  if (!semian_cb_data_lock(self))
+    return Qnil;
+  shared_cb_data *data = ptr->shm_address;
+  VALUE retval = index < (data->arr_length) ? DBL2NUM(data->errors[index]) : Qnil;
+
+  semian_cb_data_unlock(self);
+  return retval;
+
+}
+
+static VALUE
+semian_cb_data_array_set_index(VALUE self, VALUE idx, VALUE val)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+
+  // check shared memory for NULL
+  if (0 == ptr->shm_address)
+    return Qnil;
+
+  if (TYPE(idx) != T_FIXNUM && TYPE(idx) != T_FLOAT/*|| TYPE(size) != T_BIGNUM*/)
+    return Qnil;
+  if (TYPE(val) != T_FIXNUM && TYPE(val) != T_FLOAT/*|| TYPE(size) != T_BIGNUM*/)
+    return Qnil;
+
+  int index = NUM2INT(idx);
+  double value = NUM2DBL(val);
+
+  if (index <0 || index >= ptr->arr_max_size) {
+    return Qnil;
+  }
+
+  if (!semian_cb_data_lock(self)){
+    return Qnil;
+  }
+
+  ptr->shm_address->errors[index] = value;
+  ptr->shm_address->arr_length = index+1;
+
+  semian_cb_data_unlock(self);
+  return val;
+
+}
+
+static VALUE
+semian_cb_data_array_length(VALUE self)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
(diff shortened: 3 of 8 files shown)


.gitignore
```diff
@@ -1,6 +1,6 @@
 /.bundle/
-/lib/semian/*.so
-/lib/semian/*.bundle
+/lib/semian*/*.so
+/lib/semian*/*.bundle
```
byroot: Wouldn't `/lib/**/*.so` work here?

lib/semian/circuit_breaker.rb
```diff
@@ -43,13 +49,16 @@ def mark_failed(error)
 
     def mark_success
       return unless half_open?
-      @successes += 1
+      #@successes += 1
```
byroot: Commenting code out is usually a bad practice. Any reason why you'd comment it instead of just plain delete it?

kyewei: I had left it to keep the meaning clear, since writing a custom array container that only provides certain functionality (`:push`, `:pop`) while adding others that aren't normal array operations (`:successes`, etc) may leave the code confusing to read. If you think the meaning is still clear, then I'll remove it.

Rakefile
```diff
@@ -24,6 +24,10 @@ if Semian.sysv_semaphores_supported?
     ext.ext_dir = 'ext/semian'
     ext.lib_dir = 'lib/semian'
   end
+  Rake::ExtensionTask.new('semian_cb_data', GEMSPEC) do |ext|
```
csfrancis: I think I misunderstood/wasn't clear earlier. I think this code belongs in the semian .so, but not in the same source file.

lib/semian/circuit_breaker.rb
byroot: Yes it's still clear. Especially if rewritten as `@shared_circuit_breaker_data.successes += 1`

Rakefile
Sirupsen: Agree with Scott here. Can we just have a single `extconf.rb` generating one shared object? Perhaps you can get some inspiration from other larger multi-file Ruby C extensions on how to structure. [1](https://github.com/eventmachine/eventmachine/tree/master/ext) [2](https://github.com/sparklemotion/nokogiri/tree/master/ext) [3](https://github.com/tenderlove/psych/tree/master/ext)

lib/semian/circuit_breaker_shared_data.rb
```diff
@@ -0,0 +1,70 @@
+module Semian
+  class CircuitBreakerSharedData #:nodoc:
+    def initialize(name, arr_max_size, permissions)
+      _initialize(name, arr_max_size, permissions) if respond_to?(:_initialize)
+    end
+
+    # For anyone consulting this, the array stores floats. Use Time.at(_float_here_) to convert to time
+
+    def successes
```
Sirupsen: Any reason to not fall back to an in-memory implementation?

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+    semian_cb_data_memsize
+  },
+  NULL, NULL, RUBY_TYPED_FREE_IMMEDIATELY
+};
+
+/*
+ * Generate key
+ */
+static key_t
+generate_key(const char *name)
```
Sirupsen: If you compile them into the same `.so`, you shouldn't need to duplicate this.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+
+
+/*
+ * semian_cb_data_lock/unlock and associated functions decrement/increment semaphore
+ */
+
+static VALUE
+semian_cb_data_lock(VALUE self)
+{
+  return (VALUE) WITHOUT_GVL(semian_cb_data_lock_without_gvl, (void *)self, RUBY_UBF_IO, NULL);
```
Sirupsen: We'd like to move away from timeouts for OS X compatibility (\cc @charliesome), this came up [here](https://github.com/Shopify/semian/pull/51#discussion_r39794789). Can we implement a fallback for `semtimedop` with an `#ifdef` possibly using Ruby's built in sleep so it doesn't block the VM?

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+#include <stdio.h>
+
+#include <errno.h>
+#include <sys/types.h>
+#include <sys/ipc.h>
+#include <sys/sem.h>
+#include <sys/shm.h>
+#include <sys/time.h>
+
+#include <openssl/sha.h>
```
Sirupsen: Should we change this to `md5` in a follow-up to kill the OpenSSL dependency?

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  //semaphore, shared memory data and pointer
+  key_t key;
+  size_t arr_max_size;
+  bool lock_triggered;
+  int semid;
+  int shmid;
+  shared_cb_data *shm_address;
+} semian_cb_data;
+
+static int system_max_semaphore_count;
```
Sirupsen: This is never used.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+        rb_raise(eInternal, "error getting max ticket count, errno: %d (%s)", errno, strerror(errno));
+      }
+    }
+
+    // Rest of the function (originally from semian.c) was removed since it isn't needed
+  }
+}
+
+static int
+create_semaphore(int key, int permissions, int *created)
```
Sirupsen: Seems like we can re-use the implementation from `semian.c`.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  if ((stat_buf.sem_perm.mode & 0xfff) != permissions) {
+    stat_buf.sem_perm.mode &= ~0xfff;
+    stat_buf.sem_perm.mode |= permissions;
+    semctl(sem_id, 0, IPC_SET, sem_opts);
+  }
+}
+
+
+static void
+configure_tickets(int sem_id, int tickets, int should_initialize)
```
Sirupsen: Ditto with borrowing from `semian.c` with a flag to trigger the fancy 3-semaphore delta dance.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+}
+
+
+
+/*
+ * Functions set_semaphore_permissions, configure_tickets, create_semaphore
+ * are taken from semian.c with extra code removed
+ */
+static void
+set_semaphore_permissions(int sem_id, int permissions)
```
Sirupsen: `semian.c`

Sirupsen:
This needs some tests around (at least):

* Two workers in two different processes sharing the CB (and the different scenarios that follow, e.g. circuit break trips, etc)
* Forcefully killing a worker that might have the semaphore releases it (fuzzy)

Other than that I'm assuming you're borrowing from the existing test suite?

Sirupsen:
> Shared memory is not explicitly marked for deletion from the C side

Agree. We also don't need to use `IPC_RMID` for the same reason.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+static VALUE
+semian_cb_data_init(VALUE self, VALUE id, VALUE size, VALUE permissions)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+
+  if (TYPE(id) != T_SYMBOL && TYPE(id) != T_STRING)
+    rb_raise(rb_eTypeError, "id must be a symbol or string");
+  if (TYPE(size) != T_FIXNUM /*|| TYPE(size) != T_BIGNUM*/)
+    rb_raise(rb_eTypeError, "expected integer for arr_max_size");
```
Sirupsen: The name `arr_max_size` is the internal name, it won't make much sense for the Ruby caller.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  if (TYPE(permissions) != T_FIXNUM)
+    rb_raise(rb_eTypeError, "expected integer for permissions");
+
+  if (NUM2SIZET(size) <= 0)
+    rb_raise(rb_eArgError, "arr_max_size must be larger than 0");
```
Sirupsen: ditto

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+ */
+static VALUE
+semian_cb_data_init(VALUE self, VALUE id, VALUE size, VALUE permissions)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+
+  if (TYPE(id) != T_SYMBOL && TYPE(id) != T_STRING)
+    rb_raise(rb_eTypeError, "id must be a symbol or string");
+  if (TYPE(size) != T_FIXNUM /*|| TYPE(size) != T_BIGNUM*/)
```
Sirupsen: Why did you comment the `T_BIGNUM` typecheck out? I don't ever anticipate anyone will have a circuit breaker of that size, so I think that can just be removed.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+    rb_raise(rb_eArgError, "arr_max_size must be larger than 0");
+
+  const char *id_str = NULL;
+  if (TYPE(id) == T_SYMBOL) {
+    id_str = rb_id2name(rb_to_id(id));
+  } else if (TYPE(id) == T_STRING) {
+    id_str = RSTRING_PTR(id);
+  }
+  ptr->key = generate_key(id_str);
+  //rb_warn("converted name %s to key %d", id_str, ptr->key);
```
Sirupsen: Remove this?

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+
+
+
+
+/*
+ * Create or acquire previously made semaphore
+ */
+
+static VALUE
+semian_cb_data_acquire_semaphore (VALUE self, VALUE permissions)
```
Sirupsen: Use `semian.c`

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  ptr->lock_triggered = false;
+
+  semian_cb_data_acquire_semaphore(self, permissions);
+  semian_cb_data_acquire_memory(self, permissions);
+
+  return self;
+}
+
+static VALUE
+semian_cb_data_clean(VALUE self)
```
Sirupsen: When and where do you anticipate calling this? In tests? Calling it `#clear` would be more consistent with Ruby idioms.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  return self;
+}
+
+
+static VALUE
+semian_cb_data_delete_semaphore(VALUE self)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+  if (-1 == ptr->semid) // do nothing if semaphore not acquired
```
Sirupsen: Return `false` maybe?

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  semian_cb_data_acquire_memory(self, permissions);
+
+  return self;
+}
+
+static VALUE
+semian_cb_data_clean(VALUE self)
+{
+  semian_cb_data_delete_memory(self);
+  semian_cb_data_delete_semaphore(self);
```
Sirupsen: Would make sense if this returned `true`/`false`.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+  if (-1 == ptr->semid) // do nothing if semaphore not acquired
+    return self;
+
+  if (-1 == semctl(ptr->semid, 0, IPC_RMID)) {
+    if (EIDRM == errno) {
+      rb_warn("semctl: failed to delete semaphore set with semid %d: already removed", ptr->semid);
+      ptr->semid = -1;
+    } else {
+      rb_warn("semctl: failed to remove semaphore with semid %d, errno %d (%s)",ptr->semid, errno, strerror(errno));
```
Sirupsen: `raise`?

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+ * semian_cb_data_lock/unlock and associated functions decrement/increment semaphore
+ */
+
+static VALUE
+semian_cb_data_lock(VALUE self)
+{
+  return (VALUE) WITHOUT_GVL(semian_cb_data_lock_without_gvl, (void *)self, RUBY_UBF_IO, NULL);
+}
+
+static void *
```
Sirupsen: Why not just return `VALUE`?

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+    rb_raise(eInternal, "semid not set, errno %d: (%s)", errno, strerror(errno));
+    return (void *)Qfalse;
+  }
+  VALUE retval;
+
+  struct timespec ts = { 0 };
+  ts.tv_sec = kCBInternalTimeout;
+
+  if (-1 == semtimedop(ptr->semid,&decrement,1, &ts)) {
+    rb_raise(eInternal, "error with semop locking,  %d: (%s)", errno, strerror(errno));
```
Sirupsen: Error could be a bit more desciptive: `Error acquiring semaphore to mutate circuit breaker structure: %d (%s)`

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+}
+
+static VALUE
+semian_cb_data_unlock(VALUE self)
+{
+  return (VALUE) WITHOUT_GVL(semian_cb_data_unlock_without_gvl, (void *)self, RUBY_UBF_IO, NULL);
+}
+
+static void *
+semian_cb_data_unlock_without_gvl(void *self)
```
Sirupsen: These GVL-free acquiring functions can be shared with `semian.c` and, as mentioned earlier, need a fallback for platforms that don't have `semtimedop`. All the GVL stuff can then be put in the `semtimedop` platform-safe abstraction.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  ts.tv_sec = kCBInternalTimeout;
+
+  if (-1 == semtimedop(ptr->semid,&increment,1 , &ts)) {
+    rb_raise(eInternal, "error with semop unlocking, errno: %d (%s)", errno, strerror(errno));
+    retval=Qfalse;
+  } else
+    retval=Qtrue;
+
+  ptr->lock_triggered = false;
+  //rb_warn("semop unlock success");
```
Sirupsen: You can probably remove all those warnings.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  return (void *)retval;
+}
+
+static VALUE
+semian_cb_data_unlock(VALUE self)
+{
+  return (VALUE) WITHOUT_GVL(semian_cb_data_unlock_without_gvl, (void *)self, RUBY_UBF_IO, NULL);
+}
+
+static void *
```
Sirupsen: Again, why this `void *` pattern instead of just returning a Ruby `VALUE`?

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+
+  if (-1 == ptr->semid){
+    rb_raise(eInternal, "semid not set, errno %d: (%s)", errno, strerror(errno));
+    return self;
+  }
+  if (TYPE(permissions) != T_FIXNUM)
+    rb_raise(rb_eTypeError, "expected integer for permissions");
+
+  if (!semian_cb_data_lock(self))
+    return self;
```
Sirupsen: You should probably return something more useful than `self`.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+
+  if (-1 == ptr->semid){
+    rb_raise(eInternal, "semid not set, errno %d: (%s)", errno, strerror(errno));
+    return self;
+  }
+  if (TYPE(permissions) != T_FIXNUM)
+    rb_raise(rb_eTypeError, "expected integer for permissions");
+
+  if (!semian_cb_data_lock(self))
```
Sirupsen: Do you even need to acquire the semaphore here? Isn't `IPC_EXCL` enough to guarantee only one of them will initialize it?

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+    return self;
+
+  key_t key = ptr->key;
+  if (-1 == (ptr->shmid = shmget( key,
+                                  2*sizeof(int) + ptr->arr_max_size * sizeof(double),
+                                  IPC_CREAT | IPC_EXCL | FIX2LONG(permissions)))) {
```
Sirupsen: What if the size of the structure changes? This could happen if you resize the sliding window.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+    if (errno == EEXIST)
+      ptr->shmid = shmget(key, ptr->arr_max_size, IPC_CREAT);
```
Sirupsen: Why pass `IPC_CREAT` if it already exists?

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  }
+  if (-1 == ptr->shmid) {
+    rb_raise(eSyscall, "shmget() failed to acquire a memory shmid with key %d, size %zu, errno %d (%s)", key, ptr->arr_max_size, errno, strerror(errno));
+  } else {
+    //rb_warn("shmget: successfully got memory id with key %d, shmid %d, size %zu", key, ptr->shmid, ptr->arr_max_size);
```
Sirupsen: Can remove this clause.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  }
+
+  if (0 == ptr->shm_address) {
```
Sirupsen: Why would this ever be set? You're allocating an empty pointer above on the stack and only setting the `id`.

Sirupsen:
In a follow-up we'll have to figure out how to make this part of [Instrumentable](https://github.com/Shopify/semian/blob/master/lib/semian/instrumentable.rb) so we can monitor the circuit breaker stats across all hosts.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+}
+
+static VALUE
+semian_cb_data_set_push_back(VALUE self, VALUE num)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+  if (0 == ptr->shm_address)
+    return Qnil;
+  if (TYPE(num) != T_FIXNUM && TYPE(num) != T_FLOAT/*|| TYPE(size) != T_BIGNUM*/)
```
Sirupsen: Why do we push floats and not an integer (i.e. ms since epoch)? I'm not a big fan of using floats.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  if (-1 == ptr->shmid) {
+    rb_raise(eSyscall, "shmget() failed to acquire a memory shmid with key %d, size %zu, errno %d (%s)", key, ptr->arr_max_size, errno, strerror(errno));
+  } else {
+    //rb_warn("shmget: successfully got memory id with key %d, shmid %d, size %zu", key, ptr->shmid, ptr->arr_max_size);
+  }
+
+  if (0 == ptr->shm_address) {
+    ptr->shm_address = shmat(ptr->shmid, (void *)0, 0);
+    if (((void*)-1) == ptr->shm_address) {
+      rb_raise(eSyscall, "shmat() failed to attach memory with shmid %d, size %zu, errno %d (%s)", ptr->shmid, ptr->arr_max_size, errno, strerror(errno));
```
Sirupsen: Would this actually `unlock` the semaphore?

ext/semian_cb_data/semian_cb_data.c
kyewei: `WITHOUT_GVL` requires a method signature of `void *(*fn)(void *)`

ext/semian_cb_data/semian_cb_data.c
kyewei: `IPC_CREAT`: Create the segment if it doesn't already exist in the kernel.

`IPC_EXCL`: When used with `IPC_CREAT`, fail if segment already exists.

ext/semian_cb_data/semian_cb_data.c
kyewei: There's a `#clear` for clearing the array already. Maybe something like `#destroy`?

byroot:
I'd add another comment but more general: Please avoid abbreviations, like `cb`, just use `circuit_breaker` instead.

ext/semian_cb_data/semian_cb_data.c
kyewei: Should the constants be reused as well? Or abstracting that and passing that in as arguments?

lib/semian/circuit_breaker.rb
```diff
@@ -2,11 +2,17 @@ module Semian
...
 
-    def initialize(exceptions:, success_threshold:, error_threshold:, error_timeout:)
+    def initialize(name, exceptions:, success_threshold:, error_threshold:, error_timeout:, permissions: 0660)
+      @name = "#{name}_cb"
       @success_count_threshold = success_threshold
       @error_count_threshold = error_threshold
       @error_timeout = error_timeout
       @exceptions = exceptions
+
+      @shared_circuit_breaker_data = ::Semian::CircuitBreakerSharedData.new(
```
byroot: I'm not a fan of this naming. Why not just `@state` or `@shared_state`. Same with the class name.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+      ptr->shm_address = 0;
+    } else {
+      //rb_warn("shmat: successfully attached shmid %d to %p", ptr->shmid, ptr->shm_address);
+      shared_cb_data *data = ptr->shm_address;
+      data->successes = 0;
+      data->arr_length = 0;
+      int i=0;
+      for (; i< data->arr_length; ++i)
+        data->errors[i]=0;
```
Sirupsen: This means that all workers will reset the circuit breaker state when they come up. Workers disappear and get killed often and deploys also shouldn't trigger the reset of the circuit breakers.

ext/semian_cb_data/semian_cb_data.c
kyewei: Unless I'm mistaken, the macro `WITHOUT_GVL` signature requires `void *`

ext/semian_cb_data/semian_cb_data.c
Sirupsen: Right, but this code tries to either create it exclusively or grab the existing handle. When you get `EEXISTS` you don't need to pass a flag to create it, you know it's already there (unless it was deleted in the interim but I wouldn't worry about that and it'd be caught below).

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+
+  semian_cb_data_unlock(self);
+  return self;
+}
+
+static void
+semian_cb_data_delete_memory_inner (semian_cb_data *ptr)
+{
+  if (0 != ptr->shm_address){
+    if (-1 == shmdt(ptr->shm_address)) {
```
Sirupsen: I'd prefer the positive case first, easier to read.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+    ptr->shm_address = 0;
+  }
+
+  if (-1 != ptr->shmid) {
+    // Once IPC_RMID is set, no new attaches can be made
+    if (-1 == shmctl(ptr->shmid, IPC_RMID, 0)) {
+      if (errno == EINVAL) {
+        ptr->shmid = -1;
+      } {
+        rb_raise(eSyscall,"shmctl: error removing memory with shmid %d, errno %d (%s)", ptr->shmid, errno, strerror(errno));
```
Sirupsen: It doesn't quite mark the memory, it starts an async action. It's more of a marking for deletion.

ext/semian_cb_data/semian_cb_data.c
kyewei: So... is it the situation where size is changed in code, semian is restarted, and then a different memory size returned by `semget` requires handling? I'll look into it.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+
+  // check shared memory for NULL
+  if (0 == ptr->shm_address)
+    return Qnil;
+
+
+  int successes = ptr->shm_address->successes;
+
+  semian_cb_data_unlock(self);
```
Sirupsen: You don't seem to actually grab the lock anywhere in this function?

ext/semian_cb_data/semian_cb_data.c
kyewei: It was added there for reusability (i.e. deleting and re-acquiring the memory pointer). You're right, it might not be needed anymore.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+
+  if (!semian_cb_data_lock(self))
+    return Qnil;
+
+  if (TYPE(num) != T_FIXNUM && TYPE(num) != T_FLOAT/*|| TYPE(size) != T_BIGNUM*/)
+    return Qnil;
+
+  ptr->shm_address->successes = NUM2INT(num);
+
+  semian_cb_data_unlock(self);
```
Sirupsen: This could fail, no?

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  shared_cb_data *data = ptr->shm_address;
+  VALUE retval = index < (data->arr_length) ? DBL2NUM(data->errors[index]) : Qnil;
+
+  semian_cb_data_unlock(self);
+  return retval;
+
+}
+
+static VALUE
+semian_cb_data_array_set_index(VALUE self, VALUE idx, VALUE val)
```
Sirupsen: Why do you ever need to access an index of the array?

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+static VALUE
+semian_cb_data_shmid(VALUE self)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+  return INT2NUM(ptr->shmid);
+}
+
+static VALUE
+semian_cb_data_array_at_index(VALUE self, VALUE idx)
```
Sirupsen: Similarly, why do you ever need to do this?

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+semian_cb_data_array_length(VALUE self)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+  if (0 == ptr->shm_address)
+    return Qnil;
+  if (!semian_cb_data_lock(self))
+    return Qnil;
+  int arr_length =ptr->shm_address->arr_length;
+  semian_cb_data_unlock(self);
```
Sirupsen: no error handling

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  if (data->arr_length == ptr->arr_max_size) {
+    int i;
+    for (i=1; i< ptr->arr_max_size; ++i){
+      data->errors[i-1] = data->errors[i];
+    }
+    --(data->arr_length);
+  }
+  data->errors[(data->arr_length)] = NUM2DBL(num);
+  ++(data->arr_length);
+  semian_cb_data_unlock(self);
```
Sirupsen: error handling

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  if (!semian_cb_data_lock(self))
+    return Qnil;
+
+  shared_cb_data *data = ptr->shm_address;
+  if (data->arr_length == ptr->arr_max_size) {
+    int i;
+    for (i=1; i< ptr->arr_max_size; ++i){
+      data->errors[i-1] = data->errors[i];
+    }
+    --(data->arr_length);
```
Sirupsen: Why decrement and then increment, instead of just doing -1 when setting the new push_back?

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  }
+  data->errors[(data->arr_length)] = NUM2DBL(num);
+  ++(data->arr_length);
+  semian_cb_data_unlock(self);
+  return self;
+}
+
+static VALUE
+semian_cb_data_set_pop_back(VALUE self)
```
Sirupsen: Why do we ever need to pop from the front?

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  if (!semian_cb_data_lock(self))
+    return Qnil;
+
+  VALUE retval;
+  shared_cb_data *data = ptr->shm_address;
+  if (0 >= data->arr_length)
+    retval = Qnil;
+  else {
+    retval = DBL2NUM(data->errors[0]);
+    int i=0;
```
Sirupsen: Any reason you're not doing this decl. in the `for`? AFAIK in C99 you declare variables at start of function, in modern C you declare them inline. I don't really know which one is common in the Rubby world @csfrancis would know

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+      data->errors[i]=data->errors[i+1];
+    --(data->arr_length);
+  }
+
+  semian_cb_data_unlock(self);
+  return retval;
+}
+
+static VALUE
+semian_cb_data_set_push_front(VALUE self, VALUE num)
```
Sirupsen: When do we need to push to the front?

ext/semian_cb_data/semian_cb_data.c
kyewei: `SEM_UNDO` built into the decrementing `lock()` makes it safe.

lib/semian/circuit_breaker_shared_data.rb
```diff
@@ -0,0 +1,70 @@
+module Semian
+  class CircuitBreakerSharedData #:nodoc:
```
Sirupsen: I'd consider renaming this data structure to `SlidingWindow` and then make this the fallback implementation where all the counts are just in memory for the current VM. If `SysV` is installed, the `SlidingWindow` implementation is overridden by the SysV one, but could also be POSIX, Java, whatever.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  rb_define_method(cCircuitBreakerSharedData, "size", semian_cb_data_array_length, 0);
+  rb_define_method(cCircuitBreakerSharedData, "count", semian_cb_data_array_length, 0);
+  rb_define_method(cCircuitBreakerSharedData, "<<", semian_cb_data_set_push_back, 1);
+  rb_define_method(cCircuitBreakerSharedData, "push", semian_cb_data_set_push_back, 1);
+  rb_define_method(cCircuitBreakerSharedData, "pop", semian_cb_data_set_pop_back, 0);
+  rb_define_method(cCircuitBreakerSharedData, "shift", semian_cb_data_set_pop_front, 0);
+  rb_define_method(cCircuitBreakerSharedData, "unshift", semian_cb_data_set_push_front, 1);
+  rb_define_method(cCircuitBreakerSharedData, "clear", semian_cb_data_array_clear, 0);
+  rb_define_method(cCircuitBreakerSharedData, "first", semian_cb_data_array_first, 0);
+  rb_define_method(cCircuitBreakerSharedData, "last", semian_cb_data_array_last, 0);
```
Sirupsen: See my comments on whether you really need all of these (i.e. implement just a `SysVSlidingWindow`).

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+  if (0 == ptr->shm_address)
+    return Qnil;
+
+  if (!semian_cb_data_lock(self))
+    return Qnil;
+
+  VALUE retval;
+  if (ptr->shm_address->arr_length >=1 && 1 <= ptr->arr_max_size)
```
Sirupsen: Why would the `arr_max_size` ever not be 1? You already do this check when initializing, you don't need to do it again.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+// struct sembuf { // found in sys/sem.h
+//   unsigned short sem_num; /* semaphore number */
+//   short sem_op; /* semaphore operation */
+//   short sem_flg; /* operation flags */
+// };
+
+typedef struct {
+  int successes;
+  int arr_length;
+  double errors[];
```
Sirupsen: I'd rename those to `int counter; int window_length; double window[];`, in general I think it'd make sense to JUST have the C be a sliding window implementation and not couple it with the circuit breaker logic at all (just have it use it).

Thoughts @csfrancis?

ext/semian_cb_data/semian_cb_data.c
kyewei: Is the issue the precision/error of floats? In code the `FLOAT` is actually a C `double`. Otherwise, `(Time.now.to_f*1000).to_i` gives me `1443539804488` or ~ `1.4 E12`, which is a few orders of magnitude bigger than `2^31-1`, `2147483647` ~ `2.1 E9` the usual size of an `int`. 

So, issue is whether accuracy in ms is needed using `long`.

Sirupsen: `long` (or `long long`) sounds fine to me and a better alternative to `double`.

ext/semian_cb_data/semian_cb_data.c
Sirupsen: The exception could be caught up the stack and not make the process terminate.

Sirupsen:
\cc @charliesome @jnunemaker you'll be interested in this for your adoption.

ext/semian_cb_data/semian_cb_data.c
kyewei: I've always assumed `shmget` needed `IPC_CREAT` to actually fetch the shmid, although it makes sense for it to not need it. But I was also following along with what was similarly done in `create_semaphore()` from `semian.c` which was just to remove the `IPC_EXCL` flag.
```
static int
create_semaphore(int key, int permissions, int *created)
{
  int semid = 0;
  int flags = 0;

  *created = 0;
  flags = IPC_EXCL | IPC_CREAT | permissions;

  semid = semget(key, kCBSemaphoreCount, flags);
  if (semid >= 0) {
    *created = 1;
  } else if (semid == -1 && errno == EEXIST) {
    flags &= ~IPC_EXCL;
    semid = semget(key, kCBSemaphoreCount, flags);
  }
  return semid;
}
```

.. which means the code should really be this. It's safer just to have the `IPC_CREAT` flag.
```
  key_t key = ptr->key;
  int byte_size = 2*sizeof(int) + ptr->arr_max_size * sizeof(double);
  int flags = IPC_CREAT | IPC_EXCL | FIX2LONG(permissions);

  if (-1 == (ptr->shmid = shmget( key, byte_size, flags))) {
    if (errno == EEXIST)
      ptr->shmid = shmget(key, byte_size, flags &= ~IPC_EXCL);
  }
```

Sirupsen: yup

ext/semian_cb_data/semian_cb_data.c
kyewei: It was giving me warnings, which I thought was weird since it should not be...

Sirupsen:
So to sum up my initial code review, questions/changes I'd like to see are:

* Merge as much code as possible with `semian.c` and compile them into the same shared object.
* Simplify scope of C code by providing only the few primitives necessary for a sliding window with `n` counters (e.g. successes, size).
* Fall back to a sliding window implementation that's completely in Ruby and works in just the memory of that Ruby VM.
* Enable resizing the size of the sliding window dynamically.
* A fresh worker (deploy, worker killed, ..) coming up in the same SysV namespace should not reset the circuit breaker.
* Use integers and not floats to store the timestamps.
* Monitoring the circuit breaker parameters (can be addressed in follow-up).
* Improve error handling and return values from functions.
* Support a fallback for `semtimedop` on platforms that don't support it (can be addressed in follow-up).
* Needs a test suite that's specific to the worker's now sharing state (see [my comment](https://github.com/Shopify/semian/pull/54#issuecomment-144054639))

ext/semian_cb_data/semian_cb_data.c
kyewei: Since we're moving to longer integers, on 32 bit machines those large numbers will be `Bignum`s, so this needs to go back in again.

ext/semian_cb_data/semian_cb_data.c
Sirupsen: :+1:

ext/semian_cb_data/semian_cb_data.c
kyewei: Array [] is not needed, but `#<<`, `#shift`, `#first`, `#size` are in `Semian::CircuitBreaker#push_time`, `#last` is in `Semian::CircuitBreaker#error_timeout_expired`, `#clear` is in `Semian::CircuitBreaker#reset` and `Semian::CircuitBreaker#close`, and the rest were provided for completeness. I can remove some of the unused ones.

Sirupsen: Let's remove at least the array indexing to remove the surface area.

ext/semian_cb_data/semian_cb_data.c
kyewei: If it's just changing and restarting semian, it's an easier fix just to call a memory cleanup if the byte size is different. I don't know how we would be able to resize while it is running unless '3-semaphore delta dance' is used.

ext/semian_cb_data/semian_cb_data.c
Sirupsen: So we are using C99 then I suppose, that's cool.

kyewei: Yea I added `-std=c99` to the flags.

ext/semian_cb_data/semian_cb_data.c
kyewei: What's the potential error?

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  if (-1 == ptr->semid){
+    rb_raise(eInternal, "semid not set, errno %d: (%s)", errno, strerror(errno));
+    return (void *)Qfalse;
+  }
+  VALUE retval;
+
+  struct timespec ts = { 0 };
+  ts.tv_sec = kCBInternalTimeout;
+
+  if (-1 == semtimedop(ptr->semid,&increment,1 , &ts)) {
```
Sirupsen: Why is this using timedop?

.gitignore
csfrancis: This shouldn't be necessary if you move to a single `.so`.

ext/semian_cb_data/semian_cb_data.c
csfrancis: I think so, but that would require a major version bump, I think?

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+  SHA1((const unsigned char *) name, strlen(name), digest.str);
+  /* TODO: compile-time assertion that sizeof(key_t) > SHA_DIGEST_LENGTH */
+  return digest.key;
+}
+
+/*
+ * Log errors
+ */
+static void
+raise_semian_syscall_error(const char *syscall, int error_num)
```
csfrancis: This could also be reused from `semian.c`.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+/*
+ * Init function exposed as ._initialize() that is delegated by .initialize()
+ */
+static VALUE
+semian_cb_data_init(VALUE self, VALUE id, VALUE size, VALUE permissions)
+{
+  semian_cb_data *ptr;
+  TypedData_Get_Struct(self, semian_cb_data, &semian_cb_data_type, ptr);
+
+  if (TYPE(id) != T_SYMBOL && TYPE(id) != T_STRING)
```
csfrancis: Style nitpick: Always put braces on `if` statements.

lib/semian/circuit_breaker.rb
csfrancis: :+1: for getting rid of the commented out code.

ext/semian_cb_data/semian_cb_data.c
kyewei: How would the #include work? I'm imagining a common semian_main.c+h file or something that both semian.c and the circuit breaker file include.

lib/semian/circuit_breaker_shared_data.rb
csfrancis: :+1:

ext/semian_cb_data/semian_cb_data.c
csfrancis: I think we should only support `T_FIXNUM` - anything else should raise.

ext/semian_cb_data/semian_cb_data.c
```diff
@@ -0,0 +1,900 @@
...
+    return self;
+  }
+  if (TYPE(permissions) != T_FIXNUM)
+    rb_raise(rb_eTypeError, "expected integer for permissions");
+
+  if (!semian_cb_data_lock(self))
+    return self;
+
+  key_t key = ptr->key;
+  if (-1 == (ptr->shmid = shmget( key,
```
csfrancis: Throwing this out there - but it might be better to use POSIX `shm_open` and then `mmap` the file here instead of `shmget`. There's one big advantage: we can use a full filename instead of a hash key.

lib/semian/circuit_breaker.rb
```diff
@@ -81,32 +91,39 @@ def half_open?
...
     def error_timeout_expired?
-      @errors.last && (@errors.last + @error_timeout < Time.now)
+      #@errors.last && (@errors.last + @error_timeout < Time.now)
+      time_f = @shared_circuit_breaker_data.last
+      time_f && (Time.at(time_f) + @error_timeout < Time.now)
     end
 
     def push_time(window, max_size, duration:, time: Time.now)
-      window.shift while window.first && window.first + duration < time
+      #window.shift while window.first && window.first + duration < time
```
csfrancis: This method is racy - there needs to be a lock around `window`. You can easily have multiple process making changes to the underlying array in parallel.

lib/semian/circuit_breaker.rb
```diff
@@ -43,13 +49,16 @@ def mark_failed(error)
 
     def mark_success
       return unless half_open?
-      @successes += 1
+      #@successes += 1
+      @shared_circuit_breaker_data.successes = @shared_circuit_breaker_data.successes+1
```
csfrancis: This is also racy - you need to increment `successes` in an atomic operation.

ext/semian_cb_data/semian_cb_data.c
kyewei: @Sirupsen requested using integer ms https://github.com/Shopify/semian/pull/54#discussion_r40686286, but that had numbers larger than 2^31-1, which means that on 32-bit machines time in ms will exceed limits for FixNum shown [here](http://ruby-doc.org/core-1.8.6/Fixnum.html) 
```
A Fixnum holds Integer values that can be represented in a native machine word (minus 1 bit). If any operation on a Fixnum exceeds this range, the value is automatically converted to a Bignum.
```

lib/semian/circuit_breaker.rb
kyewei: Should I expose the lock and unlock methods or just make the popping-until-time-is-greater a method in itself?

ext/semian_cb_data/semian_cb_data.c
csfrancis: Isn't `size` the the [`error count threshold`](https://github.com/Shopify/semian/blob/circuit-breaker-per-host/lib/semian/circuit_breaker_shared_data.rb#L4)? Should never be larger than 31 bits.

kyewei: You're right, I thought you were referring to the actual array contents.

kyewei:
I updated the PR with new changes. See above.

ext/semian/semian_shared_memory_object.c
```diff
@@ -0,0 +1,629 @@
...
+  semian_shm_object *ptr = v_ptr;
+  if (-1 == ptr->semid){
+    rb_raise(eInternal, "semid not set, errno %d: (%s)", errno, strerror(errno));
+    return (void *)Qfalse;
+  }
+  VALUE retval;
+
+  if (1 == ptr->lock_triggered) {
+    if (-1 == semop(ptr->semid,&increment,1)) {
+      rb_raise(eInternal, "error unlocking semaphore, %d (%s)", errno, strerror(errno));
```
Sirupsen: Is it actually safe to raise an exception without the GVL?

ext/semian/semian_atomic_integer.c
```diff
@@ -0,0 +1,108 @@
...
+Init_semian_atomic_integer (void)
+{
+  // Bind methods to AtomicInteger
+  VALUE cSemianModule = rb_const_get(rb_cObject, rb_intern("Semian"));
+  VALUE cAtomicInteger = rb_const_get(cSemianModule, rb_intern("AtomicInteger"));
+
+  rb_define_method(cAtomicInteger, "bind_init_fn", semian_atomic_integer_bind_init_fn_wrapper, 0);
+  rb_define_method(cAtomicInteger, "value", semian_atomic_integer_get_value, 0);
+  rb_define_method(cAtomicInteger, "value=", semian_atomic_integer_set_value, 1);
+  rb_define_method(cAtomicInteger, "increase_by", semian_atomic_integer_increase_by, 1);
```
Sirupsen: A better name for this would be `increment` with an alias to `+=`.

ext/semian/semian_atomic_integer.c
```diff
@@ -0,0 +1,108 @@
...
+}
+
+static VALUE
+semian_atomic_integer_get_value(VALUE self)
+{
+  semian_shm_object *ptr;
+  TypedData_Get_Struct(self, semian_shm_object, &semian_shm_object_type, ptr);
+
+  // check shared memory for NULL
+  if (0 == ptr->shm_address)
```
Sirupsen: Why would this happen?

ext/semian/semian_atomic_integer.c
```diff
@@ -0,0 +1,108 @@
...
+}
+
+void
+Init_semian_atomic_integer (void)
+{
+  // Bind methods to AtomicInteger
+  VALUE cSemianModule = rb_const_get(rb_cObject, rb_intern("Semian"));
+  VALUE cAtomicInteger = rb_const_get(cSemianModule, rb_intern("AtomicInteger"));
+
+  rb_define_method(cAtomicInteger, "bind_init_fn", semian_atomic_integer_bind_init_fn_wrapper, 0);
```
Sirupsen: I don't really understand what this is for. Why isn't this an `initialize`?

ext/semian/semian_atomic_integer.c
byroot: @Sirupsen unfortunately `+=` and `-=` are not really operators in Ruby. They are "translated" by the  parser into `foo = foo + bar`.

So you have no way to make a `foo += bar` atomic.

kyewei: ^ Which is why I had to add the `increase_by`. 

On the other hand, since I do have a `execute_atomically` implemented also, we have an option of removing a bit of C code and complexity by making it 
```
def increase_by
  execute_atomically do
    @value += val
  end
end
```
The locking is done so that `lock` increments a local count, `unlock` decrements a local count, and only actually `locks`/`unlocks` when going from `0->1` and `1->0` respectively. This lets something like `lock`-`lock`-`_code_here_`-`unlock`-`unlock` execute without failing and let us reuse code. What do you guys think?

ext/semian/semian_atomic_integer.c
kyewei: Handle the possibility of some chain of events leading to the shm not being attached. However, an error of some sort will probably be thrown before then. Should I remove this?

ext/semian/semian_atomic_integer.c
kyewei: This function is overriden in every subclass to attach a function pointer of type `void (*object_init_fn)(size_t byte_size, void *dest, void *prev_data, size_t prev_data_byte_size, int prev_mem_attach_count)` that initializes the C block of memory. `object_init_fn` handles the initializing and resizing of a block of memory, so it is actually called in `acquire_memory`,  not here. The user would typically write something like

https://github.com/Shopify/semian/pull/54/files#diff-99c3116b9c2c62cc89a4ae4cc84fadaaR51

The Ruby-exposed function most similar to an `initialize` would be the `SharedMemoryObject#acquire`. Arguably that could be turned into an `initialize` but then `new` call would automatically trigger acquiring a block of memory. If it makes more sense that way though, I can change it to that.

ext/semian/semian.h
```diff
@@ -0,0 +1,49 @@
...
+
+#if defined(HAVE_RB_THREAD_CALL_WITHOUT_GVL) && defined(HAVE_RUBY_THREAD_H)
+// 2.0
+#include <ruby/thread.h>
+#define WITHOUT_GVL(fn,a,ubf,b) rb_thread_call_without_gvl((fn),(a),(ubf),(b))
+#elif defined(HAVE_RB_THREAD_BLOCKING_REGION)
+ // 1.9
+typedef VALUE (*my_blocking_fn_t)(void*);
+#define WITHOUT_GVL(fn,a,ubf,b) rb_thread_blocking_region((my_blocking_fn_t)(fn),(a),(ubf),(b))
+#endif
```
Sirupsen: We should consider doing a follow-up where we implement the blocking ones ourselves (where they're needed) for OS X support.

ext/semian/semian.h
```diff
@@ -0,0 +1,49 @@
...
+
+key_t
+generate_key(const char *name);
+
+void
+raise_semian_syscall_error(const char *syscall, int error_num);
+
+
+void
+set_semaphore_permissions(int sem_id, int permissions);
```
Sirupsen: The `<type>\n<fname>` idiom is not used [for header files in Ruby](https://github.com/ruby/ruby/blob/trunk/internal.h#L747). Just keep them on the same line.

ext/semian/semian_atomic_integer.c
Sirupsen: Narh, let's keep this defensive—but `raise` instead.

ext/semian/semian_atomic_integer.c
```diff
@@ -0,0 +1,108 @@
...
+semian_atomic_integer_set_value(VALUE self, VALUE num)
+{
+  semian_shm_object *ptr;
+  TypedData_Get_Struct(self, semian_shm_object, &semian_shm_object_type, ptr);
+
+  if (0 == ptr->shm_address)
+    return Qnil;
+
+  if (TYPE(num) != T_FIXNUM && TYPE(num) != T_FLOAT)
+    return Qnil;
```
Sirupsen: Raise an `ArgumentError` instead.

ext/semian/semian_atomic_integer.c
Sirupsen: Let me think about that when I've read all of the code :)

ext/semian/semian_shared_memory_object.c
```diff
@@ -0,0 +1,629 @@
...
+  //   (also raises concurrency errors: "object allocation during garbage collection phase")
+
+  xfree(data);
+}
+static size_t
+semian_shm_object_memsize(const void *ptr)
+{
+  return
+
+  sizeof(semian_shm_object);
```
Sirupsen: This is technically valid, but please move this up to be on the same line as the `return` :-)

lib/semian/atomic_integer.rb
```diff
@@ -0,0 +1,26 @@
+module Semian
+  class AtomicInteger < SharedMemoryObject #:nodoc:
+
+    def initialize(name, permissions)
+      data_layout = [:int]
+      if acquire(name, data_layout, permissions)
+      else
+        @value=0
+      end
```
Sirupsen: It's unclear to me what's going on here. You attempt to acquire the data structure, and if you don't, you set `@value` to `0`. So what happens when I call `#value` and it acquired the data structure?

Sirupsen: Oh, the idea is that if it can call `acquire` (and it does something), it's implemented in C?

lib/semian/atomic_integer.rb
```diff
@@ -0,0 +1,26 @@
+module Semian
+  class AtomicInteger < SharedMemoryObject #:nodoc:
```
Sirupsen: Instead of this `#acquire` trick, how about calling the SysV implementation `SysVAtomicInteger` and in `semian.rb` override the Ruby implementation if semaphores are enabled? This is pretty hard to understand.

kyewei: So `SysVAtomicInteger` inherits from `AtomicInteger`, which inherits from `SharedMemoryObject`?

Sirupsen:
@kyewei is going to split this into some smaller PRs since this is too hard to review.

PR #56, opened 2015-10-28
Separated CircuitBreaker class into smaller independent state classes

kyewei:
Note: actual library changes is ~ 225 lines, the rest are new tests.

This PR contains changes to abstract out the AtomicInteger, AtomicSlidingWindow, and AtomicEnum classes with a shared SharedMemoryObject parent class. No new C code was added in this PR. 

The older, more massive PR was here: https://github.com/Shopify/semian/pull/54
See that for the general idea of how this is being designed towards.
@Sirupsen


lib/semian/atomic_enum.rb
```diff
@@ -0,0 +1,40 @@
+module Semian
+  module AtomicEnumSharedImplementation
```
fw42: conventional, this module should be called "AtomicEnum" or the filename should be "atomic_enum_shared_implementation.rb".

lib/semian/atomic_enum.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    def initialize(name, permissions, symbol_list)
+      super(name, permissions)
+      initialize_lookup(symbol_list)
+    end
+
+    def value
+      @num_to_sym.fetch super # May raise KeyError if num is not in list (invalid enum)
```
fw42: I don't think the comment adds much value... also parentheses please

lib/semian/atomic_enum.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+    def value=(sym)
+      super (@sym_to_num.fetch(sym)) # May raise KeyError if sym is not in list (invalid enum)
```
fw42: again, comment is unnecessary, imho.

lib/semian/atomic_integer.rb
```diff
@@ -0,0 +1,21 @@
...
+  class AtomicInteger < SharedMemoryObject #:nodoc:
+    attr_accessor :value
+
+    def initialize(name, permissions)
+      @value=0
+    end
+
+    def increase_by(val)
+      @value += val
+      self
```
fw42: I would expect this method to return the new value of `@value`, not `self`

lib/semian/atomic_integer.rb
```diff
@@ -0,0 +1,21 @@
+module Semian
+  class AtomicInteger < SharedMemoryObject #:nodoc:
+    attr_accessor :value
+
+    def initialize(name, permissions)
+      @value=0
+    end
+
+    def increase_by(val)
+      @value += val
```
fw42: hmm feels like this should call `self.value += val`, no? otherwise it's not using the `value=` method that you defined in the module..

lib/semian/atomic_enum.rb
kyewei: I would have thought that since the main detail of this file is implementing the `AtomicEnum` classes, whereas `AtomicEnumSharedImplementation` is just to facilitate sharing code, it should have been called atomic_enum. It's similar to `sliding_window_rb` in that it implements a class by that name. But alright.

lib/semian/circuit_breaker.rb
```diff
@@ -1,13 +1,37 @@
...
+        "#{name}_atomic_enum",
+        permissions,
+        [:closed, :half_open, :open])
+      # We do not need to #reset here since initializing is handled like this:
+      # (0) if data is not shared, then it's zeroed already
+      # (1) if no one is attached to the memory, zero it
+      # (2) otherwise, keep the data
+    end
+
+    #Replacement for attr_reader :state
```
fw42: shouldn't this comment be on the `def state` method? ... imho comments like this are useless.. it's outdated already before you even shipped it :)

byroot:
This PR is ridded with basic syntax violations. I don't feel like nitpicking each of those, but I think we should do a quick rubocop integration.

lib/semian/atomic_integer.rb
```diff
@@ -0,0 +1,21 @@
+module Semian
+  class AtomicInteger < SharedMemoryObject #:nodoc:
+    attr_accessor :value
+
+    def initialize(name, permissions)
+      @value=0
+    end
+
+    def increase_by(val)
```
fw42: can't you `def +(val)` this?

lib/semian/shared_memory_object.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    def shared?
+      @using_shared_memory ||= Semian.extension_loaded && @using_shared_memory
+    end
+
+    def destroy
+      _destroy if respond_to?(:_destroy) && @using_shared_memory
+    end
+
+    # Functions dealing with acquiring memory
```
fw42: I don't think we need a "function dealing with acquiring memory" comment on a function that's called "acquire_memory"

lib/semian/atomic_integer.rb
kyewei: Maybe a different method name would make it clearer? The purpose is to increase the stored value by a set number. When the C implementation is added, keeping the entire mutating in one function is how an operation like `atomicinteger.value += 5` can maintain atomicity.

Sirupsen:
@byroot will that re-run here automatically?

byroot:
@Sirupsen you mean rubocop? After a rebase yes.

lib/semian/circuit_breaker.rb
```diff
@@ -1,13 +1,35 @@
 module Semian
-  class CircuitBreaker
-    attr_reader :state
-
-    def initialize(exceptions:, success_threshold:, error_threshold:, error_timeout:)
+  class CircuitBreaker #:nodoc:
+    def initialize(name, exceptions:, success_threshold:, error_threshold:, error_timeout:, permissions: 0660)
+      @name = "#{name}"
```
byroot: Why not just `@name = name`? If you want to cast it to String, `name.to_s` will do.

lib/semian/atomic_enum.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  class AtomicEnum < AtomicInteger #:nodoc:
+    undef :increase_by
+    include AtomicEnumSharedImplementation
+  end
+  class SysVAtomicEnum < SysVAtomicInteger #:nodoc:
+    undef :increase_by
+    include AtomicEnumSharedImplementation
+  end
```
byroot: THis is weird. Why not have an `AbstractAtomicEnum`, and then `SysVAtomicEnum` and `AtomicEnum` inherits from it.

kyewei: Can't do that since C methods will hook onto SysVAtomicInteger methods, so the SysV classes need to inherit from upper ones.

lib/semian/atomic_integer.rb
```diff
@@ -0,0 +1,20 @@
...
+  class AtomicInteger < SharedMemoryObject #:nodoc:
+    attr_accessor :value
+
+    def initialize(_name, _permissions)
+      @value = 0
+    end
+
+    def increase_by(val)
+      self.value += val
+    end
```
byroot: Beside tests I don't see why we would increment by anything else than 1. So I would at least provide a shorthand that is `increment`

kyewei: You mean add another method that does the following? OK.
```
def increment
  increase_by(1)
end
```

byroot: Yes, perfect.

byroot: Also `increase_by` -> `increment_by` IMHO

lib/semian.rb
```diff
@@ -154,7 +157,12 @@ def resources
 require 'semian/protected_resource'
 require 'semian/unprotected_resource'
 require 'semian/platform'
-if Semian.sysv_semaphores_supported? && Semian.semaphores_enabled?
+require 'semian/shared_memory_object'
+require 'semian/sliding_window'
+require 'semian/atomic_integer'
+require 'semian/atomic_enum'
+Semian.extension_loaded = Semian.sysv_semaphores_supported? && Semian.semaphores_enabled?
+if Semian.extension_loaded
```
Sirupsen: This basically means we now have three different ways to check for SysV: `Semian.sysv_semaphores_supported?`, `Semian.semaphores_enabled?` and now `Semian.extension_loaded`. IMO calling `Semian.semaphores_enabled?` should be enough here, but it will have to check that they're supported:

```ruby
  def semaphores_enabled?
    !ENV['SEMIAN_SEMAPHORES_DISABLED'] && Semian.sysv_semaphore_supported?
  end
```

lib/semian/shared_memory_object.rb
```diff
@@ -0,0 +1,48 @@
+module Semian
+  class SharedMemoryObject #:nodoc:
```
Sirupsen: Why is all this SysV stuff in this PR? Why not start with just the Ruby part?

lib/semian/circuit_breaker.rb
```diff
@@ -1,13 +1,35 @@
...
-      reset
+
+      @sliding_window = ::Semian::SysVSlidingWindow.new(
+        "#{name}_sliding_window",
+        @error_count_threshold,
+        permissions)
+      @successes = ::Semian::SysVAtomicInteger.new(
+        "#{name}_atomic_integer",
+        permissions)
+      @state = ::Semian::SysVAtomicEnum.new(
```
Sirupsen: Why is this SysV stuff in this PR?

Sirupsen:
Why is all this SysV stuff in this PR? Couldn't this PR just be using the `AtomicInteger`, `AtomicEnum` and `SlidingWindow` primitives from Ruby?

lib/semian/circuit_breaker.rb
kyewei: This is the extent of all the ruby code that needs to be introduced, and it's small.

Sirupsen: I think you need to split this PR in two yet again, and let the SysV stuff be reviewed independently.

kyewei:
@Sirupsen it now only includes the primitives. It (tests, library, etc) has the effect of being less coupled now as well.

lib/semian/atomic_enum.rb
```diff
@@ -0,0 +1,37 @@
+module Semian
+  class AtomicEnum < AtomicInteger #:nodoc:
+    undef :increment_by
+    undef :increment
```
Sirupsen: Why `undef` instead of overriding them with an exception? Does it actually matter that much even if they're defined? It'd work and would be equivalent to iterating on the enum, which is not a crazy thing to do.

lib/semian/atomic_enum.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    module AtomicEnumSharedImplementation
+      def initialize(name, permissions, symbol_list)
```
Sirupsen: Similar comment on `name` and `permissions`.

lib/semian/atomic_enum.rb
```diff
@@ -0,0 +1,37 @@
...
+        super(name, permissions)
+        initialize_lookup(symbol_list)
```
Sirupsen: You could replace this with:

```
@num_to_sym = Hash[symbol_list.each_with_index.to_a].invert
```

Due to the well-named variable I think it's understandable.

lib/semian/atomic_enum.rb
```diff
@@ -0,0 +1,37 @@
...
+      end
+
+      def value
+        @num_to_sym.fetch(super)
+      end
+
+      def value=(sym)
+        super(@sym_to_num.fetch(sym))
```
Sirupsen: :+1: on `#fetch` :-)

It would make sense to `alias_method` this to `#transition` or something similar, which would make the Circuit Breaker code easier to read. Yes / no?

lib/semian/atomic_enum.rb
```diff
@@ -0,0 +1,37 @@
...
+
+        @sym_to_num = {}
+        symbol_list.each.with_index do |sym, idx|
+          @sym_to_num[sym] = idx
+        end
+        @num_to_sym = @sym_to_num.invert
+      end
+    end
+
+    include AtomicEnumSharedImplementation
```
Sirupsen: I do not really understand why this has to be a module? The SysV implementation can just inherit from this, no?

lib/semian/atomic_integer.rb
```diff
@@ -0,0 +1,28 @@
+module Semian
+  class AtomicInteger #:nodoc:
```
Sirupsen: Once we're on the SysV implementation we'll want to create a Thread Safe variant of this as well.

lib/semian/atomic_integer.rb
```diff
@@ -0,0 +1,28 @@
...
+    attr_accessor :value
+
+    def initialize(_name, _permissions)
```
Sirupsen: Why the underscored arguments here? These don't matter for `AtomicInteger`. They only matter for the SysV adapter which will override the initializer regardless.

lib/semian/atomic_integer.rb
```diff
@@ -0,0 +1,28 @@
...
+      @value = 0
+    end
+
+    def increment_by(val)
+      self.value += val
+    end
+
+    def increment
+      increment_by(1)
+    end
```
Sirupsen: It'd be simpler (and more common in Ruby) to just do:

```ruby
def increment(val = 1)
  value += val
end
```

lib/semian/atomic_integer.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    def execute_atomically
+      yield if block_given?
+    end
```
Sirupsen: I'm not a huge fan of this method name, but I see what you're trying to do. How about just `#transaction`?

lib/semian/atomic_integer.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    def shared?
+      false
+    end
```
Sirupsen: Why is this important? It's not actually used anywhere (yet?).

lib/semian/atomic_integer.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    def destroy
+    end
```
Sirupsen: Should this set the value back to 0?

lib/semian/circuit_breaker.rb
```diff
@@ -1,13 +1,35 @@
 module Semian
-  class CircuitBreaker
-    attr_reader :state
-
-    def initialize(exceptions:, success_threshold:, error_threshold:, error_timeout:)
+  class CircuitBreaker #:nodoc:
+    def initialize(name, exceptions:, success_threshold:, error_threshold:, error_timeout:, permissions: 0660)
+      @name = name.to_s
```
Sirupsen: Can you explain why it needs a name now? Is this purely so you have something to hash on for the SysV implementation? Wouldn't it be better to pass a reference to the resource which has a name?

lib/semian/circuit_breaker.rb
```diff
@@ -1,13 +1,35 @@
...
       @success_count_threshold = success_threshold
       @error_count_threshold = error_threshold
       @error_timeout = error_timeout
       @exceptions = exceptions
-      reset
+
+      @sliding_window = ::Semian::SlidingWindow.new(
```
Sirupsen: Why leak the abstraction of the data structure when you could retain the previous name of `@errors`?

lib/semian/circuit_breaker.rb
```diff
@@ -1,13 +1,35 @@
...
+        "#{name}_sliding_window",
+        @error_count_threshold,
+        permissions)
+      @successes = ::Semian::AtomicInteger.new(
+        "#{name}_atomic_integer",
+        permissions)
+      @state = ::Semian::AtomicEnum.new(
+        "#{name}_atomic_enum",
+        permissions,
```
Sirupsen: Permissions don't make sense here.

lib/semian/circuit_breaker.rb
```diff
@@ -1,13 +1,35 @@
...
+        [:closed, :half_open, :open])
+      # We do not need to #reset here since initializing is handled like this:
+      # (0) if data is not shared, then it's zeroed already
+      # (1) if no one is attached to the memory, zero it
+      # (2) otherwise, keep the data
+    end
+
+    def shared?
+      @sliding_window.shared? && @successes.shared? && @state.shared?
```
Sirupsen: Similarly, why does this need to be shared?

lib/semian/circuit_breaker.rb
```diff
@@ -1,13 +1,35 @@
...
+    end
+
+    def state
+      @state.value
```
Sirupsen: You never use this.

lib/semian/sliding_window.rb
```diff
@@ -0,0 +1,55 @@
...
+      @window.pop while @window.size >= @max_size
+      @window.unshift(time_ms)
+      self
+    end
+
+    def clear
+      @window = []
+    end
+
+    def execute_atomically
```
Sirupsen: Similar concern about naming.

lib/semian/sliding_window.rb
```diff
@@ -0,0 +1,55 @@
...
+      push(time_ms)
+    end
+
+    def push(time_ms)
+      @window.shift while @window.size >= @max_size
+      @window << time_ms
+      self
+    end
+
+    def unshift(time_ms)
```
Sirupsen: Calling this `#prepend` is more true to the data structure IMO. But why do we even need this method? If this is purely for testing (which it seems to be) it should be a helper instead of a public method on the data structure.

lib/semian/sliding_window.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    def <<(time_ms)
+      push(time_ms)
+    end
+
+    def push(time_ms)
+      @window.shift while @window.size >= @max_size
+      @window << time_ms
+      self
```
Sirupsen: Why return `self`?

lib/semian/sliding_window.rb
```diff
@@ -0,0 +1,55 @@
...
+      @window = []
+    end
+
+    def execute_atomically
+      yield if block_given?
+    end
+
+    def shared?
+      false
+    end
```
Sirupsen: Again, why is this important? Why does this abstraction have to leak?

lib/semian/sliding_window.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    def destroy
+    end
```
Sirupsen: Call `#reset`?

lib/semian/unprotected_resource.rb
```diff
@@ -36,5 +36,12 @@ def request_allowed?
 
     def mark_failed(_error)
     end
+
+    def mark_success
+    end
+
+    def circuit_breaker_shared?
+      false
+    end
```
Sirupsen: Why does this matter?

lib/semian/atomic_enum.rb
kyewei: I suppose to do that I'll have to add a `num%size` in to keep it within range.

test/atomic_enum_test.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+  end
+
+  private
+
+  def atomic_enum_classes
+    @classes ||= retrieve_descendants(::Semian::AtomicEnum)
+  end
+
+  def run_test_with_atomic_enum_classes(klasses = atomic_enum_classes)
```
Sirupsen: This is a little crazy :-) I think we should just write tests for each one. They may have different tests too, e.g. guarantee they're atomic, shared, and so on.

test/atomic_enum_test.rb
```diff
@@ -0,0 +1,35 @@
+require 'test_helper'
+
+class TestAtomicEnum < MiniTest::Unit::TestCase
```
Sirupsen: When writing unit tests, I like to browse the class I'm testing and look at each line and think: If I delete this line of code, will a test fail? If I change this greater than to a less than, would a test fail? What if it's a less than or equal?

I bet you can find a lot of lines for which that's not true and increase test coverage.

test/atomic_integer_test.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+  end
+
+  private
+
+  def atomic_integer_classes
+    @classes ||= retrieve_descendants(::Semian::AtomicInteger).select { |klass| /Integer/.match(klass.name) }
+  end
+
+  def run_test_with_atomic_integer_classes(klasses = atomic_integer_classes)
```
Sirupsen: Similar note.

test/atomic_integer_test.rb
```diff
@@ -0,0 +1,57 @@
...
+      @integer.increment_by(4)
+      assert_equal(4, @integer.value)
+      @integer.increment
+      assert_equal(5, @integer.value)
+      @integer.value = 10
+      assert_equal(10, @integer.value)
+    end
+  end
+
+  def test_execute_atomically_actually_is_atomic
```
Sirupsen: Why all this when the implementation doesn't even have a mutex?

test/atomic_integer_test.rb
```diff
@@ -0,0 +1,57 @@
+require 'test_helper'
+
+class TestAtomicInteger < MiniTest::Unit::TestCase
+  def test_operations
```
Sirupsen: When doing a lot of assertions in one test in a unit test, that's usually a smell you could have more well-named tests. Note that one of those may be exactly equal to this one, but has a good name on what you're trying to test and perhaps only one or two assertions.

test/sliding_window_test.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+  end
+
+  private
+
+  def sliding_window_classes
+    @classes ||= retrieve_descendants(::Semian::SlidingWindow)
+  end
+
+  def run_test_with_sliding_window_classes(klasses = sliding_window_classes)
```
Sirupsen: Similar comment.

test/sliding_window_test.rb
```diff
@@ -0,0 +1,63 @@
+require 'test_helper'
+
+class TestSlidingWindow < MiniTest::Unit::TestCase
+  def test_sliding_window_functionality
```
Sirupsen: Similar comments on more tests + smaller tests.

lib/semian/sliding_window.rb
```diff
@@ -0,0 +1,55 @@
...
+    def initialize(_name, max_size, _permissions)
+      @max_size = max_size
+      @window = []
+    end
+
+    # A sliding window is a structure that stores the most @max_size recent timestamps
+    # like this: if @max_size = 4, current time is 10, @window =[5,7,9,10].
+    # Another push of (11) at 11 sec would make @window [7,9,10,11], shifting off 5.
+
+    def resize_to(size)
```
Sirupsen: You only use this in tests. Shouldn't this be used instead of the very similar implementation in `push_time`?

lib/semian/sliding_window.rb
```diff
@@ -0,0 +1,55 @@
...
+      throw ArgumentError.new('size must be larger than 0') if size < 1
+      @max_size = size
+      @window.shift while @window.size > @max_size
+      self
```
Sirupsen: Why return self?

lib/semian/sliding_window.rb
```diff
@@ -0,0 +1,55 @@
...
+      @max_size = max_size
+      @window = []
+    end
+
+    # A sliding window is a structure that stores the most @max_size recent timestamps
+    # like this: if @max_size = 4, current time is 10, @window =[5,7,9,10].
+    # Another push of (11) at 11 sec would make @window [7,9,10,11], shifting off 5.
+
+    def resize_to(size)
+      throw ArgumentError.new('size must be larger than 0') if size < 1
```
Sirupsen: We don't use `throw` and `try` in Shopify Ruby, use `raise`.

lib/semian/atomic_enum.rb
kyewei: So something like
```ruby
@sym_to_num = Hash[symbol_list.each_with_index.to_a]
@num_to_sym = @sym_to_num.invert
```
? I'd still like to keep it in a separate function though, since the SysV version will definitely reuse it

lib/semian/atomic_enum.rb
kyewei: There was a diamond problem in the fact that if we let SysVAtomicInteger inherit from AtomicInteger, and hook the methods in SysVAtomicInteger, to maintain both the Enum functionality and the SysV functionality in SysVAtomicEnum you'd have to get SysVAtomicEnum to einherit from SysVAtomicInteger and mixin the enum functionality... or hook twice.

lib/semian/atomic_enum.rb
Sirupsen: Yup, that's fine with me. I think you can give it a better name too.

lib/semian/atomic_enum.rb
kyewei: What's the comment? I can't find it.

Sirupsen: https://github.com/Shopify/semian/pull/56/files#r43757316

lib/semian/atomic_enum.rb
Sirupsen: Yeah. On the other hand, it wouldn't actually be used anywhere. I'm somewhat indifferent.

lib/semian/atomic_integer.rb
kyewei: Are you suggesting just making it a 0 parameter initializer? I'd like to keep the interface the same though.

lib/semian/atomic_integer.rb
kyewei: Are you referencing something like this? http://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html
I've never personally seen that use but if it's idiomatic I'll put that in.

Sirupsen: Yeah, it's common in AR, and e.g. `PStore` in the std lib uses that name as well.

lib/semian/circuit_breaker.rb
kyewei: It's so that the individual structures have a name. If you want to invert it, my argument against that would be that you'd need to have this code:
```
::Semian::AtomicEnum.new(
        name,
        permissions,
        [:closed, :half_open, :open])
```
outside the circuit breaker class and would be that wouldn't be great.

lib/semian/atomic_integer.rb
kyewei: If we're going to add a shared version, then we need to be able to distinguish between for example `AtomicInteger`, `SysVAtomicInteger`, and the fallback `SysVAtomicInteger` on a semaphore-less system.

lib/semian/circuit_breaker.rb
kyewei: What do you mean?

lib/semian/sliding_window.rb
kyewei: Maintain the abstraction that the wrapper is responsible for handling appending, even if it is delegating it to an actual array. In `sliding_window << 1 << 2`, if there was no self, then after the first push it would leak the array.

lib/semian/atomic_integer.rb
Sirupsen: Why does the distinction matter?

kyewei: I'd think that users would want to know if their structures are shared or not, and also that the data may be changed from a different worker. It also makes testing shared components easier since you can test the expected behaviour when you change another structure with the same name.

Sirupsen: Why would the user care by calling a method? They'd find out which adapter is used, but they don't need to pragmatically know.

I don't understand why it's easier to test. Each adapter should be tested independently.

kyewei: The user should care because they might want to handle the two cases that can occur (shared, not shared) separately. 

Easier to test and less repeated code through something like this:
```ruby
def test_shared_attribute
  run_test_with_sliding_window_classes do |klass, instance|
    return unless instance.shared?
    # test contents
  end
end
```
There is no point in writing the same code x times, once for each type of class and once for a shared vs not-shared instance of a class. 

How else can you distinguish between a shared-memory version of a class and a not-shared-memory version? There is no other method that can distinguish this without explicitly testing to see if it is. When the `SharedMemoryObject` superclass comes in, `shared?`, `execute_atomically`, and `destroy` will be moved there. 

I'm pretty adamant about this being included since many later tests for testing the shared component depend on this as well and otherwise there is no way to distinguish them making writing tests more difficult than it should be. If the issue is that it doesn't make sense to include it at this iteration, but that it will be included when I add in the shared memory class, then that's OK.

Sirupsen: I'm not a big fan of this pattern for testing adapters. At least the classes to test should be hardcoded. @byroot can you weight in here for a third opinion? Do you think having `#shared` and iterating over all the adapter classes and sharing the tests is OK?

lib/semian/circuit_breaker.rb
Sirupsen: Permissions don't make sense for a Ruby implementation. This is adapter specific.

lib/semian/atomic_integer.rb
Sirupsen: Yes. I don't have a good idea of what this will look like, but the in-memory Ruby adapter can't take permissions just because another adapter happens to need it.

lib/semian/atomic_enum.rb
Sirupsen: Ah, I see what you mean.

lib/semian/atomic_integer.rb
byroot: Sharing tests are a good idea. A single test testing them all is not.

What I would recommend is simply using a module:

```ruby
class SysVAtomicIntegerTest < Minitest::TestCase
  include SharedAtomicIntegerTestCases

  # Add tests specific to SysVAtomicInteger here.
end
```

kyewei: Here is the type of code I have ready for the iteration after this one for this type of purpose:
```ruby
def test_memory_is_shared
  run_test_with_atomic_integer_classes do |klass|
    next unless @integer.shared?
    integer_2 = klass.new('TestAtomicInteger', 0660)
    integer_2.value = 100
    assert_equal 100, @integer.value
    @integer.value = 200
    assert_equal 200, integer_2.value
    @integer.value = 0
    assert_equal 0, integer_2.value
  end
end

def test_memory_not_reset_when_at_least_one_worker_using_it
  run_test_with_atomic_integer_classes do |klass|
    next unless @integer.shared?

    @integer.value = 109
    integer_2 = klass.new('TestAtomicInteger', 0660)
    assert_equal @integer.value, integer_2.value
    pid = fork do
      integer_3 = klass.new('TestAtomicInteger', 0660)
      assert_equal 109, integer_3.value
      sleep
    end
    sleep 1
    Process.kill("KILL", pid)
    Process.waitall
    fork do
      integer_3 = klass.new('TestAtomicInteger', 0660)
      assert_equal 109, integer_3.value
    end
    Process.waitall
  end
end

private

def atomic_integer_classes
  @classes ||= retrieve_descendants(::Semian::AtomicInteger)
end

def run_test_with_atomic_integer_classes(klasses = atomic_integer_classes)
  klasses.each do |klass|
    begin
      @integer = klass.new('TestAtomicInteger', 0660)
      @integer.value = 0
      yield(klass)
    ensure
      @integer.destroy
    end
  end
end


```

With this you can test the structures without regard to whether it is a shared `SysVAtomicInteger`, a non-shared (disabled or OS not supporting) `SysVAtomicInteger`, or a regular `AtomicInteger`. Furthermore, you can even make a subclass like this to emulate shareness for testing correctness purposes (since on non-semaphore systems, you want to have at least one set of tests run) as long as `#shared?` is implemented and is overriden to be true:
```ruby
class FakeSysVAtomicInteger < Semian::AtomicInteger
  class << self
    attr_accessor :resources
  end
  self.resources = {}
  attr_accessor :name
  def self.new(name, _permissions)
    obj = resources[name] ||= super
    obj.name = name
    obj
  end

  def destroy
    self.class.resources.delete(@name)
    super
  end

  def shared?
    true
  end
end
```
Without something like `#shared?` it is hard to do a thorough test that captures all possible conditions without repeating code excessively.

kyewei: ^See above for more.

So @byroot, how would you handle testing `SysVAtomicInteger` on a system that doesn't support semaphores? You would expect `SysVAtomicIntegerTest` to have assertions that things like
```
integer_1 = Semian::SysVAtomicInteger.new("example", 0660)
integer_2 = Semian::SysVAtomicInteger.new("example", 0660)
integer_2.value = 2
assert_equal(integer_2.value, integer_1.value)
```
That last line would definitely fail since you're running a fallback code while expecting it not to have fallback.

byroot: > how would you handle testing SysVAtomicInteger on a system that doesn't support semaphores?

You don't test it. It's what we always did. 

> next unless shared? then it would be fine.

If that is what shared is for, then it's bad naming. 
I encourage you to take a look at the `thread_safe` gem, on how they handle code radically different based on the platform: https://github.com/ruby-concurrency/thread_safe

kyewei: Yes, essentially I want something that lets me differentiate between a structure that has shared memory (and therefore needs a different set of methods for `destroy` (since we now need to free the semaphore, etc), `execute_atomically` (which locks, calls yield, and unlocks), but because we now have a fallback implementation as well, this functionality can't just be a complete override, so we need some `bool` of some sort to determine whether we are actually using shared memory. Thats what the `#shared?` was for. 

I can concede the iterating over all the classes for testing part, but that `bool` has to be there for things like `destroy` and `execute_atomically`. `#shared?` can be private if that seems like a better idea.

kyewei: So what's the consensus for `#shared?`? @byroot @Sirupsen. The most important reason is outlined above:
> Yes, essentially I want something that lets me differentiate between a structure that has shared memory (and therefore needs a different set of methods for `destroy` (since we now need to free the semaphore, etc), `execute_atomically` (which locks, calls yield, and unlocks). However, because we now have a fallback implementation as well, this functionality can't just be a complete override, so we need some `bool` of some sort to determine whether we are actually using shared memory and branch off. Thats what the `#shared?` was for.

If `#shared?` isn't included (it could be `private` for all I care) then I need a different way of implementing falbacks.

byroot: If it's private then I don't mind.

lib/semian/atomic_enum.rb
kyewei: I am thinking now that it may be possible to just use `extend Forwardable` and delegate. I usually strayed away from this because in most cases since there is delegation happening, lets say to an instance variable, the instance variable also needs to access content of the wrapper class, so there is back and forth communication, which is not good. However, in this case I think it works okay since the AtomicInteger instance doesn't need to know about the enum abstraction.

It'll just have a huge list of delegators in the end, something like this:
```
def_delegators :@integer, :semid, :shmid, :execute_atomically, :shared?,
                   :destroy, :acquire_memory_object, :bind_init_fn
```

lib/semian/atomic_integer.rb
kyewei: Perhaps with keyword arguments? For example:
```ruby
class AtomicEnum
  def initialize(symbol_list)
  ...

class SysVAtomicEnum < AtomicEnum
  def initialize(symbol_list, name: , permissions: )
  ...

```
This would work. Would there be any compatibility issues?

lib/semian/unprotected_resource.rb
kyewei: If... you're referring to `circuit_breaker_shared?`, then I'm going to remove it along with making `#shared?` private. If it was the `mark_success`, there was a test somewhere that checked the `#public_instance_methods` of both types of the `Resource` wrappers to make sure they were the same.

lib/semian/sliding_window.rb
kyewei: This one changes the max_size as well. Although, the similar implementation in `push_time` has this line, which is what I think you're referring to:
```ruby
window.shift if window.size == max_size
```
I realized that line is redundant since `#<<` manages this already, so I'm removing that one line.

lib/semian/sliding_window.rb
kyewei: I put it in so cascading could work `window.resize_to(3).pop(1)`. 
Would returning the new size be more logical? It'll prevent ^ from being possible though.

kyewei:
This is probably good for another review. Things I changed: testing style, the initializer (now has keyword arguments for `name:`, `permissions:`, fixed `destroy`, made `shared?` private, added more tests.

lib/semian.rb
```diff
@@ -162,7 +167,7 @@ def resources
     Semian.logger.info("Semian sysv semaphores are not supported on #{RUBY_PLATFORM} - all operations will no-op")
   end
 
-  unless Semian.semaphores_enabled?
+  if ENV['SEMIAN_SEMAPHORES_DISABLED']
```
fw42: why this change? seems more reasonable to use the method instead of check the env var again

lib/semian/atomic_enum.rb
```diff
@@ -0,0 +1,39 @@
+require 'forwardable'
+
+module Semian
+  class AtomicEnum #:nodoc:
+    extend Forwardable
+
+    def_delegators :@integer, :execute_atomically, :transaction, :shared?, :destroy
+    private :shared?
+
+    def initialize(symbol_list, **_options)
```
fw42: I don't understand, why does this have an `options` parameter? You are not using those values anywhere, are you?

lib/semian/atomic_integer.rb
```diff
@@ -0,0 +1,29 @@
+module Semian
+  class AtomicInteger #:nodoc:
+    attr_accessor :value
+
+    def initialize(**_options)
```
fw42: What's with the options again? Why does this exist if it's not used?

lib/semian/atomic_integer.rb
fw42: I don't understand this method or this name... I don't see anything begin atomic.. Can you explain?

lib/semian/circuit_breaker.rb
```diff
@@ -1,13 +1,24 @@
...
+                                            permissions: permissions)
+      @successes = ::Semian::AtomicInteger.new(name: "#{name}_atomic_integer",
+                                               permissions: permissions)
+      @state = ::Semian::AtomicEnum.new([:closed, :half_open, :open],
+                                        name: "#{name}_atomic_enum",
+                                        permissions: permissions)
+      # We do not need to #reset here since initializing is handled like this:
+      # (0) if data is not shared, then it's zeroed already
+      # (1) if no one is attached to the memory, zero it
+      # (2) otherwise, keep the data
```
fw42: This comment is gonna get outdated... also I don't understand it... "if no one is attached to the memory, zero it"? what does attached mean? and "who" would be attached to things? this is very confusing, imho

lib/semian/circuit_breaker.rb
```diff
@@ -43,70 +53,78 @@ def mark_failed(_error)
...
+      time_ms = @errors.last
+      time_ms && (Time.at(time_ms / 1000) + @error_timeout < Time.now)
     end
 
-    def push_time(window, max_size, duration:, time: Time.now)
-      window.shift while window.first && window.first + duration < time
-      window.shift if window.size == max_size
-      window << time
+    def push_time(window, duration:, time: Time.now)
+      @errors.execute_atomically do # Store an integer amount of milliseconds since epoch
```
fw42: what is this comment referring to? to this line? or the next line? or this whole method? I don't think this adds much value

lib/semian/sliding_window.rb
```diff
@@ -0,0 +1,55 @@
+module Semian
+  class SlidingWindow #:nodoc:
+    extend Forwardable
+
+    def_delegators :@window, :size, :pop, :shift, :first, :last
+    attr_reader :max_size
+
+    def initialize(max_size, **_)
```
fw42: `**_`? I don't understand

lib/semian/sliding_window.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    def push(time_ms)
+      @window.shift while @window.size >= @max_size
+      @window << time_ms
+      self
+    end
+
+    def clear
+      @window = []
```
fw42: could use `@window.clear` to save the unnecessary allocation..

test/atomic_enum_test.rb
fw42: This class should be called `AtomicEnumTest` for consistent naming (the file is called `atomic_enum_test.rb`)

test/test_helper.rb
```diff
@@ -6,6 +6,7 @@
 require 'timecop'
 require 'tempfile'
 require 'fileutils'
+require 'timeout'
```
fw42: is this used anywhere?

lib/semian.rb
kyewei: This was done a few commits ago, when Simon suggested `#semaphores_enabled?` should include `&& Semian.sysv_semaphores_supported?`, which means the line here needs to be changed.

lib/semian/atomic_enum.rb
kyewei: I wanted to keep the interface the same. Otherwise things like `name: 'blah'`, etc in the initializer will throw errors.

test/atomic_enum_test.rb
kyewei: Actually, if you look at the naming of all the test files in this folder, it's written like this.

lib/semian/atomic_enum.rb
fw42: What interface? I don't understand.. Why would you give this thing a `name: 'bla'` parameter if the class is not doing anything with it?

test/atomic_enum_test.rb
fw42: hrm ok if all the other tests are already doing it wrong then leave it like this... :-(

lib/semian/circuit_breaker.rb
kyewei: That... applies to the C portion. I got rid of it.

kyewei:
Any other issues?

test/atomic_enum_test.rb
kyewei: I kinda dislike doing things like that prematurely since I assume they're there for a reason. If I do, someone always dislikes the change (and is usually very vocal about it). But OK, I'll do that after.

lib/semian/atomic_enum.rb
kyewei: When the SysV stuff come in, they need a name to hash on, and permissions to use for the `semget` functions, etc. I was talking about how it keeps the interface the same if the non SysV structures accept the same, but just don't do anything with it. But if you give keyword arguments to an initializer that doesn't accept keyword arguments it throws an exception.

lib/semian/atomic_integer.rb
kyewei: Look at this first: https://github.com/Shopify/semian/pull/54#discussion-diff-42760663

I don't think we've decided yet what to do with it in this case of AtomicInteger, its part of `SharedMemoryObject#execute_atomically`

kyewei: The reason I found this "separating the big PR into small PR's" tedious is that some code is placed out of context and I have to re-explain every implementation reasoning that would otherwise make perfect sense as a whole, or redirect people to PR https://github.com/Shopify/semian/pull/54 :(

fw42: I see. I didn't even know there was another PR.

kyewei:
I rebased + removed `#shared`, `#execute_atomically`. There should be no more code that doesn't make sense in context.

lib/semian/atomic_enum.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    def_delegators :@integer, :destroy
+
+    def initialize(symbol_list)
+      @integer = Semian::AtomicInteger.new
+      initialize_lookup(symbol_list)
+    end
+
+    def increment(val = 1)
+      @integer.value = (@integer.value + val) % @sym_to_num.size
```
byroot: How is this atomic? You read the value, do your computation in memory, and then assign it back.

Am I missing something?

I'm also not sure why you'd what to "increment" an Enum.

kyewei: Simon mentioned iterating through the enums like this, I originally had `undef :increment`, but he said he didn't care whether the iterating should be added or not, so I could remove it.

lib/semian/atomic_enum.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    def initialize_lookup(symbol_list)
+      # Assume symbol_list[0] is mapped to 0
+      # Cannot just use #object_id since #object_id for symbols is different in every run
+      # For now, implement a C-style enum type backed by integers
+
+      @sym_to_num = Hash[symbol_list.each_with_index.to_a]
+      @num_to_sym = @sym_to_num.invert
+    end
+  end
```
byroot: Do you really need that Enum to be generalized? I think the code would read easier if the enum would clearly limit itself to the 3 states we need.

lib/semian/atomic_enum.rb
kyewei: Should I add a mutex-lock at this stage? https://github.com/Shopify/semian/pull/56/files#r43757308

byroot: It's what I'm trying to say. Even if your `AtomicInteger` was actually atomic, the way to you use it is not.

byroot: But maybe I'm missing something. Are all these classes just stubs?

lib/semian/atomic_enum.rb
kyewei: I've been working towards this for a while now, the motivation was to separate/abstract out the classes into distinct structures. 

What part of the code is difficult to read? Maybe that is the issue here instead. Is it the one liner on line 34? See this: https://github.com/Shopify/semian/pull/56#discussion_r43757296

lib/semian/atomic_enum.rb
kyewei: Most are really small stubs, since when I put in the SharedMemoryObject stuff, Simon said that was already too big, so I had to rip that out, and so the mutex stuff, which I would put in there, is gone. Maybe a rename to just `Integer` than `AtomicInteger` ?

kyewei: Here's what it should look like: 
https://github.com/Shopify/semian/blob/circuit-breaker-per-host/lib/semian/atomic_integer.rb

Notice the `surround_with_mutex`, ... but I threw that in `SharedMemoryObject`

PR #57, opened 2015-10-28
Style guide enforced on CI

byroot:
No offense but @kyewei PRs have been ridded with totally non idiomatic code formatting and syntax (which is totally understandable for a Ruby newcomer).

I know it's not what is important, but it's hard to focus on real problems when it gets in the way.
The idea here is to enforce basic rules automatically, so that we don't have to waste time on PRs.

I took the rules from shipit-engine, with a couple changes. 

@fw42 @Sirupsen thoughts? I'm gonna be honest I don't want to debate any single rule, it's more a "take it or leave it" PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,10 @@ sudo: true
 before_install:
   - scripts/install_toxiproxy.sh
 
+script:
+  - bundle exec rake
+  - bundle exec rubocop
+
 rvm:
   - 2.1.1
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,6 @@ sudo: true
 before_install:
   - scripts/install_toxiproxy.sh
 
-script:
-  - bundle exec rake
-  - bundle exec rubocop
-
 rvm:
   - 2.1.1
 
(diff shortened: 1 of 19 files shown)


fw42:
:heart: rubocop

.travis.yml
```diff
@@ -5,6 +5,10 @@ sudo: true
 before_install:
   - scripts/install_toxiproxy.sh
 
+script:
+  - bundle exec rake
+  - bundle exec rubocop
```
fw42: can you make this also part of the default rake task? otherwise people will run the tests locally and be surprised why they don't pass on CI

byroot: I can try. But rubocop Rake task has this annoying property of always running before the test. I don't want the tests not to be run if syntax is not ok.

Anyway let's try

byroot:
@fw42 ok, I got it working as a default rake task: https://travis-ci.org/Shopify/semian/builds/87937792

:shipit: ?

Sirupsen:
Sure.

BTW Kye is working on splitting his PR into much smaller PRs.

PR #58, opened 2015-10-28
Net::HTTP Semian Adapter

kyewei:
^ What the title says. 

If `semian/net_http` is included in a rails server instance, then some sort of initialization like this has to occur:
In `http.yml`:
```yaml
development:
  default:
    tickets: 1
    success_threshold: 1
    error_threshold: 3
    error_timeout: 10
...
```
In an initialization file somewhere:
```ruby

# If we want default semian options:
http_config = YAML.load_file("#{Rails.root}/path_to_file/http.yml")
Semian::NetHTTP.semian_configuration = proc do |host, port|
  semian_identifier = "nethttp_#{host}_#{port}"
  if !http_config[Rails.env].key?(semian_identifier)
    [semian_identifier, http_config[Rails.env]["nethttp_default"]]
  else
    [semian_identifier, http_config[Rails.env][semian_identifier]]
  end
end

# If we don't want default semian options, 
# i.e. adapter is disabled unless parameters are given:
Semian::NetHTTP.semian_configuration = proc do |host, port|
  semian_identifier = "nethttp_#{host}_#{port}"
  [semian_identifier, http_config[Rails.env][semian_identifier]] # nil means disabled
end

# If we want to disable completely
Semian::NetHTTP.raw_semian_options = proc do |semian_identifier|
  nil
end

# If we need extra errors, concat more onto array
# assert_equal(Semian::NetHTTP.exceptions, Semian::NetHTTP::DEFAULT_ERRORS)
Semian::NetHTTP.exceptions.concat([::Resolv::ResolvError, ::OpenURI::HTTPError, ::OpenSSL::SSL::SSLError])
```

Review? @Sirupsen @byroot

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,8 +144,7 @@ SEMIAN_PARAMETERS = { tickets: 1,
 Semian::NetHTTP.semian_configuration = proc do |host, port|
   # Let's make it only active for github.com
   if host == "github.com" && port == "80"
-    SEMIAN_PARAMETERS.dup.tap do |options|
-      options[:identifier] = nethttp_github.com_80
+    SEMIAN_PARAMETERS.merge(identifier: "nethttp_github.com_80")
     end
   else
     nil
--- a/README.md
+++ b/README.md
@@ -133,31 +133,41 @@ HTTP connections on the user's behalf, the parameters are instead provided
 by calling specific functions in `Semian::NetHTTP`, perhaps in an initialization file.
 
 To give Semian parameters, assign a `proc` to `Semian::NetHTTP.semian_configuration`
-that takes two parameters, `host` and `port like `127.0.0.1` and `80` or `github_com` and `80`,
-and returns two values, a `semian_identifier` and a `Hash` with the keys as follows.
+that takes a two parameters, `host` and `port` like `127.0.0.1` and `80` or `github_com` and `80`,
+and returns a `Hash` with keys as follows.
 
 ```ruby
-SEMIAN_OPTIONS = { tickets: 1,
-                   success_threshold: 1,
-                   error_threshold: 3,
-                   error_timeout: 10 }
-Semian::NetHTTP.semian_configuration = proc do |host, port|
-  semian_identifier = "nethttp_#{host}_#{port}"
+SEMIAN_PARAMETERS = { tickets: 1,
+                      success_threshold: 1,
+                      error_threshold: 3,
+                      error_timeout: 10 }
+Semian::NetHTTP.raw_semian_options = proc do |host, port|
   # Let's make it only active for github.com
-  if semian_identifier == "nethttp_github.com_80"
-    [semian_identifier, SEMIAN_OPTIONS]
+  if host == "github.com" && port == "80"
+    SEMIAN_PARAMETERS.dup.tap do |options|
+      options[:name] = nethttp_github.com_80
+    end
   else
-    [semian_identifier, nil]
+    nil
   end
 end
 
 # Called from within API:
-# semian_identifier, semian_options = Semian::NetHTTP.semian_configuration("github.com", 80)
+# configuration = Semian::NetHTTP.semian_configuration("github.com", 80)
+# semian_identifier = configuration[:name]
+# semian_options = configuration.delete(:name) || configuration
 ```
 
-The `semian_identifier` used as the name of the resource, and the `semian_options` are the
-configuration parameters. A return value of `nil` means Semian is disabled for that
-`Net::HTTP` endpoint. This works well since the result of a failed Hash lookup is `nil` also.
+It is important to carefully choose the `name` provided, since it is used to identify,
+track, and fail quickly, if necessary, the resource it protects. The `name` specifies
+the `semian_identifier` of the resource for which the given `semian_options` apply.
+For most purposes, `"nethttp_#{host}_#{port}"` is a good default identifier.
+This can become useful when grouping related subdomains as one resource,
+so that they all contribute to the same circuit breaker and bulkhead state
+and fail quickly together.
+
+A return value of `nil` means Semian is disabled for that `Net::HTTP` endpoint.
+This works well since the result of a failed Hash lookup is `nil` also.
 As such, this particular adapter defaults to whitelisting, although the
 behavior can be changed to blacklisting or even be completely disabled by varying
 the use of returning `nil`.
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 ![](http://i.imgur.com/7Vn2ibF.png)
 
 Semian is a library for controlling access to slow or unresponsive external
-services to avoid cascading failures. 
+services to avoid cascading failures.
 
 When services are down they typically fail fast with errors like `ECONNREFUSED`
 and `ECONNRESET` which can be rescued in code. However, slow resources fail
@@ -71,6 +71,7 @@ version is the version of the public gem with the same name:
 
 * [`semian/mysql2`][mysql-semian-adapter] (~> 0.3.16)
 * [`semian/redis`][redis-semian-adapter] (~> 3.2.1)
+* [`semian/net_http`][nethttp-semian-adapter]
 
 ### Creating Adapters
 
@@ -124,6 +125,49 @@ client = Redis.new(semian: {
   error_threshold: 4,
   error_timeout: 20
 })
+```
+
+#### Net::HTTP
+For the `Net::HTTP` specific Semian adapter, since many external libraries may create
+HTTP connections on the user's behalf, the parameters are instead provided
+by calling specific functions in `Semian::NetHTTP`, perhaps in an initialization file.
+
+To give Semian parameters, assign a `proc` to `Semian::NetHTTP.raw_semian_options`
+that takes a `semian_identifier` like `http_127_0_0_1_80` or `http_github_com_80`,
+and returns a `Hash` with keys as follows.
+
+```ruby
+SEMIAN_PARAMETERS = { tickets: 1,
+                      success_threshold: 1,
+                      error_threshold: 3,
+                      error_timeout: 10 }
+Semian::NetHTTP.raw_semian_options = proc do |semian_identifier|
+  # Let's make it only active for github.com
+  if semian_identifier == "http_github_com_80"
+    SEMIAN_PARAMETERS
+  else
+    nil
+  end
+end
+```
+
+A return value of `nil` means Semian is disabled for that `Net::HTTP` endpoint.
+This works well since the result of a failed Hash lookup is `nil` also.
+As such, this particular adapter defaults to whitelisting, although the
+behavior can be changed to blacklisting or even be completely disabled by varying
+the use of returning `nil`.
+
+Since we envision this particular adapter can be used in combination with many
+external libraries, we added functionality to expand the Exceptions that can be
+tracked as part of Semian's circuit breaker. Add exceptions and reset to the
+[`default`][nethttp-default-errors] list using the following:
+
+```ruby
+# assert_equal(Semian::NetHTTP.exceptions, Semian::NetHTTP::DEFAULT_ERRORS)
+Semian::NetHTTP.concat_exceptions([::OpenSSL::SSL::SSLError])
+
+Semian::NetHTTP.reset_exceptions
+# assert_equal(Semian::NetHTTP.exceptions, Semian::NetHTTP::DEFAULT_ERRORS)
 ```
 
 # Understanding Semian
@@ -248,7 +292,7 @@ response time. This is the problem Semian solves by failing fast.
 
 Semian consists of two parts: circuit breaker and bulkheading. To understand
 Semian, and especially how to configure it, we must understand these patterns
-and their implementation. 
+and their implementation.
 
 ### Circuit Breaker
 
@@ -319,7 +363,7 @@ five workers to talk to a resource at any given point in time, and accept the
 a connection instantly. This means that while the five workers are waiting for a
 timeout, all the other `W-5` workers on the node will instantly be failing on
 checking out the connection and opening their circuits. Our capacity is only
-degraded by a relatively small amount. 
+degraded by a relatively small amount.
 
 We call this limitation primitive "tickets". In this case, the resource access
 is limited to 5 tickets (see Configuration). The timeout value specifies the
@@ -361,7 +405,7 @@ will affect the circuit and Semian, which can make future calls fail faster.
 ## Failing gracefully
 
 Ok, great, we've got a way to fail fast with slow resources, how does that make
-my application more resilient? 
+my application more resilient?
 
 Failing fast is only half the battle. It's up to you what you do with these
 errors, in the [session example](#real-world-example) we handle it gracefully by
@@ -496,13 +540,15 @@ non-IO.
 [hystrix]: https://github.com/Netflix/Hystrix
 [release-it]: https://pragprog.com/book/mnee/release-it
 [shopify]: http://www.shopify.com/
-[mysql-semian-adapter]: https://github.com/Shopify/semian/blob/master/lib/semian/mysql2.rb
-[redis-semian-adapter]: https://github.com/Shopify/semian/blob/master/lib/semian/redis.rb
-[semian-adapter]: https://github.com/Shopify/semian/blob/master/lib/semian/adapter.rb
-[semian-instrumentable]: https://github.com/Shopify/semian/blob/master/lib/semian/instrumentable.rb
+[mysql-semian-adapter]: lib/semian/mysql2.rb
+[redis-semian-adapter]: lib/semian/redis.rb
+[semian-adapter]: lib/semian/adapter.rb
+[nethttp-semian-adapter]: lib/semian/net_http.rb
+[nethttp-default-errors]: lib/semian/net_http.rb#L29-L39
+[semian-instrumentable]: lib/semian/instrumentable.rb
 [statsd-instrument]: http://github.com/shopify/statsd-instrument
 [resiliency-blog-post]: http://www.shopify.com/technology/16906928-building-and-testing-resilient-ruby-on-rails-applications
 [toxiproxy]: https://github.com/Shopify/toxiproxy
 [sysv]: http://man7.org/linux/man-pages/man7/svipc.7.html
 [cbp]: https://en.wikipedia.org/wiki/Circuit_breaker_design_pattern
-[namespaces]: http://man7.org/linux/man-pages/man7/namespaces.7.html 
+[namespaces]: http://man7.org/linux/man-pages/man7/namespaces.7.html
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ For the `Net::HTTP` specific Semian adapter, since many external libraries may c
 HTTP connections on the user's behalf, the parameters are instead provided
 by calling specific functions in `Semian::NetHTTP`, perhaps in an initialization file.
 
+##### Naming and Options
 To give Semian parameters, assign a `proc` to `Semian::NetHTTP.semian_configuration`
 that takes a two parameters, `host` and `port` like `127.0.0.1` and `80` or `github_com` and `80`,
 and returns a `Hash` with keys as follows.
@@ -144,38 +145,38 @@ SEMIAN_PARAMETERS = { tickets: 1,
 Semian::NetHTTP.semian_configuration = proc do |host, port|
   # Let's make it only active for github.com
   if host == "github.com" && port == "80"
-    SEMIAN_PARAMETERS.merge(identifier: "nethttp_github.com_80")
-    end
+    SEMIAN_PARAMETERS.merge(name: "github.com_80")
   else
     nil
   end
 end
 
 # Called from within API:
-# configuration = Semian::NetHTTP.semian_configuration("github.com", 80)
-# semian_identifier = configuration[:identifier]
-# semian_options = configuration.delete(:identifier) || configuration
+# semian_options = Semian::NetHTTP.semian_configuration("github.com", 80)
+# semian_identifier = "nethttp_#{semian_options[:name]}"
 ```
 
-It is important to carefully choose the `name` provided, since it is used to identify,
-track, and fail quickly, if necessary, the resource it protects. The `name` specifies
-the `semian_identifier` of the resource for which the given `semian_options` apply.
-For most purposes, `"nethttp_#{host}_#{port}"` is a good default identifier.
-This can become useful when grouping related subdomains as one resource,
-so that they all contribute to the same circuit breaker and bulkhead state
-and fail quickly together.
+The `name` should be carefully chosen since it identifies the resource being protected.
+The `semian_options` passed apply to that resource. Semian creates the `semian_identifier`
+from the `name` to look up and store changes in the circuit breaker and bulkhead states
+and associate successes, failures, errors with the protected resource.
 
-A return value of `nil` means Semian is disabled for that `Net::HTTP` endpoint.
-This works well since the result of a failed Hash lookup is `nil` also.
-As such, this particular adapter defaults to whitelisting, although the
+For most purposes, `"#{host}_#{port}"` is a good default `name`. Custom `name` formats
+can be useful to grouping related subdomains as one resource, so that they all
+contribute to the same circuit breaker and bulkhead state and fail together.
+
+A return value of `nil` for `semian_configuration` means Semian is disabled for that
+HTTP endpoint. This works well since the result of a failed Hash lookup is `nil` also.
+This behavior lets the adapter default to whitelisting, although the
 behavior can be changed to blacklisting or even be completely disabled by varying
-the use of returning `nil`.
+the use of returning `nil` in the assigned closure.
 
+##### Additional Exceptions
 Since we envision this particular adapter can be used in combination with many
-external libraries, we added functionality to expand the Exceptions that can be
-tracked as part of Semian's circuit breaker. This may be necessary for libraries
-that introduce new exceptions that want to be tracked. Add exceptions and reset
-to the [`default`][nethttp-default-errors] list using the following:
+external libraries, that can raise additional exceptions, we added functionality to
+expand the Exceptions that can be tracked as part of Semian's circuit breaker.
+This may be necessary for libraries that introduce new exceptions or re-raise them.
+Add exceptions and reset to the [`default`][nethttp-default-errors] list using the following:
 
 ```ruby
 # assert_equal(Semian::NetHTTP.exceptions, Semian::NetHTTP::DEFAULT_ERRORS)
@@ -559,7 +560,7 @@ non-IO.
 [redis-semian-adapter]: lib/semian/redis.rb
 [semian-adapter]: lib/semian/adapter.rb
 [nethttp-semian-adapter]: lib/semian/net_http.rb
-[nethttp-default-errors]: lib/semian/net_http.rb#L29-L39
+[nethttp-default-errors]: lib/semian/net_http.rb#L33-L43
 [semian-instrumentable]: lib/semian/instrumentable.rb
 [statsd-instrument]: http://github.com/shopify/statsd-instrument
 [resiliency-blog-post]: http://www.shopify.com/technology/16906928-building-and-testing-resilient-ruby-on-rails-applications
--- a/README.md
+++ b/README.md
@@ -141,11 +141,11 @@ SEMIAN_PARAMETERS = { tickets: 1,
                       success_threshold: 1,
                       error_threshold: 3,
                       error_timeout: 10 }
-Semian::NetHTTP.raw_semian_options = proc do |host, port|
+Semian::NetHTTP.semian_configuration = proc do |host, port|
   # Let's make it only active for github.com
   if host == "github.com" && port == "80"
     SEMIAN_PARAMETERS.dup.tap do |options|
-      options[:name] = nethttp_github.com_80
+      options[:identifier] = nethttp_github.com_80
     end
   else
     nil
@@ -154,8 +154,8 @@ end
 
 # Called from within API:
 # configuration = Semian::NetHTTP.semian_configuration("github.com", 80)
-# semian_identifier = configuration[:name]
-# semian_options = configuration.delete(:name) || configuration
+# semian_identifier = configuration[:identifier]
+# semian_options = configuration.delete(:identifier) || configuration
 ```
 
 It is important to carefully choose the `name` provided, since it is used to identify,
@@ -174,12 +174,13 @@ the use of returning `nil`.
 
 Since we envision this particular adapter can be used in combination with many
 external libraries, we added functionality to expand the Exceptions that can be
-tracked as part of Semian's circuit breaker. Add exceptions and reset to the
-[`default`][nethttp-default-errors] list using the following:
+tracked as part of Semian's circuit breaker. This may be necessary for libraries
+that introduce new exceptions that want to be tracked. Add exceptions and reset
+to the [`default`][nethttp-default-errors] list using the following:
 
 ```ruby
 # assert_equal(Semian::NetHTTP.exceptions, Semian::NetHTTP::DEFAULT_ERRORS)
-Semian::NetHTTP.concat_exceptions([::OpenSSL::SSL::SSLError])
+Semian::NetHTTP.exceptions += [::OpenSSL::SSL::SSLError]
 
 Semian::NetHTTP.reset_exceptions
 # assert_equal(Semian::NetHTTP.exceptions, Semian::NetHTTP::DEFAULT_ERRORS)
--- a/README.md
+++ b/README.md
@@ -132,27 +132,32 @@ For the `Net::HTTP` specific Semian adapter, since many external libraries may c
 HTTP connections on the user's behalf, the parameters are instead provided
 by calling specific functions in `Semian::NetHTTP`, perhaps in an initialization file.
 
-To give Semian parameters, assign a `proc` to `Semian::NetHTTP.raw_semian_options`
-that takes a `semian_identifier` like `http_127_0_0_1_80` or `http_github_com_80`,
-and returns a `Hash` with keys as follows.
+To give Semian parameters, assign a `proc` to `Semian::NetHTTP.semian_configuration`
+that takes two parameters, `host` and `port like `127.0.0.1` and `80` or `github_com` and `80`,
+and returns two values, a `semian_identifier` and a `Hash` with the keys as follows.
 
 ```ruby
-SEMIAN_PARAMETERS = { tickets: 1,
-                      success_threshold: 1,
-                      error_threshold: 3,
-                      error_timeout: 10 }
-Semian::NetHTTP.raw_semian_options = proc do |semian_identifier|
+SEMIAN_OPTIONS = { tickets: 1,
+                   success_threshold: 1,
+                   error_threshold: 3,
+                   error_timeout: 10 }
+Semian::NetHTTP.semian_configuration = proc do |host, port|
+  semian_identifier = "nethttp_#{host}_#{port}"
   # Let's make it only active for github.com
-  if semian_identifier == "http_github_com_80"
-    SEMIAN_PARAMETERS
+  if semian_identifier == "nethttp_github.com_80"
+    [semian_identifier, SEMIAN_OPTIONS]
   else
-    nil
+    [semian_identifier, nil]
   end
 end
+
+# Called from within API:
+# semian_identifier, semian_options = Semian::NetHTTP.semian_configuration("github.com", 80)
 ```
 
-A return value of `nil` means Semian is disabled for that `Net::HTTP` endpoint.
-This works well since the result of a failed Hash lookup is `nil` also.
+The `semian_identifier` used as the name of the resource, and the `semian_options` are the
+configuration parameters. A return value of `nil` means Semian is disabled for that
+`Net::HTTP` endpoint. This works well since the result of a failed Hash lookup is `nil` also.
 As such, this particular adapter defaults to whitelisting, although the
 behavior can be changed to blacklisting or even be completely disabled by varying
 the use of returning `nil`.
--- a/lib/semian/net_http.rb
+++ b/lib/semian/net_http.rb
@@ -42,6 +42,15 @@ module Semian
       ::SystemCallError, # includes ::Errno::EINVAL, ::Errno::ECONNRESET, ::Errno::ECONNREFUSED, ::Errno::ETIMEDOUT, and more
     ].freeze # Net::HTTP can throw many different errors, this tries to capture most of them
 
+    # The naked methods are exposed as `raw_query` and `raw_connect` for instrumentation purpose
+    def self.included(base)
+      base.send(:alias_method, :raw_request, :request)
+      base.send(:remove_method, :request)
+
+      base.send(:alias_method, :raw_connect, :connect)
+      base.send(:remove_method, :connect)
+    end
+
     class << self
       attr_accessor :semian_configuration
       attr_accessor :exceptions
@@ -72,19 +81,15 @@ module Semian
     end
 
     def connect
-      return super if disabled?
-      acquire_semian_resource(adapter: :http, scope: :connection) do
-        super
-      end
+      return raw_connect if disabled?
+      acquire_semian_resource(adapter: :http, scope: :connection) { raw_connect }
     end
 
-    def request(*req)
-      return super if disabled?
-      acquire_semian_resource(adapter: :http, scope: :query) do
-        super
-      end
+    def request(req, body = nil, &block)
+      return raw_request(req, body, &block) if disabled?
+      acquire_semian_resource(adapter: :http, scope: :query) { raw_request(req, body, &block) }
     end
   end
 end
 
-Net::HTTP.prepend(Semian::NetHTTP)
+Net::HTTP.include(Semian::NetHTTP)
--- a/lib/semian/net_http.rb
+++ b/lib/semian/net_http.rb
@@ -22,8 +22,12 @@ module Semian
     ResourceBusyError = ::Net::ResourceBusyError
     CircuitOpenError = ::Net::CircuitOpenError
 
+    def semian_configuration
+      Semian::NetHTTP.retrieve_semian_configuration(address, port)
+    end
+
     def semian_identifier
-      Semian::NetHTTP.retrieve_semian_configuration(address, port)[:identifier]
+      "nethttp_#{semian_configuration[:name]}"
     end
 
     DEFAULT_ERRORS = [
@@ -54,11 +58,8 @@ module Semian
     Semian::NetHTTP.reset_exceptions
 
     def raw_semian_options
-      options = Semian::NetHTTP.retrieve_semian_configuration(address, port)
-      unless options.nil?
-        options = options.dup
-        options.delete(:identifier)
-      end
+      options = semian_configuration
+      options = options.dup unless options.nil?
       options
     end
 
@@ -67,7 +68,7 @@ module Semian
     end
 
     def disabled?
-      raw_semian_options.nil?
+      semian_configuration.nil?
     end
 
     def connect
--- a/lib/semian/net_http.rb
+++ b/lib/semian/net_http.rb
@@ -55,8 +55,10 @@ module Semian
       ::SystemCallError, # includes ::Errno::EINVAL, ::Errno::ECONNRESET, ::Errno::ECONNREFUSED, ::Errno::ETIMEDOUT, and more
     ].freeze # Net::HTTP can throw many different errors, this tries to capture most of them
 
+    Semian::NetHTTP.exceptions = Semian::NetHTTP::DEFAULT_ERRORS.dup
+
     def resource_exceptions
-      Semian::NetHTTP.exceptions ||= Semian::NetHTTP::DEFAULT_ERRORS.dup
+      Semian::NetHTTP.exceptions
     end
 
     def enabled?
--- a/lib/semian/net_http.rb
+++ b/lib/semian/net_http.rb
@@ -1,0 +1,85 @@
+require 'semian'
+require 'semian/adapter'
+require 'net/https'
+require 'open-uri'
+
+module Net
+  ::Net::ProtocolError.include(::Semian::AdapterError)
+
+  class SemianError < ::Net::ProtocolError
+    def initialize(semian_identifier, *args)
+      super(*args)
+      @semian_identifier = semian_identifier
+    end
+  end
+
+  Net::ResourceBusyError = Class.new(SemianError)
+  Net::CircuitOpenError = Class.new(SemianError)
+end
+
+module Semian
+  module NetHTTP
+    include Semian::Adapter
+
+    ResourceBusyError = ::Net::ResourceBusyError
+    CircuitOpenError = ::Net::CircuitOpenError
+
+    DEFAULT_SEMIAN_OPTIONS = @raw_semian_options = {
+      tickets: 3,
+      success_threshold: 1,
+      error_threshold: 3,
+      error_timeout: 10,
+    }
+
+    def semian_identifier
+      "http_#{address.tr('.', '_')}_#{port}"
+    end
+
+    def self.raw_semian_options(semian_identifier = "http_default".freeze)
+      if @raw_semian_options.respond_to?(:call)
+        @raw_semian_options.call(semian_identifier)
+      else
+        @raw_semian_options ||= DEFAULT_SEMIAN_OPTIONS
+      end
+    end
+
+    def self.raw_semian_options=(options)
+      @raw_semian_options = options
+    end
+
+    def raw_semian_options
+      Semian::NetHTTP.raw_semian_options(semian_identifier)
+    end
+
+    def resource_exceptions
+      @exceptions ||= [
+        ::Timeout::Error, # includes ::Net::ReadTimeout and ::Net::OpenTimeout
+        ::TimeoutError,
+        ::Net::ProtocolError,
+        ::Net::HTTPBadResponse,
+        ::Net::HTTPHeaderSyntaxError,
+        ::SocketError,
+        ::IOError,
+        ::EOFError,
+        ::Resolv::ResolvError,
+        ::OpenURI::HTTPError,
+        ::OpenSSL::SSL::SSLError,
+        ::SystemCallError, # includes ::Errno::EINVAL, ::Errno::ECONNRESET, ::Errno::ECONNREFUSED, ::Errno::ETIMEDOUT, and more
+      ]
+    end
+
+    def connect
+      acquire_semian_resource(adapter: :nethttp, scope: :connection) do
+        super
+      end
+    end
+
+    def request(*req)
+      acquire_semian_resource(adapter: :nethttp, scope: :query) do
+        super
+      end
+    end
+  end
+end
+
+Net::HTTP.send(:prepend, Semian::NetHTTP)
--- a/lib/semian/net_http.rb
+++ b/lib/semian/net_http.rb
@@ -23,7 +23,7 @@ module Semian
     CircuitOpenError = ::Net::CircuitOpenError
 
     def semian_identifier
-      @identifier ||= Semian::NetHTTP.retrieve_semian_configuration_by_host_port(address, port)[:name]
+      Semian::NetHTTP.retrieve_semian_configuration(address, port)[:identifier]
     end
 
     DEFAULT_ERRORS = [
@@ -42,45 +42,43 @@ module Semian
       attr_accessor :semian_configuration
       attr_accessor :exceptions
 
-      def retrieve_semian_configuration_by_host_port(host, port)
-        proc_result = @semian_configuration.call(host, port) if @semian_configuration.respond_to?(:call)
-        proc_result ||= {}
-        proc_result[:name] ||= "nethttp_#{host}_#{port}"
-        proc_result
+      def retrieve_semian_configuration(host, port)
+        @semian_configuration.call(host, port) if @semian_configuration.respond_to?(:call)
       end
 
       def reset_exceptions
         self.exceptions = Semian::NetHTTP::DEFAULT_ERRORS.dup
-      end
-
-      def concat_exceptions(exceptions)
-        self.exceptions.concat(exceptions)
       end
     end
 
     Semian::NetHTTP.reset_exceptions
 
     def raw_semian_options
-      @options ||= Semian::NetHTTP.retrieve_semian_configuration_by_host_port(address, port).dup.tap { |o| o.delete(:name) }
+      options = Semian::NetHTTP.retrieve_semian_configuration(address, port)
+      unless options.nil?
+        options = options.dup
+        options.delete(:identifier)
+      end
+      options
     end
 
     def resource_exceptions
       Semian::NetHTTP.exceptions
     end
 
-    def enabled?
-      !raw_semian_options.empty?
+    def disabled?
+      raw_semian_options.nil?
     end
 
     def connect
-      return super unless enabled?
+      return super if disabled?
       acquire_semian_resource(adapter: :http, scope: :connection) do
         super
       end
     end
 
     def request(*req)
-      return super unless enabled?
+      return super if disabled?
       acquire_semian_resource(adapter: :http, scope: :query) do
         super
       end
--- a/lib/semian/net_http.rb
+++ b/lib/semian/net_http.rb
@@ -23,7 +23,8 @@ module Semian
     CircuitOpenError = ::Net::CircuitOpenError
 
     def semian_identifier
-      "http_#{address.tr('.', '_')}_#{port}"
+      @configuration ||= Semian::NetHTTP.retrieve_semian_configuration_by_host_port(address, port)
+      @configuration[0]
     end
 
     DEFAULT_ERRORS = [
@@ -39,14 +40,14 @@ module Semian
     ].freeze # Net::HTTP can throw many different errors, this tries to capture most of them
 
     class << self
-      attr_accessor :raw_semian_options
+      attr_accessor :semian_configuration
       attr_accessor :exceptions
 
-      def retrieve_semian_options_by_identifier(semian_identifier)
-        if @raw_semian_options.respond_to?(:call)
-          @raw_semian_options.call(semian_identifier)
+      def retrieve_semian_configuration_by_host_port(host, port)
+        if @semian_configuration.respond_to?(:call)
+          @semian_configuration.call(host, port) || ["nethttp_#{host}_#{port}", nil]
         else
-          @raw_semian_options ||= nil # disabled by default
+          ["nethttp_#{host}_#{port}", nil] # adapter is disabled by default
         end
       end
 
@@ -62,7 +63,8 @@ module Semian
     Semian::NetHTTP.reset_exceptions
 
     def raw_semian_options
-      Semian::NetHTTP.retrieve_semian_options_by_identifier(semian_identifier)
+      @configuration ||= Semian::NetHTTP.retrieve_semian_configuration_by_host_port(address, port)
+      @configuration[1]
     end
 
     def resource_exceptions
--- a/lib/semian/net_http.rb
+++ b/lib/semian/net_http.rb
@@ -23,8 +23,7 @@ module Semian
     CircuitOpenError = ::Net::CircuitOpenError
 
     def semian_identifier
-      @configuration ||= Semian::NetHTTP.retrieve_semian_configuration_by_host_port(address, port)
-      @configuration[0]
+      @identifier ||= Semian::NetHTTP.retrieve_semian_configuration_by_host_port(address, port)[:name]
     end
 
     DEFAULT_ERRORS = [
@@ -44,11 +43,10 @@ module Semian
       attr_accessor :exceptions
 
       def retrieve_semian_configuration_by_host_port(host, port)
-        if @semian_configuration.respond_to?(:call)
-          @semian_configuration.call(host, port) || ["nethttp_#{host}_#{port}", nil]
-        else
-          ["nethttp_#{host}_#{port}", nil] # adapter is disabled by default
-        end
+        proc_result = @semian_configuration.call(host, port) if @semian_configuration.respond_to?(:call)
+        proc_result ||= {}
+        proc_result[:name] ||= "nethttp_#{host}_#{port}"
+        proc_result
       end
 
       def reset_exceptions
@@ -63,8 +61,7 @@ module Semian
     Semian::NetHTTP.reset_exceptions
 
     def raw_semian_options
-      @configuration ||= Semian::NetHTTP.retrieve_semian_configuration_by_host_port(address, port)
-      @configuration[1]
+      @options ||= Semian::NetHTTP.retrieve_semian_configuration_by_host_port(address, port).dup.tap { |o| o.delete(:name) }
     end
 
     def resource_exceptions
@@ -72,7 +69,7 @@ module Semian
     end
 
     def enabled?
-      !raw_semian_options.nil?
+      !raw_semian_options.empty?
     end
 
     def connect
--- a/lib/semian/net_http.rb
+++ b/lib/semian/net_http.rb
@@ -56,7 +56,7 @@ module Semian
     ].freeze # Net::HTTP can throw many different errors, this tries to capture most of them
 
     def resource_exceptions
-      Semian::NetHTTP.exceptions ||= DEFAULT_ERRORS.dup
+      Semian::NetHTTP.exceptions ||= Semian::NetHTTP::DEFAULT_ERRORS.dup
     end
 
     def enabled?
@@ -65,14 +65,14 @@ module Semian
 
     def connect
       return super unless enabled?
-      acquire_semian_resource(adapter: :nethttp, scope: :connection) do
+      acquire_semian_resource(adapter: :http, scope: :connection) do
         super
       end
     end
 
     def request(*req)
       return super unless enabled?
-      acquire_semian_resource(adapter: :nethttp, scope: :query) do
+      acquire_semian_resource(adapter: :http, scope: :query) do
         super
       end
     end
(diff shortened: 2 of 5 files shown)


lib/semian/net_http.rb
```diff
@@ -0,0 +1,85 @@
+require 'semian'
+require 'semian/adapter'
+require 'net/https'
+require 'open-uri'
```
byroot: I believe this is a mistake.

kyewei: `open-uri`? It's needed to catch the `::OpenURI::HTTPError`

byroot: But why would you have theses? `open-uri` is nice for quick and dirty script, but it's the last thing you want to be required in your application.

kyewei: `app/controllers/admin/settings/files_controller.rb` needs it.

lib/semian/net_http.rb
```diff
@@ -0,0 +1,85 @@
...
+    def raw_semian_options
+      Semian::NetHTTP.raw_semian_options(semian_identifier)
+    end
+
+    def resource_exceptions
+      @exceptions ||= [
+        ::Timeout::Error, # includes ::Net::ReadTimeout and ::Net::OpenTimeout
+        ::TimeoutError,
+        ::Net::ProtocolError,
+        ::Net::HTTPBadResponse,
```
byroot: BadResponse if 400 right? I don't think semain should care about anything is the 4XX range.

kyewei: Again, I took that from core code. See `003_network_exceptions.rb`, which lists a large amount of them. Being able to add additional errors seems like a good idea I think.

lib/semian/net_http.rb
byroot: > Perhaps I'll add code that lets you add more errors?

Sounds good to me.

Otherwise another thing you can do it to register the exception name, not a reference:

```ruby
error.class.name == 'OpenURI::HTTPError'
```

It doesn't work with subclasses, but allow you to handle exceptions that shouldn't be loaded by Semian.

lib/semian/net_http.rb
byroot: This is an open source project, not a sub part of Shopify core. You have to think generically. If Shopify need customization, then we make it possible. But we shouldn't force our legacy code on anyone.

Please start from an empty list and populate it with things that make sense in isolation.

byroot: By that I mean, generate all kind of possible network issues with Toxiproxy and record those exceptions.

From there when can debate what else should be added. The 5xx range might be a sensible addition.

kyewei: Sounds good.

kyewei: Also, `BadResponse` is a protocl error from what this says anyways ... https://github.com/ruby/ruby/blob/trunk/lib/net/http.rb#L382, 400 is `BadRequest`

byroot: Semian should only consider network and server errors.

If it was counting client errors, a single malformed query could mark the resource as dead.

Sometimes it's not possible to differenciate client and server errors, but when we can, we should disregard client errors.

lib/semian/net_http.rb
```diff
@@ -0,0 +1,85 @@
+require 'semian'
+require 'semian/adapter'
+require 'net/https'
+require 'open-uri'
+
+module Net
+  ::Net::ProtocolError.include(::Semian::AdapterError)
```
fw42: nitpick: I've never seen it done like this, seems a little bit unconventional. I would explicitly open the class... but no big deal.

lib/semian/net_http.rb
```diff
@@ -0,0 +1,85 @@
...
+end
+
+module Semian
+  module NetHTTP
+    include Semian::Adapter
+
+    ResourceBusyError = ::Net::ResourceBusyError
+    CircuitOpenError = ::Net::CircuitOpenError
+
+    DEFAULT_SEMIAN_OPTIONS = @raw_semian_options = {
```
fw42: This is confusing, you have a class instance variable here called `@raw_semian_options` and below you have an instance variable which is also called `@raw_semian_options`. Those two are not the same! I'm gonna assume that wasn't on purpose?

fw42: oh sorry, misread, below are class methods... hmm.. still, I find this confusing, especially with line 42 which says `@raw_semian_options ||= DEFAULT_SEMIAN_OPTIONS`. I'm not sure I understand the logic. If you define it here, why would it ever be not defined?

lib/semian/net_http.rb
```diff
@@ -0,0 +1,85 @@
...
+      tickets: 3,
+      success_threshold: 1,
+      error_threshold: 3,
+      error_timeout: 10,
+    }
```
fw42: Since this is being assigned to a constant, can you `.freeze` the hash so that nobody can mutate it?

lib/semian/net_http.rb
```diff
@@ -0,0 +1,85 @@
...
+
+    def semian_identifier
+      "http_#{address.tr('.', '_')}_#{port}"
```
fw42: Hmm I'm a little worried that this will create a LOT of different SysV semaphores (like, tens of thousands per day at least, in Shopify core). Is that fine? Is there no limit in the kernel for how many there can be? @Sirupsen?

lib/semian/net_http.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+
+    def self.raw_semian_options(semian_identifier = "http_default".freeze)
+      if @raw_semian_options.respond_to?(:call)
+        @raw_semian_options.call(semian_identifier)
+      else
+        @raw_semian_options ||= DEFAULT_SEMIAN_OPTIONS
```
fw42: you might want to `.dup` the constant hash, otherwise someone might try to mutate it and cause weirdness...

lib/semian/net_http.rb
```diff
@@ -0,0 +1,85 @@
...
+        ::TimeoutError,
+        ::Net::ProtocolError,
+        ::Net::HTTPBadResponse,
+        ::Net::HTTPHeaderSyntaxError,
+        ::SocketError,
+        ::IOError,
+        ::EOFError,
+        ::Resolv::ResolvError,
+        ::OpenURI::HTTPError,
+        ::OpenSSL::SSL::SSLError,
```
fw42: Hmmmm... not super sure about this one..

lib/semian/net_http.rb
```diff
@@ -0,0 +1,85 @@
...
+
+    def request(*req)
+      acquire_semian_resource(adapter: :nethttp, scope: :query) do
+        super
+      end
+    end
+  end
+end
+
+Net::HTTP.send(:prepend, Semian::NetHTTP)
```
fw42: why are you using `.send` here? is this a Ruby 2.0 compatibility thing (where `prepend` was private)? Because it's public in >2.1 iirc.

test/net_http_test.rb
```diff
@@ -0,0 +1,218 @@
...
+      Net::HTTP.get(uri)
+    end
+  end
+
+  def test_trigger_close_after_open
+    open_circuit!
+    close_circuit!
+  end
+
+  def test_blockheads_tickets_are_working
```
fw42: blockhead? did you mean bulkhead?

test/net_http_test.rb
```diff
@@ -0,0 +1,218 @@
...
+    ticket_count = Net::HTTP.new("localhost", 41_235).raw_semian_options[:tickets]
+
+    @proxy.downstream(:latency, latency: 500).apply do
+      uri = URI('http://localhost:41235/200')
+      (ticket_count).times do
+        threads << Thread.new do
+          Net::HTTP.get(uri)
+        end
+      end
+      sleep 0.3
```
fw42: what's the sleep for? is there a way to get rid of this?

fw42: can't you `join` the threads first and avoid the sleep that way?

test/net_http_test.rb
```diff
@@ -0,0 +1,218 @@
...
+      open_circuit!
+      assert_raises Net::CircuitOpenError do
+        request = Net::HTTP::Get.new uri
+        http.request(request)
+      end
+      assert_raises Net::CircuitOpenError do
+        request = Net::HTTP::Post.new uri
+        http.request(request)
+      end
+      # and so on...
```
fw42: ?

test/net_http_test.rb
```diff
@@ -0,0 +1,218 @@
...
+    end
+  end
+
+  def test_custom_raw_semian_options_work
+    orig_semian_options = Semian::NetHTTP.raw_semian_options
+    yaml_sample_config = {}
+    yaml_sample_config["development"] = {}
+    yaml_sample_config["development"]["http_default"] = {"tickets" => 1,
+                                                         "success_threshold" => 1,
```
fw42: can you use 2 space indentation, not 42 space indentation? :-)

test/net_http_test.rb
```diff
@@ -0,0 +1,218 @@
...
+      end
+    end
+    Net::HTTP.start("localhost", 41_235) do |http|
+      assert_equal yaml_sample_config["development"][http.semian_identifier], http.raw_semian_options
+    end
+    Net::HTTP.start("localhost", 41_234) do |http|
+      assert_equal yaml_sample_config["development"]["http_default"], http.raw_semian_options
+    end
+    assert_equal yaml_sample_config["development"]["http_default"], Semian::NetHTTP.raw_semian_options
+  ensure
```
fw42: can you extract this pattern into a `with_semian_options(...) { ... }` helper?

lib/semian/net_http.rb
kyewei: Use `ipcs -l` to check current limits. However, we can change it through some modifiying of `/proc/sys/kernel/sem` I think.

lib/semian/net_http.rb
kyewei: So are you suggesting something like `Net::HTTP.prepend(Semian::NetHTTP)` ? It seems to work, so I can change it to that if there are not compatibility issues.

lib/semian/net_http.rb
kyewei: Yup, it was suggested earlier that I make a smaller set of default errors and then provide a way to add additional ones after.

test/net_http_test.rb
kyewei: .. I do. I'll fix that.

test/net_http_test.rb
kyewei: There were a whole lot more, `Get`, `Post`, `Put`, `Head`, etc. I suppose I can list the ones in an array and `#each` over them.

test/net_http_test.rb
kyewei: I have to keep at least `ticket_count` requests blocked while running a `ticket_count+1`th request. The 500 ms latency keeps it blocked while after 300 ms the `ticket_count+1`th request runs.

fw42: This seems racy and fragile :-(

kyewei: It does. I couldn't figure out why just doing `ticket_count+1` threads doesn't raise anything so...

lib/semian/net_http.rb
byroot: I do this all the time.

lib/semian/net_http.rb
```diff
@@ -0,0 +1,90 @@
+require 'semian'
+require 'semian/adapter'
+
+module Net
+  ::Net::ProtocolError.include(::Semian::AdapterError)
```
Sirupsen: `::Net::ProtocolError` is the top-level? Can you prove it somehow?

lib/semian/net_http.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  class SemianError < ::Net::ProtocolError
+    def initialize(semian_identifier, *args)
+      super(*args)
+      @semian_identifier = semian_identifier
+    end
+  end
+
+  Net::ResourceBusyError = Class.new(SemianError)
+  Net::CircuitOpenError = Class.new(SemianError)
```
Sirupsen: Can you just do `ResourceBusyError` instead of prefixing with the namespace you're already in?

lib/semian/net_http.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    DEFAULT_SEMIAN_OPTIONS = {
+      tickets: 3,
+      success_threshold: 1,
+      error_threshold: 3,
+      error_timeout: 10,
+    }.freeze
+
+    def semian_identifier
+      "http_#{address.tr('.', '_')}_#{port}"
```
Sirupsen: Why does it matter to do `#tr`? This is hashed anyway.

lib/semian/net_http.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+
+    def self.raw_semian_options(semian_identifier = "http_default".freeze)
```
Sirupsen: Why have a default?

lib/semian/net_http.rb
```diff
@@ -0,0 +1,90 @@
...
+        @raw_semian_options ||= DEFAULT_SEMIAN_OPTIONS.dup
+      end
+    end
+
+    def self.raw_semian_options=(options)
+      @raw_semian_options = options
+    end
+
+    def raw_semian_options
+      Semian::NetHTTP.raw_semian_options(semian_identifier)
```
Sirupsen: I'd do `self.class` instead.

lib/semian/net_http.rb
```diff
@@ -0,0 +1,90 @@
...
+      ::Timeout::Error, # includes ::Net::ReadTimeout and ::Net::OpenTimeout
+      ::TimeoutError, # alias for above
+      ::SocketError,
+      ::Net::HTTPBadResponse,
+      ::Net::HTTPHeaderSyntaxError,
+      ::Net::ProtocolError,
+      ::EOFError,
+      ::IOError,
+      ::SystemCallError, # includes ::Errno::EINVAL, ::Errno::ECONNRESET, ::Errno::ECONNREFUSED, ::Errno::ETIMEDOUT, and more
+    ].freeze
```
Sirupsen: I'd attach a comment why you need to do this, because it's obvious it sucks.

lib/semian/net_http.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    def self.custom_exceptions
+      @custom_exceptions ||= []
+    end
+
+    def self.custom_exceptions=(exceptions)
+      @custom_exceptions = (DEFAULT_ERRORS + exceptions).uniq - DEFAULT_ERRORS
```
Sirupsen: Instead you should allow appending, e.g. do `push_exceptions` or something. That's more intuitive.

test/net_http_test.rb
```diff
@@ -0,0 +1,241 @@
...
+        raise "Couldn't reach the service on port #{port} after #{time_to_wait}s"
+      else
+        retry
+      end
+    end
+  end
+
+  def after_tests
+    self.class.server_thread.kill
+  end
```
Sirupsen: I'm not a big fan of this being global for the entire test suite. I'd prefer having a server that's available for the duration of the block, similar to [what we do in `nginx-routing-modules`](https://github.com/Shopify/nginx-routing-modules/blob/master/test/helpers/rack_test_helper.rb).

test/net_http_test.rb
```diff
@@ -0,0 +1,241 @@
...
+
+  def test_adding_custom_errors_work
+    with_custom_errors([::OpenSSL::SSL::SSLError]) do
+      http = Net::HTTP.new("localhost", 41_235)
+      assert http.resource_exceptions.include?(::OpenSSL::SSL::SSLError)
+    end
+  end
+
+  def with_semian_options(options)
```
Sirupsen: All of these methods should be `private` to make it clear where the tests stop and the helpers start.

test/net_http_test.rb
```diff
@@ -0,0 +1,241 @@
...
+      response_code = '200' if response_code == ""
+      [response_code, {'Content-Type' => 'text/html'}, ['Success']]
+    end
+  end
+
+  def setup
+    # create a simple rack server so we can use toxiproxy
+    start_server_singleton
+    Semian["http_localhost_41234"].reset if Semian["http_localhost_41234"]
+    Semian["http_localhost_41235"].reset if Semian["http_localhost_41235"]
```
Sirupsen: These ports are in the Linux ephemeral port range:

```
simon@sb1.chi2:~ 1$ cat /proc/sys/net/ipv4/ip_local_port_range 
32768	61000
```

This means you can experience intermittent failures because an ephemeral port may already be started on this port. Just make it `< 32768` and you're good.

test/net_http_test.rb
```diff
@@ -0,0 +1,241 @@
...
+
+    uri = URI('http://localhost:41235/200')
+    @proxy.downstream(:latency, latency: 300).apply do
+      time = Time.now
+      Net::HTTP.get(uri)
+
+      time_diff = Time.now - time
+      assert time_diff >= 0.3
+    end
+  end
```
Sirupsen: You don't need to write a test for Toxiproxy.

test/net_http_test.rb
```diff
@@ -0,0 +1,241 @@
...
+      end
+    end
+  end
+
+  def close_circuit!
+    http = Net::HTTP.new("localhost", 41_235)
+    Timecop.travel(http.raw_semian_options[:error_timeout] + 1)
+    # Cause successes success_threshold times so circuit closes
+    http.raw_semian_options[:success_threshold].times do
+      http.get("/200")
```
Sirupsen: You should assert for success here.

test/net_http_test.rb
```diff
@@ -0,0 +1,241 @@
...
+            http.request(request)
+          end
+        end
+      end
+    end
+  end
+
+  def close_circuit!
+    http = Net::HTTP.new("localhost", 41_235)
+    Timecop.travel(http.raw_semian_options[:error_timeout] + 1)
```
Sirupsen: Wouldn't it be enough to travel to just the `error_timeout`? Do you need the 1?

test/net_http_test.rb
```diff
@@ -0,0 +1,241 @@
...
+          assert_raises Net::ReadTimeout do
+            http.request(request)
+          end
+        end
+      end
+    end
+  end
+
+  def close_circuit!
+    http = Net::HTTP.new("localhost", 41_235)
```
Sirupsen: Can we make this port a default argument instead?

test/net_http_test.rb
```diff
@@ -0,0 +1,241 @@
...
+  ensure
+    Semian::NetHTTP.custom_exceptions = orig_errors
+  end
+
+  def get_subclasses(klass)
+    ObjectSpace.each_object(klass.singleton_class).to_a - [klass]
+  end
+
+  def open_circuit!
+    Net::HTTP.start("localhost", 41_235) do |http|
```
Sirupsen: Similarly, can you make this a default argument?

Sirupsen:
There needs to be a test for multiple different endpoints, endpoints on the same address (but not port), closing a server and then starting it on the same endpoint, etc.

test/net_http_test.rb
```diff
@@ -0,0 +1,241 @@
...
+          Net::HTTP.get(uri)
+        end
+      end
+      assert_raises Net::ResourceBusyError do
+        threads.each(&:join)
+      end
+    end
+  end
+
+  def test_get_type_1_is_protected
```
Sirupsen: What do you mean by `type_[0..3]`?

test/net_http_test.rb
```diff
@@ -0,0 +1,241 @@
...
+
+    @proxy.downstream(:latency, latency: 500).apply do
+      uri = URI('http://localhost:41235/200')
+      (ticket_count + 1).times do
+        threads << Thread.new do
+          Net::HTTP.get(uri)
+        end
+      end
+      assert_raises Net::ResourceBusyError do
+        threads.each(&:join)
```
Sirupsen: These threads could complete in order which would never cause the bulkheads to trigger, and this test would then fail (and be flaky).

test/net_http_test.rb
```diff
@@ -0,0 +1,241 @@
...
+    end
+  end
+
+  def test_http_start_and_inner_methods_are_protected
+    open_circuit!
+
+    uri = URI('http://localhost:41235/200')
+    assert_raises Net::CircuitOpenError do
+      Net::HTTP.start(uri.host, uri.port) do |_|
+      end
```
Sirupsen: Just do an empty block with `{ .. }` here.

lib/semian/net_http.rb
kyewei: `https://github.com/ruby/ruby/blob/trunk/lib/net/http/responses.rb` shows the error codes and the exceptions they raise. `https://github.com/ruby/ruby/blob/trunk/lib/net/http/exceptions.rb` and `https://github.com/ruby/ruby/blob/trunk/lib/net/protocol.rb` lines 39-47 show all the exceptions being subclasses of `Net::ProtocolError`

test/net_http_test.rb
```diff
@@ -0,0 +1,241 @@
...
+    with_semian_options(semian_options_proc) do
+      Net::HTTP.start("localhost", 41_235) do |http|
+        assert_equal yaml_sample_config["development"][http.semian_identifier], http.raw_semian_options
+      end
+      Net::HTTP.start("localhost", 41_234) do |http|
+        assert_equal yaml_sample_config["development"]["http_default"], http.raw_semian_options
+      end
+      assert_equal yaml_sample_config["development"]["http_default"], Semian::NetHTTP.raw_semian_options
+    end
+  end
```
Sirupsen: This is pretty awesome :-)

Sirupsen:
How would we best enable this for just some HTTP endpoints to gracefully roll this out?

lib/semian/net_http.rb
Sirupsen: Agree with @byroot here.

Sirupsen:
I have the same concern as @byroot on the `BadResponse`, I don't think it's a 4xx, that'd be `BadRequest`—but it'd be nice if we could figure out what that was.

byroot:
About the IPC limits, they seems quite high:

```
------ Shared Memory Limits --------
max number of segments = 4096
max seg size (kbytes) = 18014398509465599
max total shared memory (kbytes) = 18446744073642442748
min seg size (bytes) = 1

------ Semaphore Limits --------
max number of arrays = 32000
max semaphores per array = 32000
max semaphores system wide = 1024000000
max ops per semop call = 500
semaphore max value = 32767
```

So it's probably ok, but I'm still a bit worried. We have functionalities in Shopify that basically reach any domain (images download for example).

I don't think we'd run out, but how is it garbage collected?

lib/semian/net_http.rb
kyewei: I assumed `#tr` was faster than `#gsub` since it handles only single character replacements.

lib/semian/net_http.rb
kyewei: I have a feeling you were the one that put this in when you looked at this before. So...

Sirupsen:
@byroot it wouldn't be, we haven't discussed the rollout too much but we have instrumentation in core that we need to analyze before enabling it everywhere. To start with, we'll enable it for just higher risk endpoints to roll out gradually.

byroot:
@Sirupsen perfect. We'd need to introduce a toggle mechanism then, or and hostname whitelist.

lib/semian/net_http.rb
kyewei: `self.class` doesn't work. `self` refers to `Net::HTTP`

test/net_http_test.rb
kyewei: Different ways to do a get, such as creating an instance and `#get`, using the class method `::get`, and `::get_response`

kyewei:
`HTTPBadResponse` is raised when the header doesn't match some variant `HTTP/1.1 200 OK` or the chunk size is wrong. `HTTPHeaderSyntaxError` is something do with header byte range being wrong. SO both are HTTP header errors. 

See `https://github.com/ruby/ruby/blob/trunk/lib/net/http/response.rb` or  `https://github.com/ruby/ruby/blob/trunk/lib/net/http/header.rb` for use and `https://github.com/ruby/ruby/blob/trunk/test/net/http/test_httpheader.rb` for tests.

Sirupsen:
I'm OK with circuit breaking on those TBH. @byroot @fw42 ?

byroot:
Yeah, I'm fine too. Sorry for mixing up BadResponse and BadRequest.

Sirupsen:
I don't really care, the argument against it is that it might mask an actual problem with the software behind a circuit breaker, the other argument is that that actual problem may be caused due to load and if we're hitting that many exceptions it should probably avoid being requested.

I'd say leave it in and add `HTTPHeaderSyntaxError` too.

kyewei:
Alright. Can we also have a discussion on what exceptions should be excluded? @byroot mentioned that the 4xx range shouldn't be tracked. What else? Since that requires removing the `HTTPProtocolError` and breaking it up into subclasses so might as well do it now.

Sirupsen:
Great overview @kyewei, thanks. When is `HTTPError` raised? What about `HTTPFatalError` and `HTTPServerException`? IMO making the microdecisions isn't a big deal, I'd just stick with `ProtocolError, `HTTPHeaderSyntaxError`, `HTTPBadResponse`.

kyewei:
All the [1-5]xx errors are raised after the `HTTPResponse` object is created, during the processing of the response text, but before it is given to the user. I would rather stick with the `ProtocolError` as well.

lib/semian/net_http.rb
```diff
@@ -0,0 +1,90 @@
...
+end
+
+module Semian
+  module NetHTTP
+    include Semian::Adapter
+
+    ResourceBusyError = ::Net::ResourceBusyError
+    CircuitOpenError = ::Net::CircuitOpenError
+
+    DEFAULT_SEMIAN_OPTIONS = {
```
fw42: hm none of the other adapters have default options defined, do they? why this one?

lib/semian/net_http.rb
fw42: I think he means why not leave the `.`?

fw42: What exactly is `address`? is that only the host? or also the path?

fw42: if it contains the path as well then this mapping is not 1-to-1, you can easily find two addresses that will map to the same identifier here...

lib/semian/net_http.rb
```diff
@@ -0,0 +1,90 @@
...
+    def self.exceptions
+      @exceptions ||= DEFAULT_ERRORS.dup
+    end
+
+    def self.exceptions=(exceptions)
+      @exceptions = exceptions
+    end
+
+    def resource_exceptions
+      Semian::NetHTTP.exceptions ||= DEFAULT_ERRORS.dup
```
fw42: I don't think the `||=` is necessary here since you already have it in `self.exceptions`

fw42:
I think this PR needs a way to do a slow or partial rollout. I would imagine that many users (including Shopify) would want to use this only for certain hosts (at least at the beginning). There should be a way to define which hosts that is and which ones are left "unprotected".

test/net_http_test.rb
```diff
@@ -0,0 +1,318 @@
...
+      ticket_count = Net::HTTP.new("localhost", TOXIC_PORT).raw_semian_options[:tickets]
+
+      @proxy.downstream(:latency, latency: 500).apply do
+        uri = URI('http://localhost:31051/200')
+        (ticket_count + 1).times do
+          threads << Thread.new do
+            Net::HTTP.get(uri)
+          end
+        end
+        assert_raises Net::ResourceBusyError do
```
fw42: Hmmm, you want to assert that exactly one of the threads raises, right? I think your test here would pass even if more than one would raise. And doesn't the exception potentially prevent the other threads from being joined?

How about
```ruby
raised = 0

threads.each do |thread|
  begin
    thread.join
  rescue Net::ResourceBusyError => e
    raised += 1
  end
end

assert_equal 1, raised
```

test/net_http_test.rb
fw42: I think I would make that more explicit.. `test_get_is_protected`, `test_instance_get_is_protected`, `test_get_response_is_protected`, or something like that

lib/semian/net_http.rb
Sirupsen: That was a prototype, it doesn't answer my question :-)

test/net_http_test.rb
Sirupsen: Similarly you could have a situation where none of them raise.. that's up to the scheduler. This test is fundamentally racy

test/net_http_test.rb
```diff
@@ -0,0 +1,318 @@
...
+            http.request(request)
+          end
+        end
+      end
+    end
+  end
+
+  def test_custom_raw_semian_options_work
+    with_server do
+      yaml_sample_config = {}
```
fw42: why is this called yaml? I don't see any yaml anywhere

test/net_http_test.rb
```diff
@@ -0,0 +1,318 @@
...
+      yaml_sample_config["development"] = {}
+      yaml_sample_config["development"]["http_default"] =
+        {"tickets" => 1,
+         "success_threshold" => 1,
+         "error_threshold" => 3,
+         "error_timeout" => 10}
+      yaml_sample_config["development"]["http_localhost_31051"] =
```
fw42: use PORT here instead of hardcoding?

lib/semian/net_http.rb
kyewei: Whereas the other adapters you can do a `Mysql2::Client.new(..., semian_options: { ... } )` you can't do it here since you may not be actually making the `Net::Http` instance

lib/semian/net_http.rb
kyewei: Address is just the host, so like `google.com` or `172.0.0.1`

test/net_http_test.rb
kyewei: Any suggestions?

Sirupsen: Set the tickets to one and have another thread that acquire the semaphore (without net/http for simplicity) and has a conditional variable to only give up the semaphore once you've made an assertion that the net/http request failed. Look at https://github.com/Shopify/semian/blob/master/test/resource_test.rb for some inspiration.

lib/semian/net_http.rb
Sirupsen: Even then, the other adapters don't have default parameters. To follow that, we should probably just disable this adapter until they're set maybe?

kyewei:
I refactored it to make partial rollout possible, plus a few other changes that were suggested. I removed the default semian options, and now a value of `nil`, such as an assigned key in a Hash lookup, for `#raw_semian_options` disables the adapter.

Essentially the adapter defaults to whitelisting, but it is possible to change the `Semian::NetHTTP.raw_semian_options` `proc` to blacklist instead. See https://github.com/Shopify/semian/pull/58#issue-113919464 for uses.

test/net_http_test.rb
```diff
@@ -0,0 +1,389 @@
...
+    http = Net::HTTP.new("localhost", toxic_port)
+    Timecop.travel(http.raw_semian_options[:error_timeout])
+    # Cause successes success_threshold times so circuit closes
+    http.raw_semian_options[:success_threshold].times do
+      response = http.get("/200")
+      assert(200, response.code)
+    end
+  end
+
+  def with_server(port: PORT, reset_semian_state: true)
```
Sirupsen: How about making `with_server` check the `Host` header of the HTTP request to make sure you aren't receiving connections for hosts you don't want to? That also means you can test we trigger circuits etc. in that case.

test/net_http_test.rb
```diff
@@ -0,0 +1,389 @@
...
+            Net::HTTP.get(URI("http://localhost:#{port + 1}/200"))
+          end
+        end
+      end
+    end
+  end
+
+  private
+
+  def with_semian_options(options)
```
Sirupsen: Provide a default argument here, so you don't have to pass it every time.

```ruby
def with_semian_options(options = DEFAULT_SEMIAN_OPTIONS)
 # implementation
end
```

test/net_http_test.rb
```diff
@@ -0,0 +1,389 @@
...
+        Semian["http_localhost_#{port + 1}"].reset if Semian["http_localhost_#{port + 1}"]
+        Semian.destroy('http_localhost_#{port}')
+        Semian.destroy('http_localhost_#{port + 1}')
+        with_server(port: port, reset_semian_state: false) do
+          with_toxic(upstream_port: port, toxic_port: port + 1) do |name|
+            open_circuit!(toxic_port: port + 1, toxic_name: name)
+            close_circuit!(toxic_port: port + 1)
+          end
+        end
+      end
```
Sirupsen: This is also a confusing test, I think it could be cleaned up tremendously if we go with the `with_server` model.

test/net_http_test.rb
```diff
@@ -0,0 +1,389 @@
...
+        http = Net::HTTP.new("localhost", "#{TOXIC_PORT}")
+        assert_equal true, http.enabled?
+
+        http = Net::HTTP.new("localhost", "#{TOXIC_PORT + 100}")
+        assert_equal false, http.enabled?
+      end
+    end
+  end
+
+  def test_adding_custom_errors_work
```
Sirupsen: You're testing here that adding custom errors append it to the list. You should have a test that actually asserts that when it's raised it's caught and Semian counts it towards the circuit breaker.

test/net_http_test.rb
```diff
@@ -0,0 +1,389 @@
...
+        Net::HTTP.start("localhost", PORT) do |http|
+          assert_equal semian_config["development"]["http_default"], http.raw_semian_options
+        end
+        assert_equal semian_config["development"]["http_default"],
+                     Semian::NetHTTP.retrieve_semian_options_by_identifier("http_default")
+      end
+    end
+  end
+
+  def test_custom_raw_semian_options_can_disable
```
Sirupsen: This should be two tests instead of one.

test/net_http_test.rb
```diff
@@ -0,0 +1,389 @@
...
+              request = action.new uri
+              http.request(request)
+            end
+          end
+        end
+      end
+    end
+  end
+
+  def test_custom_raw_semian_options_work
```
Sirupsen: This should be two tests. One that test the fallback to the default, one that tests

test/net_http_test.rb
```diff
@@ -0,0 +1,389 @@
...
+        open_circuit!
+        assert_raises Net::CircuitOpenError do
+          uri = URI("http://localhost:#{TOXIC_PORT}/200")
+          Net::HTTP.post_form(uri, 'q' => 'ruby', 'max' => '50')
+        end
+      end
+    end
+  end
+
+  def test_http_start_and_inner_methods_are_protected
```
Sirupsen: An `and` is typically a smell saying you should have two tests. I think this is an example of this.

test/net_http_test.rb
```diff
@@ -0,0 +1,389 @@
...
+        open_circuit!
+        assert_raises Net::CircuitOpenError do
+          uri = URI("http://localhost:#{TOXIC_PORT}/200")
+          Net::HTTP.get_response(uri)
+        end
+      end
+    end
+  end
+
+  def test_post_type_1_is_protected
```
Sirupsen: `post_form` instead of `post_type_1` would be a better description of this test.

test/net_http_test.rb
```diff
@@ -0,0 +1,389 @@
...
+    with_semian_options(DEFAULT_SEMIAN_OPTIONS) do
+      with_server do
+        open_circuit!
+        assert_raises Net::CircuitOpenError do
+          uri = URI("http://localhost:#{TOXIC_PORT}/200")
+          Net::HTTP.get_response(uri)
+        end
+      end
+    end
+  end
```
Sirupsen: Good job on covering all of these cases.

test/net_http_test.rb
```diff
@@ -0,0 +1,389 @@
...
+          sleep(1) until acquired_count == ticket_count
+          http = Net::HTTP.new("localhost", "#{TOXIC_PORT}")
+          assert_raises Net::ResourceBusyError do
+            http.get("/")
+          end
+          threads.each(&:join)
+        end
+      end
+    end
+  end
```
Sirupsen: This test is incredibly difficult to read. It needs to be easier to understand. I'm not even going to try :)

test/net_http_test.rb
```diff
@@ -0,0 +1,389 @@
...
+      Net::HTTP.start("localhost", TOXIC_PORT) do |http|
+        assert_equal "http_localhost_#{TOXIC_PORT}", http.semian_identifier
+      end
+    end
+  end
+
+  def test_trigger_open
+    with_semian_options(DEFAULT_SEMIAN_OPTIONS) do
+      with_server do
+        open_circuit!
```
Sirupsen: I think this is a good example of a test helper that shouldn't have a default argument, because it makes the tests more confusing to read. I'd define the URL you're testing at the start of the test and then pass it in here.

test/net_http_test.rb
```diff
@@ -0,0 +1,389 @@
...
+    tickets: 3,
+    success_threshold: 1,
+    error_threshold: 3,
+    error_timeout: 10,
+  }
+
+  def test_semian_identifier
+    with_server do
+      Net::HTTP.start("localhost", TOXIC_PORT) do |http|
+        assert_equal "http_localhost_#{TOXIC_PORT}", http.semian_identifier
```
Sirupsen: In the spirit of my comment on your other PR, if I removed the `#tr` in the `#semian_identifier` implementation no test would fail, since all of them use `localhost`. What about `google.com`? `sïmånhøråpéskîldßæñ.com`?

test/net_http_test.rb
```diff
@@ -0,0 +1,389 @@
...
+
+  DEFAULT_SEMIAN_OPTIONS = {
+    tickets: 3,
+    success_threshold: 1,
+    error_threshold: 3,
+    error_timeout: 10,
+  }
+
+  def test_semian_identifier
+    with_server do
```
Sirupsen: Pass the `port` of the server as an argument to the block. This will easily allow you to create a new test where multiple servers are alive and you query them. Or even better, pass the entire hostname + port it is set up for (see my comment on `with_server` to make it reject other connections):

```ruby
with_server do |hostname|
  open_circuit!(hostname)
  Net::HTTP.get(URI(hostname))
end
```

And so on :-)

lib/semian/net_http.rb
```diff
@@ -0,0 +1,82 @@
...
+    def connect
+      return super unless enabled?
+      acquire_semian_resource(adapter: :nethttp, scope: :connection) do
+        super
+      end
+    end
+
+    def request(*req)
+      return super unless enabled?
+      acquire_semian_resource(adapter: :nethttp, scope: :query) do
```
Sirupsen: I would just use `http` instead of `nethttp` which is just whatever Ruby called the library.

lib/semian/net_http.rb
```diff
@@ -0,0 +1,82 @@
...
+      @semian_identifier = semian_identifier
+    end
+  end
+
+  ResourceBusyError = Class.new(SemianError)
+  CircuitOpenError = Class.new(SemianError)
+end
+
+module Semian
+  module NetHTTP
```
Sirupsen: This looks very clean now. Good job.

test/net_http_test.rb
kyewei: I remember putting in a test for `127.0.0.1` though. Maybe it's not in that method hmm.

test/net_http_test.rb
kyewei: Where is it difficult to read? All it does is acquire and sleep `ticket_count` threads and then assert the next one acquired throws a Semian error.

test/net_http_test.rb
kyewei: Can you clarify what you mean by this?

test/net_http_test.rb
kyewei: Yeah I'm not completely sure what you mean by the hostname part. I did add the ability to pass in multiple ports as an array to `with_server`. https://github.com/Shopify/semian/pull/58#discussion_r43900596
Do you mean that you can technically give it different hostnames to make a server on? How would you handle if they give some hostname like `google.com` which you can't bind a server to? Or can you?

test/net_http_test.rb
Sirupsen: Your explanation is easy to read. You need to write some helpers or similar to make this code easier to parse. Going from this code to the concise explanation you just made is difficult.

test/net_http_test.rb
Sirupsen: You need to handle the rejections of invalid hostnames at the HTTP level. I'm sure Thin gives you access to this.

kyewei:
I incorporated the suggested changes.

kyewei:
Can someone take another look at this?

fw42:
can you rebase? this has merge conflicts

kyewei:
It is rebased. Hopefully it builds fine, although those 404 toxiproxy errors might come up by the time it finishes...

sunblaze:
Tried adding some exceptions like your example in an initializer
```
# If we need extra errors, concat more onto array
# assert_equal(Semian::NetHTTP.exceptions, Semian::NetHTTP::DEFAULT_ERRORS)
Semian::NetHTTP.exceptions.concat([::Resolv::ResolvError, ::OpenURI::HTTPError, ::OpenSSL::SSL::SSLError])
```

but resulted it a ```undefined method `concat' for nil:NilClass```

Looks to not be initialized at this stage.

kyewei:
@sunblaze It was the default parameters being initialized in a different place. Confirm it works now?

sunblaze:
Yep that fixed it :+1:

Sirupsen:
This is ready IMO. Let's have @sunblaze get his usage to production before merging, and then we'll merge it and start testing in Shopify before cutting a release.

lib/semian/net_http.rb
```diff
@@ -0,0 +1,84 @@
...
+      ::EOFError,
+      ::IOError,
+      ::SystemCallError, # includes ::Errno::EINVAL, ::Errno::ECONNRESET, ::Errno::ECONNREFUSED, ::Errno::ETIMEDOUT, and more
+    ].freeze # Net::HTTP can throw many different errors, this tries to capture most of them
+
+    Semian::NetHTTP.exceptions = Semian::NetHTTP::DEFAULT_ERRORS.dup
+
+    def resource_exceptions
+      Semian::NetHTTP.exceptions
+    end
```
Sirupsen: I think we should add a public helper method here to add custom exceptions. I know it's currently possible, but I'm concerned it won't be obvious to users this is something to think about when they read the code. On the other hand, most people will likely not even read the code and will probably just cargo-cult examples.

Sirupsen:
If @sunblaze's testing of this is successful and adding new exceptions becomes more intuitive, I think this is ready to merge along with a section in the README on how to use this adapter as it's more difficult to configure than others.

kyewei:
Added the README stuff. 

I just realized that we never actually distinguish between `http` and `https` connections, and the `semian_identifier` always being hard-coded to `"http_#{address.tr('.', '_')}_#{port}"` may be a bit misleading. Should `https` be added?

README.md
```diff
@@ -126,6 +127,49 @@ client = Redis.new(semian: {
...
+and returns a `Hash` with keys as follows, perhaps in an initialization file.
+
+```ruby
+SEMIAN_PARAMETERS = { tickets: 1,
+                      success_threshold: 1,
+                      error_threshold: 3,
+                      error_timeout: 10 }
+Semian::NetHTTP.raw_semian_options = proc do |semian_identifier|
+  # Let's make it only active for github.com
+  if semian_identifier == "http_github_com_80"
```
byroot: Think it was already discussed, but the `tr('.', '_')` makes this really weird. Can't we just not do the `tr`?

byroot:
LGTM beside the underscore thing.

README.md
kyewei: Would it seem a bit odd since all the other ones are like `mysql2_shard0` or something? What format are you suggesting? Also is `nethttp` a better prefix?

byroot: The other ones are arbitrarily named from their config. We use underscores, but anyone can use anything.

I think this adapter is fundamentally different from the others, and that you want to configure Semian on a host basis. So IMO the config lookup should happen with the domain, not the "identifier". 

I wouldn't even include the port TBH, I can't think of an use case were you'd want different configuration on different ports.

Also I can envision use cases where you want multiple domain to use the same CB, e.g. `customer1.provider.com` and `customer2.provider.com`.

So maybe the collabck should pass the URI, and get the identifier in return. (not 100% sure here, it would require more thoughts, @Sirupsen @fw42 what do you say?)

But that last thing can be addressed in a followup.

kyewei: The "not including port" thing will break a few tests for sure.

Also, going on that callback for the identifier thing, the adapter would need two callbacks, one for the uri -> identifier, and one for identifier -> parameters? We can't just merge them since the identifier is definitely needed for the name. That's quite a bit of user-involved customizing going on, best to have some default as well then.

byroot: > one for the uri -> identifier, and one for identifier -> parameters?

That, or the actual ` identifier -> parameters?` is just a static hash.

This also allow Semian to offer a default `uri -> identifier` callback, so that for simple case you just have to set a mapping of `domain -> parameters`.

kyewei: I think I am correct in saying that we want this in before code freeze? @Sirupsen you probably want to read this before then.

Sirupsen: @kyewei the reason to get this in master is for @sunblaze's PR, we're not getting it in Shopify before BFCM.

As for @byroot's point, I think it's good—you may want to make decisions about grouping a larger set of domains in the same Semian resource, and you cannot do that with the current design. Invoking a closure on every single resource call that returns the resource name and the config, and make that mandatory, seems like the way to go.

sunblaze: I agree with the `nethttp_#{addr}_#{port}` format without any of the `tr('.', '_')` but it would be even better if the host and port was just passed separately to the block, then I wouldn't have to do any string manipulation.

kyewei:
Refactored it to have one closure that takes `host`, `port` and returns `semian_identifier` and `semian_options`. Got rid of the `tr`, made it default to "nethttp_#{host}_#{port}". Added a test for letting multiple hosts combine into one semian resource.

README.md
```diff
@@ -126,6 +127,54 @@ client = Redis.new(semian: {
...
+that takes two parameters, `host` and `port like `127.0.0.1` and `80` or `github_com` and `80`,
+and returns two values, a `semian_identifier` and a `Hash` with the keys as follows.
+
+```ruby
+SEMIAN_OPTIONS = { tickets: 1,
+                   success_threshold: 1,
+                   error_threshold: 3,
+                   error_timeout: 10 }
+Semian::NetHTTP.semian_configuration = proc do |host, port|
+  semian_identifier = "nethttp_#{host}_#{port}"
```
sunblaze: Would be nice if I had the option of not having to build the semian_identifier. But I gather you might need to if you want multiple host and port combinations to use the same semian resource.

README.md
```diff
@@ -126,6 +127,54 @@ client = Redis.new(semian: {
...
+  # Let's make it only active for github.com
+  if semian_identifier == "nethttp_github.com_80"
+    [semian_identifier, SEMIAN_OPTIONS]
```
sunblaze: Could we instead return a hash like the following, and have the identifier be optional (use default).
```
{identifier: semian_identifier, options: SEMIAN_OPTIONS}
```
It makes it a bit clearer instead of using an array.

lib/semian/net_http.rb
```diff
@@ -0,0 +1,94 @@
...
+module Semian
+  module NetHTTP
+    include Semian::Adapter
+
+    ResourceBusyError = ::Net::ResourceBusyError
+    CircuitOpenError = ::Net::CircuitOpenError
+
+    def semian_identifier
+      @configuration ||= Semian::NetHTTP.retrieve_semian_configuration_by_host_port(address, port)
+      @configuration[0]
```
sunblaze: The redis and mysql adapters return symbols for identifiers should we do the same to be consistent?

README.md
Sirupsen: I don't understand why identifier changed from `http_` to `nethttp_`, that leaks an abstraction of an internal Ruby library in the name, which doesn't make sense to me. I'm completely OK with not having an option to switch, but simply documenting this as a simple case instead of adding it as a default fallback, so :+1: to this except the default naming.

README.md
```diff
@@ -126,6 +127,54 @@ client = Redis.new(semian: {
 })
 ```
 
+#### Net::HTTP
+For the `Net::HTTP` specific Semian adapter, since many external libraries may create
+HTTP connections on the user's behalf, the parameters are instead provided
+by calling specific functions in `Semian::NetHTTP`, perhaps in an initialization file.
+
+To give Semian parameters, assign a `proc` to `Semian::NetHTTP.semian_configuration`
+that takes two parameters, `host` and `port like `127.0.0.1` and `80` or `github_com` and `80`,
```
Sirupsen: You have a markdown error here, you need to close the backticks after `port`.

README.md
```diff
@@ -126,6 +127,54 @@ client = Redis.new(semian: {
...
+and returns two values, a `semian_identifier` and a `Hash` with the keys as follows.
```
Sirupsen: You're not making it clear from your example that you're sharing the resource, and what that really means. I think this documentation would be fairly hard to understand for someone not intimate with Semian and the implementation.

lib/semian/net_http.rb
```diff
@@ -0,0 +1,94 @@
...
+
+    class << self
+      attr_accessor :semian_configuration
+      attr_accessor :exceptions
+
+      def retrieve_semian_configuration_by_host_port(host, port)
+        if @semian_configuration.respond_to?(:call)
+          @semian_configuration.call(host, port) || ["nethttp_#{host}_#{port}", nil]
+        else
+          ["nethttp_#{host}_#{port}", nil] # adapter is disabled by default
```
Sirupsen: IMO let's kill a default option, as this is fairly dangerous. Better have people write a closure to make sure they understand what they're doing.

lib/semian/net_http.rb
```diff
@@ -0,0 +1,94 @@
...
+      ::SystemCallError, # includes ::Errno::EINVAL, ::Errno::ECONNRESET, ::Errno::ECONNREFUSED, ::Errno::ETIMEDOUT, and more
+    ].freeze # Net::HTTP can throw many different errors, this tries to capture most of them
+
+    class << self
+      attr_accessor :semian_configuration
+      attr_accessor :exceptions
+
+      def retrieve_semian_configuration_by_host_port(host, port)
+        if @semian_configuration.respond_to?(:call)
+          @semian_configuration.call(host, port) || ["nethttp_#{host}_#{port}", nil]
```
Sirupsen: ditto

lib/semian/net_http.rb
```diff
@@ -0,0 +1,94 @@
...
+
+module Semian
+  module NetHTTP
+    include Semian::Adapter
+
+    ResourceBusyError = ::Net::ResourceBusyError
+    CircuitOpenError = ::Net::CircuitOpenError
+
+    def semian_identifier
+      @configuration ||= Semian::NetHTTP.retrieve_semian_configuration_by_host_port(address, port)
```
Sirupsen: Why is this getting cached? Doesn't that mean it'd be the same thing on the next, but different host and port? Why does no test fail if I'm right?

README.md
Sirupsen: The identifier could be merged into the `SEMIAN_OPTIONS`, but I'm not sure whether that's more clear. I think a hash would make it more readable, sure.

README.md
sunblaze: My understanding of why the switch to nethttp was to help differentiate from http and https. Since it shouldn't matter since the port is included.

README.md
```diff
@@ -126,6 +127,54 @@ client = Redis.new(semian: {
...
+    [semian_identifier, nil]
+  end
+end
+
+# Called from within API:
+# semian_identifier, semian_options = Semian::NetHTTP.semian_configuration("github.com", 80)
+```
+
+The `semian_identifier` used as the name of the resource, and the `semian_options` are the
+configuration parameters. A return value of `nil` means Semian is disabled for that
```
Sirupsen: For ease of understanding, I'd explain what you mean by "resource" here. As in, something like: the `semian_identifier` specifies the name of the resource for which the `semian_options` apply. In practise this means that all endpoints that resolve to the same `semian_identifier` share their circuit breaker and bulkhead configuration. It could for sure be even more clear than that, play around with it a little.

README.md
```diff
@@ -126,6 +127,54 @@ client = Redis.new(semian: {
...
+`Net::HTTP` endpoint. This works well since the result of a failed Hash lookup is `nil` also.
+As such, this particular adapter defaults to whitelisting, although the
+behavior can be changed to blacklisting or even be completely disabled by varying
+the use of returning `nil`.
+
+Since we envision this particular adapter can be used in combination with many
+external libraries, we added functionality to expand the Exceptions that can be
+tracked as part of Semian's circuit breaker. Add exceptions and reset to the
+[`default`][nethttp-default-errors] list using the following:
```
Sirupsen: You should give an example of why this is useful, e.g. a library might re-raise Net::HTTP exceptions already and then you need this.

lib/semian/net_http.rb
```diff
@@ -0,0 +1,94 @@
...
+        else
+          ["nethttp_#{host}_#{port}", nil] # adapter is disabled by default
+        end
+      end
+
+      def reset_exceptions
+        self.exceptions = Semian::NetHTTP::DEFAULT_ERRORS.dup
+      end
+
+      def concat_exceptions(exceptions)
```
Sirupsen: I'm not in love with this name since it's fairly low-level, but failing to come up with a good alternative.

lib/semian/net_http.rb
```diff
@@ -0,0 +1,94 @@
...
+        self.exceptions.concat(exceptions)
+      end
+    end
+
+    Semian::NetHTTP.reset_exceptions
+
+    def raw_semian_options
+      @configuration ||= Semian::NetHTTP.retrieve_semian_configuration_by_host_port(address, port)
```
Sirupsen: ditto, I don't think you can ever cache this value and it concerns me no test fails when you have two different addresses and ports. That test needs to be added, along with one that changes the closure at runtime.

README.md
kyewei: ^ I changed to `nethttp` since if doesn't make more sense to have `http_` if `https_` is not there also, and that introduces two possible `semian_identifier` prefixes.

If we provide the ability to customize the `semian_identifier`, then it's easier to do work with one `proc` that outputs both the identifier and the configuration options than to have two `proc`'s that probably repeat the same `switch`-`case` logic. I guess returning a `{identifier: semian_identifier, options: SEMIAN_OPTIONS}` would be better since you *can* make the user not specify the identifier. If no one objects I'll change it to that

README.md
kyewei: Good catch

lib/semian/net_http.rb
kyewei: ^It's instance level. So instance of `Net::HTTP`. So it's fine.

README.md
Sirupsen: I think this problem disappears if we don't allow defaults.

lib/semian/net_http.rb
kyewei: See my comment on the first post. `@configuration` is an instance variable on an `Net::HTTP` instance since it is within an instance method. If this were in a method with `def self.method_name` or just not inside a method, then it wouldn't be OK.

lib/semian/net_http.rb
kyewei: Kill as in disable? It disables if the `raw_semian_options` is `nil`. Unless you mean literally the adapter doesn't work and raises errors unless the closure is written?

README.md
sunblaze: How about we use a `name` option like this https://github.com/Shopify/semian/blob/master/lib/semian/redis.rb#L47-L53

Then the `semian_identifier` method can prefix whatever it wants and the library users aren't exposed to the prefix.

lib/semian/net_http.rb
sunblaze: Can we DRY this out and move this memoization logic into a configuration reader method instead?

lib/semian/net_http.rb
sunblaze: Could we drop this method and let the API users use << or += on Semian::NetHTTP.exceptions instead?

kyewei: https://github.com/Shopify/semian/pull/58#discussion_r44857311

lib/semian/net_http.rb
kyewei: Well... it's like saying put the `return super unless enabled?` line in the `connect` and `request` in it's own method.

It's going to end up being something like
```ruby
def access_options
  @configuration ||= Semian::NetHTTP.retrieve_semian_configuration_by_host_port(address, port)
  yield @configuration
end

def raw_semian_options
  access_options { |configuration| ... # stuff }
end
```
Meh.

sunblaze: That was a little more complicated then what I was expecting. This is more what I was thinking of.

```ruby
def configuration
  @configuration ||= Semian::NetHTTP.retrieve_semian_configuration_by_host_port(address, port)
end

def semian_identifier
  configuration[0]
end

def raw_semian_options
  configuration[1]
end
```

Sirupsen: You still have the problem of the closure returning different identifiers and options over time. What happens? Would it resize properly? Would the old resource get destroyed? Should an exception get raised?

kyewei: Meh, I'll see. If I do the `:name` thing, then I'll want to cache the `Hash` with the `:name` key-value pair removed. So it might not work like ^ at all.

kyewei: @Sirupsen, I can't imagine a situation where a library will use a `Net::HTTP` connection through an instance of `Net::HTTP` object, and then somehow have hot-swapped closures, and still use the same `Net::HTTP` instance. If you think it's a valid concern, then I'll just remove the memoization

Sirupsen: Imagine if someone did something like this (which reminds me you should be able to set it like this, which should be possible by just setting it when `block_given?`, then you can remove the `#=` variant IMO):

```ruby
Semian::NetHTTP.raw_semian_options do |host, port|
  ["#{host}_#{port}_#{rand(2)}", some_default_options]
end
```

It's not really a legit use-case, but there could be other examples where the `semian_identifier` returned differ between multiple invocations. Currently we just don't support it at all, which is likely fine, I just want to make sure that's done on purpose. When I talk about switching closures, I mean overriding `raw_semian_options` at runtime, which might be later in the code where the above is:

```ruby
Semian::NetHTTP.raw_semian_options do |host, port|
  ["#{host}_#{port}_omghardcoded", some_default_options]
end
```

Now it wouldn't actually change the Semian identifier, since it's memoized on that object.

lib/semian/net_http.rb
Sirupsen: yeah, I think when doing the comment @kyewei links to I was coloured by the examples in the code that did some more complicated stuff, but `+=` sounds fine to me, and since we have the README entry it's simple to document. I'm fine with reverting this.

kyewei:
See commit msg for changes^ 

One change I didn't make is the `retrieve_semian_configuration` receiving a block and `block_given?` checking to store the closure. Is that really necessary?

README.md
```diff
@@ -126,6 +127,65 @@ client = Redis.new(semian: {
...
+```ruby
+SEMIAN_PARAMETERS = { tickets: 1,
+                      success_threshold: 1,
+                      error_threshold: 3,
+                      error_timeout: 10 }
+Semian::NetHTTP.semian_configuration = proc do |host, port|
+  # Let's make it only active for github.com
+  if host == "github.com" && port == "80"
+    SEMIAN_PARAMETERS.dup.tap do |options|
+      options[:identifier] = nethttp_github.com_80
```
sunblaze: nethttp_github.com_80 isn't quoted

README.md
```diff
@@ -126,6 +127,65 @@ client = Redis.new(semian: {
...
+
+```ruby
+SEMIAN_PARAMETERS = { tickets: 1,
+                      success_threshold: 1,
+                      error_threshold: 3,
+                      error_timeout: 10 }
+Semian::NetHTTP.semian_configuration = proc do |host, port|
+  # Let's make it only active for github.com
+  if host == "github.com" && port == "80"
+    SEMIAN_PARAMETERS.dup.tap do |options|
```
sunblaze: I'd use `SEMIAN_PARAMETERS.merge(identifier: "nethttp_github.com_80")` instead of the dup and tap

lib/semian/net_http.rb
```diff
@@ -0,0 +1,89 @@
...
+
+module Semian
+  module NetHTTP
+    include Semian::Adapter
+
+    ResourceBusyError = ::Net::ResourceBusyError
+    CircuitOpenError = ::Net::CircuitOpenError
+
+    def semian_identifier
+      Semian::NetHTTP.retrieve_semian_configuration(address, port)[:identifier]
```
sunblaze: I still feel like we should use `:name`, add a prefix and make it a symbol here. I don't think we should let the user set `semian_identifier` this directly here. Both the mysql and redis adapters already do this and also have a default.

kyewei: One of the issues was that if this is used with a lot of hostnames, then there are a lot of symbols lying around that can't be garbage collected.

sunblaze: That is true, but I still think adding a prefix and having a default using the address and port is a good idea.

lib/semian/net_http.rb
```diff
@@ -0,0 +1,90 @@
...
+      ::IOError,
+      ::SystemCallError, # includes ::Errno::EINVAL, ::Errno::ECONNRESET, ::Errno::ECONNREFUSED, ::Errno::ETIMEDOUT, and more
+    ].freeze # Net::HTTP can throw many different errors, this tries to capture most of them
+
+    class << self
+      attr_accessor :semian_configuration
+      attr_accessor :exceptions
+
+      def retrieve_semian_configuration(host, port)
+        @semian_configuration.call(host, port) if @semian_configuration.respond_to?(:call)
```
sunblaze: Can we add a check here to ensure they pass :name in the hash if they don't pass nil from the proc's call?

kyewei: Why does that even matter though? The instructions show the `:name` being included already.

sunblaze: Just makes it easier to find a bug or if they haven't read the instructions.

kyewei: They should probably read the instructions before using the adapter. That's the whole point of making things disable-until-explicitly-enabled - so that people who use it have to know what they're doing. 
Plus, I don't think it'll be an issue, since in order to figure out how to use this adapter they would have seen the example in the README, which shows (1) merging the `:name`, and (2) using `[:name]` it to form the `semian_identifier`.

methodmissing:
Added a small hook specific change in https://github.com/Shopify/semian/commit/dc69465c821921948953a51f91a8e1b10ccccc27

The original prepend API doesn't yield itself well to other libs also with fingers in the Net::HTTP pie such as VCR, Fakeweb etc. for remote tests.

Example from a test suite (the segmentation fault off course being out of bounds for this PR, happens in Ruby):

```
# Running:

.............................................................................................................................................................................................................................................................................E.........EEEEESEEESESegmentation fault
```

Backtrace when that happens:

```
(gdb) bt
#0  0x00007fd592f2b129 in rb_vm_get_ruby_level_next_cfp (cfp=0x7fd59592b810, th=0x7fd5958d6000) at vm.c:250
#1  rb_sourcefile () at vm.c:990
#2  0x00007fd592dc46f2 in rb_bug (fmt=fmt@entry=0x7fd592fc4ccc "Segmentation fault at %p") at error.c:334
#3  0x00007fd592eb251e in sigsegv (sig=<optimized out>, info=0x7fd5958ddbf0, ctx=0x7fd5958ddac0) at signal.c:828
#4  <signal handler called>
#5  0x00007fd592dca3de in errinfo_place (th=0x7fd5958d6000) at eval.c:1451
#6  get_thread_errinfo (th=0x7fd5958d6000) at eval.c:1469
#7  setup_exception (th=th@entry=0x7fd5958d6000, tag=tag@entry=6, mesg=<optimized out>) at eval.c:479
#8  0x00007fd592dcaa78 in rb_longjmp (tag=tag@entry=6, mesg=140555315305560) at eval.c:559
#9  0x00007fd592dcd5e4 in rb_exc_raise (mesg=140555315305560) at eval.c:572
#10 0x00007fd592d6c66b in vm_stackoverflow () at vm_insnhelper.c:30
#11 0x00007fd592f2f6ad in vm_push_frame (local_size=0, stack_max=<optimized out>, me=0x7fd59a891680, sp=<optimized out>, pc=<optimized out>, specval=<optimized out>, klass=<optimized out>, 
    self=<optimized out>, type=17, iseq=<optimized out>, th=0x7fd5958d6000) at vm_insnhelper.c:50
#12 vm_call_iseq_setup_normal (ci=<optimized out>, cfp=0x7fd59592b810, th=0x7fd5958d6000) at vm_insnhelper.c:1283
```

Happy to help iron out any remaining issues and give this adapter runtime. Nice work!

@Sirupsen @fw42 @kyewei

kyewei:
@methodmissing That should be fine. That's definitely a compatibility issue between `:include` and `:prepend`. I remember seeing it somewhere else as well in a similar situation.

methodmissing:
Yes - and wasn't expecting the SIGSEGV in Ruby either.

Sirupsen:
I'm OK with cutting a release for this since @methodmissing and @sunblaze need it, but I'm not sure the test code here is done. It still seems extremely complex, and creates Toxiproxy proxies at run-time instead of at boot-time.

methodmissing:
I can allocate some time to cleanups during the Dec 1 through 18 window - never worked directly with Toxiproxy and have an interest to learn more about it. I think it's better to cut now in favor of more runtime exposure and contexts and handle cleaner tests async ...

Sirupsen:
@methodmissing that'd be amazing.

PR #60, opened 2015-11-04
allow rollbacks

fw42:
Semian should not prevent MySQL rollbacks, otherwise queries that are made within a transaction can't get properly rolled back by things like ActiveRecord, because the rollback query also fails, which might keep transactions open for an unnecessary amount of time, causing database contention.

@tjoyal @camilo @Sirupsen @sroysen @byroot

### DIFF
--- a/lib/semian/mysql2.rb
+++ b/lib/semian/mysql2.rb
@@ -32,7 +32,11 @@ module Semian
     DEFAULT_HOST = 'localhost'
     DEFAULT_PORT = 3306
 
-    QUERY_WHITELIST = Regexp.union(/\A\s*ROLLBACK/i, /\A\s*RELEASE\s+SAVEPOINT/i)
+    QUERY_WHITELIST = Regexp.union(
+      /\A\s*ROLLBACK/i,
+      /\A\s*COMMIT/i,
+      /\A\s*RELEASE\s+SAVEPOINT/i,
+    )
 
     # The naked methods are exposed as `raw_query` and `raw_connect` for instrumentation purpose
     def self.included(base)
--- a/lib/semian/mysql2.rb
+++ b/lib/semian/mysql2.rb
@@ -32,6 +32,8 @@ module Semian
     DEFAULT_HOST = 'localhost'
     DEFAULT_PORT = 3306
 
+    QUERY_WHITELIST = [/\A\s*ROLLBACK/i, /\A\s*RELEASE\s+SAVEPOINT/i].freeze
+
     # The naked methods are exposed as `raw_query` and `raw_connect` for instrumentation purpose
     def self.included(base)
       base.send(:alias_method, :raw_query, :query)
@@ -53,10 +55,18 @@ module Semian
     end
 
     def query(*args)
-      acquire_semian_resource(adapter: :mysql, scope: :query) { raw_query(*args) }
+      if query_whitelisted?(*args)
+        raw_query(*args)
+      else
+        acquire_semian_resource(adapter: :mysql, scope: :query) { raw_query(*args) }
+      end
     end
 
     private
+
+    def query_whitelisted?(sql, *)
+      QUERY_WHITELIST.any? { |pattern| sql =~ pattern }
+    end
 
     def connect(*args)
       acquire_semian_resource(adapter: :mysql, scope: :connection) { raw_connect(*args) }
--- a/lib/semian/mysql2.rb
+++ b/lib/semian/mysql2.rb
@@ -70,6 +70,12 @@ module Semian
 
     def query_whitelisted?(sql, *)
       QUERY_WHITELIST =~ sql
+    rescue ArgumentError
+      # The above regexp match can fail if the input SQL string contains binary
+      # data that is not recognized as a valid encoding, in which case we just
+      # return false.
+      return false unless sql.valid_encoding?
+      raise
     end
 
     def connect(*args)
--- a/lib/semian/mysql2.rb
+++ b/lib/semian/mysql2.rb
@@ -32,7 +32,7 @@ module Semian
     DEFAULT_HOST = 'localhost'
     DEFAULT_PORT = 3306
 
-    QUERY_WHITELIST = [/\A\s*ROLLBACK/i, /\A\s*RELEASE\s+SAVEPOINT/i].freeze
+    QUERY_WHITELIST = Regexp.union(/\A\s*ROLLBACK/i, /\A\s*RELEASE\s+SAVEPOINT/i)
 
     # The naked methods are exposed as `raw_query` and `raw_connect` for instrumentation purpose
     def self.included(base)
@@ -65,7 +65,7 @@ module Semian
     private
 
     def query_whitelisted?(sql, *)
-      QUERY_WHITELIST.any? { |pattern| sql =~ pattern }
+      QUERY_WHITELIST =~ sql
     end
 
     def connect(*args)
--- a/test/mysql2_test.rb
+++ b/test/mysql2_test.rb
@@ -180,6 +180,15 @@ class TestMysql2 < MiniTest::Unit::TestCase
     end
   end
 
+  def test_query_whitelisted_returns_false_for_binary_sql
+    client = connect_to_mysql!
+
+    q = "INSERT IGNORE INTO `theme_template_bodies` (`cityhash`, `body`, `created_at`) VALUES ('716374049952273167', \
+'\xB1\x01\xD0{\\\"current\\\":{\\\"bg_color\\\":\\\"#ff0000\\\"},\\\"presets\\\":{\\\"sandbox>,\\0\\07\x05\x01\x01, \
+grey_bg\\\":6M\\0\\06\x05\x01\x01!\fblueJ!\\0\x01l\x04ff\x01!\bredJ \\0$ff0000\\\"}}}', '2015-11-06 19:08:03.498432')"
+    refute client.send(:query_whitelisted?, q)
+  end
+
   def test_semian_allows_rollback_to_safepoint
     client = connect_to_mysql!
 
--- a/test/mysql2_test.rb
+++ b/test/mysql2_test.rb
@@ -170,6 +170,16 @@ class TestMysql2 < MiniTest::Unit::TestCase
     end
   end
 
+  def test_semian_allows_commit
+    client = connect_to_mysql!
+
+    client.query('START TRANSACTION;')
+
+    Semian[:mysql_testing].acquire do
+      client.query('COMMIT;')
+    end
+  end
+
   def test_semian_allows_rollback_to_safepoint
     client = connect_to_mysql!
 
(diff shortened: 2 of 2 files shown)


lib/semian/mysql2.rb
```diff
@@ -53,11 +53,19 @@ def semian_identifier
...
+        raw_query(*args)
+      else
+        acquire_semian_resource(adapter: :mysql, scope: :query) { raw_query(*args) }
+      end
     end
 
     private
 
+    def query_whitelisted?(sql, *)
+      sql =~ /\A\s*ROLLBACK/i
```
camilo: is rubby smart enough to not instantiate a new regexp every time? if not make it a constant

byroot: Look like it is.
```
>> puts RubyVM::InstructionSequence.disasm(Proc.new{ bar =~ /plop/ })
== disasm: <RubyVM::InstructionSequence:block in irb_binding@(irb)>=====
== catch table
| catch type: redo   st: 0002 ed: 0011 sp: 0000 cont: 0002
| catch type: next   st: 0002 ed: 0011 sp: 0000 cont: 0011
|------------------------------------------------------------------------
0000 trace            256                                             (   7)
0002 trace            1
0004 putself          
0005 opt_send_simple  <callinfo!mid:bar, argc:0, FCALL|VCALL|ARGS_SKIP>
0007 putobject        /plop/
0009 opt_regexpmatch2 <callinfo!mid:=~, argc:1, ARGS_SKIP>
0011 trace            512
0013 leave  
```

tjoyal:
I would add an extra test for rollback to savepoints. Just to be sure we do not break it in the future.

lib/semian/mysql2.rb
```diff
@@ -32,6 +32,8 @@ module Mysql2
     DEFAULT_HOST = 'localhost'
     DEFAULT_PORT = 3306
 
+    QUERY_WHITELIST = [/\A\s*ROLLBACK/i].map(&:freeze).freeze
```
byroot: It doesn't harm, but Regexp are immutable already.

lib/semian/mysql2.rb
```diff
@@ -53,11 +55,19 @@ def semian_identifier
...
+        raw_query(*args)
+      else
+        acquire_semian_resource(adapter: :mysql, scope: :query) { raw_query(*args) }
+      end
     end
 
     private
 
+    def query_whitelisted?(sql, *)
+      QUERY_WHITELIST.any? { |pattern| sql =~ pattern }
```
byroot: Might as well generate a big regepx with `Regexp.union`: http://ruby-doc.org/core-2.1.1/Regexp.html#method-c-union

byroot:
Seems sensible. Should we do the same with savepoints?

fw42:
added a test for savepoints

byroot:
Sorry by savepoint I meant the `RELEASE SAVEPOINT` syntax. IIRC it's what ActiveRecord uses: http://dev.mysql.com/doc/refman/5.7/en/savepoint.html

test/mysql2_test.rb
```diff
@@ -160,6 +160,26 @@ def test_resource_acquisition_for_query
...
 
+  def test_semian_allows_rollback
+    client = connect_to_mysql!
+
+    client.query('START TRANSACTION;')
+
+    Semian[:mysql_testing].acquire do
+      client.query('ROLLBACK;')
+    end
+  end
```
Sirupsen: You haven't opened the circuits and the bulkhead count is not 0 in the tests, so why would this fail if this patch wasn't here?

lib/semian/mysql2.rb
Sirupsen: This is going to be done on EVERY single query. Perhaps we should do `trim` + `starts_with?` instead of a regexp?

test/mysql2_test.rb
fw42: because of the acquire? removing my code changes does make this test fail...

fw42: check out the test right above (line 153 to 161)

Sirupsen: This Semian resource has one ticket. Nothing else has claimed it during this block as far as I can see, so I'm confused.

fw42: dunnolol, @byroot wrote that code

byroot: @Sirupsen `Mysql::Client#query` is monkey patched to call acquire too.

So this come down to:

```ruby
acquire do
  acquire do
    original_query
  end
end
```

fw42:
yay/nay?

camilo:
@fw42 I think we need to prove this actually solves the potential for dangling transactions  before shipping

test/mysql2_test.rb
Sirupsen: Ooooh, of course. :-) *stoopid me*

sroysen:
I think that I may be able to prove that semian opencircuiting  a ROLLBACK caused MySQL to keep a transaction open when it shouldn't (I had a ton of help from @tjoyal!).
 
I run a genghis earlier today with similar effects as we saw this week.

This is semian opencircuiting (sorry if I am not using the proper terminology) a ROLLBACK.
https://logs.shopify.com/en-US/app/search/search?sid=1446744769.24062&display.events.type=list&display.page.search.mode=verbose&q=search%20job_id%3D0b99fe46-305d-4cf1-8480-a4b924974217%20sourcetype%3Drails&earliest=-24h%40h&latest=now

At the same time, I was capturing data on the mysql side of things
This was a transaction that timed out due to a lock:
```
-------------
              wait_started: 2015-11-05 13:42:18
                  wait_age: 00:00:09
              locked_table: `shopify_shard_1`.`payment_details`
              locked_index: PRIMARY
               locked_type: RECORD
            waiting_trx_id: 433663523
       waiting_trx_started: 2015-11-05 13:42:18
           waiting_trx_age: 00:00:09
   waiting_trx_rows_locked: 1
 waiting_trx_rows_modified: 1
               waiting_pid: 6948341
             waiting_query: UPDATE `payment_details` SET `order_id` = '1880370370', `updated_at` = '2015-11-05 13:42:18' WHERE `payment_details`.`id` = '911534274' /*application:Shopify,shard_id:1,role:rw,job:CheckoutCompletionJob,job_id:7b730594-8129-4d2d-b5b6-d9e8261db39c*/
           waiting_lock_id: 433663523:261:43260:92
         waiting_lock_mode: X
           blocking_trx_id: 431071130
              blocking_pid: 7060933
            blocking_query: NULL
          blocking_lock_id: 431071130:261:43260:92
        blocking_lock_mode: X
      blocking_trx_started: 2015-11-05 13:40:20
---------
```
The offending blocking transaction was the one with thread_id = 7060933.
According to mysql, that transaction was started at ` 2015-11-05 13:40:20` and it was still open at `2015-11-05 13:42:18`

I had setup up mysql to log _all_ traffic to a file.
With that, I was able to isolate all traffic that was handled by that thread_id (7060933).
You will be able to see all queries with that thread id after it starts its transaction in this [gist](https://gist.github.com/sroysen/0938bd47717558205a00)
The job_id for that transaction was `job_id:0b99fe46-305d-4cf1-8480-a4b924974217`
You will notice that at `151105 13:40:21.555593` (id = 1236178 in that gist) it updates checkouts with the following message:
```
 ``payment_processing_error_message` = 'There was an error processing your payment. Please try again.'`.
```
I assume that was the result of semian interfering with that transaction, hence the job trying to clean up.

The problem now is that that connection/thread_id is used to process another job, and then another, and then another.
MySQL never sees a rollback or commit of the transaction that began at `151105 13:40:20.075844` (id = 1190113), hence it is keeping it open all the time.
Not only that, but since there is still traffic coming from that connection, mysql does not consider it an idle transaction: It's a really active one.
Side effect is that all the other jobs that are going to try to grab that PaymentProcessingJob are going to be locked by this dangling transaction.

byroot:
If it wasn't clear I'm totally yay on this. The exposed problem is very scary and this solves it.

lib/semian/mysql2.rb
camilo: + 1 to non regexp if we can avoid it this is a hot hot path

byroot: Here the bench: https://gist.github.com/byroot/6079402bfcc538948957

Note that the string comparison one could be tricked if there is more then one space between `RELEASE` and `SAVEPOINT`.

fw42: I think avoiding the RegExp is a total micro-optimization... This is gonna be a tiny fraction compared to how long the actual query will take...

camilo: > how long the actual query will take.

That does not matter the queries will be as fast as they were before or after this patch. But it seems that regexing will consume overall more CPU if you say do many queries on 100s of thousands of request per minute that adds up. We had a similar problem with `any?` on strings at some point

fw42: I really don't think that this is a hotpath worth optimizing for. If we have some data that proves me wrong I'm happy to change this, but for now a regexp seems more reasonable. Especially since it's not trivial to correct match this without a regexp. Even the one I have here I could poke holes into... `RELEASE/**/SAVEPOINT` is totally valid SQL but slips through the cracks.. And so is `RELEASE        SAVEPOINT`...

byroot: I'm not advocating for the start_with? version, but I think the `Regexp.union` one is a trivial win.

csfrancis: I think the regexp should be fine. My **ass**umption is that the anchor at the beginning of both `QUERY_WHITELIST` expressions would effectively force them to run in constant-time, regardless of the size of the input string.

csfrancis: And FWIW I agree with Flo that this is a total micro-optimization. @byroot's `Regexp.union` makes sense to me though and is trivial to add.

jnunemaker:
Doesn't this PR mean that if the circuit is open whitelisted queries will fail slowly instead of quickly? Perhaps I'm overlooking something or maybe that is not an issue in practice as the transaction would probably not be opened in the first place.

camilo:
@jnunemaker it is possible that the rollbacks will be slow and deminish capacity, the problem that is being solved here is blocking the rollbacks, makes it so mysql can keep transactions open for very long (seconds/minutes if you are unlucky) that creates a cascading failure if the table you are holding those transactions open on is hot on write operations.

The tradeoff here is risk some slow failures to make sure the state on the db is reset properly and over all the systems is more stable.

byroot:
Not just that, it also make it possible for a transaction to leak to another request / job, which is real bad.

kyewei:
Hey, er... there hasn't been any issues with this in production right? If not, I'd like to merge into this into master so I can go ahead and point the production Gemfile at my branch to test? Production is currently using this branch.

csfrancis:
My two cents is we should merge this as-is for now.

PR #61, opened 2015-11-09
toxiproxy: update to v2.0.0rc1

Sirupsen:
@csfrancis @fw42 will fix your issues since it's compatible with v2.0.0rc1 which is what we run internally now.


csfrancis:
:+1:

PR #62, opened 2015-11-10
Extracted CircuitBreaker states into 3 structures

kyewei:
Older closed PR is this: https://github.com/Shopify/semian/pull/56
Even older massive PR that we're trying to split is this (probably a bit outdated by this point): https://github.com/Shopify/semian/pull/54

Three structures are for holding the `@success`, `@state`, and `@sliding_window` of the `CircuitBreaker` class. 

They are now under `Semian::Simple`, and future SysV implementation will be under `Semian::SysV`. 

Issue that came up and may/may not have been addressed: 

* For methods in the classes that explicitly return `self` as the last line, its so things like `#<<` don't expose the inner array and leak the implementation, and instead return the appropriate wrapper class. Cascading still works: `wrapper_class_instance << 1 << 2`
* The `Atomic` prefix was removed since it is misleading, and we probably won't be providing a `Mutex` or `Monitor`
* it was suggested to rename the enum type from `Semian::Simple::Enum` to `Semian::Simple::State` and be specialized for the three possible items: `:open`, `:closed`, `:half_open`. Similarly, also rename from `Integer` to `SuccessCount`. **I haven't changed it to those names yet**, since it requires removing some code and makes some of the code a bit weird.
 * Pro: specialized, more simple 
 * Con: naming would definitely be weird, even if the in-memory version from `Simple` stores the values directly, the `SysV` enum type delegates the shared-ness to the `SysV` integer type, and would need to keep a `SysV::SuccessCount` type in the `SysV::State` if we use those names. 
* A bit looking into the future, but with the `SysV` versions of the structure needing permissions and a name, `Semian::Simple::SlidingWindow.new(max_size: 4)` and `Semian::SysV::SlidingWindow.new(max_size: 4, permissions: 0660, name: "sample_int")` have different interfaces. Ways to solve
 * do `def initialize(max_size: , **_)`
 * do `def initialize(options = {})` or something of the sort
 * not pass in `permissions` and `name` at all, but they have to come from somewhere else without leaking the implementation, possible from `Semian.register`
* The enum class has an `#increment` because it was suggested that it could be used to iterate through the enums. May/may not be needed.
* Slightly extraneous: it was previously mentioned that test suite naming is a bit off, for example `TestCircuitBreaker` is `circuit_breaker_test.rb`

### DIFF
--- a/lib/semian/circuit_breaker.rb
+++ b/lib/semian/circuit_breaker.rb
@@ -1,13 +1,17 @@
 module Semian
-  class CircuitBreaker
-    attr_reader :state
+  class CircuitBreaker #:nodoc:
+    extend Forwardable
 
-    def initialize(exceptions:, success_threshold:, error_threshold:, error_timeout:)
+    def initialize(name, exceptions:, success_threshold:, error_threshold:, error_timeout:, implementation:)
+      @name = name.to_sym
       @success_count_threshold = success_threshold
       @error_count_threshold = error_threshold
       @error_timeout = error_timeout
       @exceptions = exceptions
-      reset
+
+      @errors = implementation::SlidingWindow.new(max_size: @error_count_threshold)
+      @successes = implementation::Integer.new
+      @state = implementation::State.new
     end
 
     def acquire
@@ -32,8 +36,7 @@ module Semian
     end
 
     def mark_failed(_error)
-      push_time(@errors, @error_count_threshold, duration: @error_timeout)
-
+      push_time(@errors, duration: @error_timeout)
       if closed?
         open if error_threshold_reached?
       elsif half_open?
@@ -43,70 +46,68 @@ module Semian
 
     def mark_success
       return unless half_open?
-      @successes += 1
+      @successes.increment
       close if success_threshold_reached?
     end
 
     def reset
-      @errors = []
-      @successes = 0
+      @errors.clear
+      @successes.reset
       close
+    end
+
+    def destroy
+      @errors.destroy
+      @successes.destroy
+      @state.destroy
     end
 
     private
 
-    def closed?
-      state == :closed
-    end
+    def_delegators :@state, :closed?, :open?, :half_open?
+    private :closed?, :open?, :half_open?
 
     def close
       log_state_transition(:closed)
-      @state = :closed
-      @errors = []
-    end
-
-    def open?
-      state == :open
+      @state.close
+      @errors.clear
     end
 
     def open
       log_state_transition(:open)
-      @state = :open
-    end
-
-    def half_open?
-      state == :half_open
+      @state.open
     end
 
     def half_open
       log_state_transition(:half_open)
-      @state = :half_open
-      @successes = 0
+      @state.half_open
+      @successes.reset
     end
 
     def success_threshold_reached?
-      @successes >= @success_count_threshold
+      @successes.value >= @success_count_threshold
     end
 
     def error_threshold_reached?
-      @errors.count == @error_count_threshold
+      @errors.size == @error_count_threshold
     end
 
     def error_timeout_expired?
-      @errors.last && (@errors.last + @error_timeout < Time.now)
+      time_ms = @errors.last
+      time_ms && (Time.at(time_ms / 1000) + @error_timeout < Time.now)
     end
 
-    def push_time(window, max_size, duration:, time: Time.now)
-      window.shift while window.first && window.first + duration < time
-      window.shift if window.size == max_size
-      window << time
+    def push_time(window, duration:, time: Time.now)
+      # The sliding window stores the integer amount of milliseconds since epoch as a timestamp
+      window.shift while window.first && window.first / 1000 + duration < time.to_i
+      window << (time.to_f * 1000).to_i
     end
 
     def log_state_transition(new_state)
-      return if @state.nil? || new_state == @state
+      return if @state.nil? || new_state == @state.value
 
-      str = "[#{self.class.name}] State transition from #{@state} to #{new_state}."
-      str << " success_count=#{@successes} error_count=#{@errors.count}"
+      str = "[#{self.class.name}] State transition from #{@state.value} to #{new_state}."
+      str << " success_count=#{@successes.value} error_count=#{@errors.size}"
       str << " success_count_threshold=#{@success_count_threshold} error_count_threshold=#{@error_count_threshold}"
       str << " error_timeout=#{@error_timeout} error_last_at=\"#{@error_last_at}\""
       Semian.logger.info(str)
--- a/lib/semian/simple_sliding_window.rb
+++ b/lib/semian/simple_sliding_window.rb
@@ -1,0 +1,43 @@
+module Semian
+  module Simple
+    class SlidingWindow #:nodoc:
+      extend Forwardable
+
+      def_delegators :@window, :size, :pop, :shift, :first, :last
+      attr_reader :max_size
+
+      # A sliding window is a structure that stores the most @max_size recent timestamps
+      # like this: if @max_size = 4, current time is 10, @window =[5,7,9,10].
+      # Another push of (11) at 11 sec would make @window [7,9,10,11], shifting off 5.
+
+      def initialize(max_size:)
+        @max_size = max_size
+        @window = []
+      end
+
+      def resize_to(size)
+        raise ArgumentError.new('size must be larger than 0') if size < 1
+        @max_size = size
+        @window.shift while @window.size > @max_size
+        self
+      end
+
+      def push(value)
+        @window.shift while @window.size >= @max_size
+        @window << value
+        self
+      end
+
+      alias_method :<<, :push
+
+      def clear
+        @window.clear
+        self
+      end
+
+      def destroy
+        clear
+      end
+    end
+  end
+end
--- a/lib/semian/simple_state.rb
+++ b/lib/semian/simple_state.rb
@@ -1,0 +1,43 @@
+module Semian
+  module Simple
+    class State #:nodoc:
+      def initialize
+        reset
+      end
+
+      attr_reader :value
+
+      def open?
+        value == :open
+      end
+
+      def closed?
+        value == :closed
+      end
+
+      def half_open?
+        value == :half_open
+      end
+
+      def open
+        @value = :open
+      end
+
+      def close
+        @value = :closed
+      end
+
+      def half_open
+        @value = :half_open
+      end
+
+      def reset
+        close
+      end
+
+      def destroy
+        reset
+      end
+    end
+  end
+end
--- a/test/simple_integer_test.rb
+++ b/test/simple_integer_test.rb
@@ -1,0 +1,49 @@
+require 'test_helper'
+
+class TestSimpleInteger < MiniTest::Unit::TestCase
+  CLASS = ::Semian::Simple::Integer
+
+  def setup
+    @integer = CLASS.new
+  end
+
+  def teardown
+    @integer.destroy
+  end
+
+  module IntegerTestCases
+    def test_access_value
+      assert_equal(0, @integer.value)
+      @integer.value = 99
+      assert_equal(99, @integer.value)
+      time_now = (Time.now).to_i
+      @integer.value = time_now
+      assert_equal(time_now, @integer.value)
+      @integer.value = 6
+      assert_equal(6, @integer.value)
+      @integer.value = 6
+      assert_equal(6, @integer.value)
+    end
+
+    def test_increment
+      @integer.increment(4)
+      assert_equal(4, @integer.value)
+      @integer.increment
+      assert_equal(5, @integer.value)
+      @integer.increment(-2)
+      assert_equal(3, @integer.value)
+    end
+
+    def test_reset_on_init
+      assert_equal(0, @integer.value)
+    end
+
+    def test_reset
+      @integer.increment(5)
+      @integer.reset
+      assert_equal(0, @integer.value)
+    end
+  end
+
+  include IntegerTestCases
+end
--- a/test/simple_sliding_window_test.rb
+++ b/test/simple_sliding_window_test.rb
@@ -1,0 +1,65 @@
+require 'test_helper'
+
+class TestSimpleSlidingWindow < MiniTest::Unit::TestCase
+  CLASS = ::Semian::Simple::SlidingWindow
+
+  def setup
+    @sliding_window = CLASS.new(max_size: 6)
+    @sliding_window.clear
+  end
+
+  def teardown
+    @sliding_window.destroy
+  end
+
+  module SlidingWindowTestCases
+    def test_sliding_window_push
+      assert_equal(0, @sliding_window.size)
+      @sliding_window << 1
+      assert_sliding_window(@sliding_window, [1], 6)
+      @sliding_window << 5
+      assert_sliding_window(@sliding_window, [1, 5], 6)
+    end
+
+    def test_sliding_window_resize
+      assert_equal(0, @sliding_window.size)
+      @sliding_window << 1 << 2 << 3 << 4 << 5 << 6
+      assert_sliding_window(@sliding_window, [1, 2, 3, 4, 5, 6], 6)
+      @sliding_window.resize_to 6
+      assert_sliding_window(@sliding_window, [1, 2, 3, 4, 5, 6], 6)
+      @sliding_window.resize_to 5
+      assert_sliding_window(@sliding_window, [2, 3, 4, 5, 6], 5)
+      @sliding_window.resize_to 6
+      assert_sliding_window(@sliding_window, [2, 3, 4, 5, 6], 6)
+    end
+
+    def test_sliding_window_edge_falloff
+      assert_equal(0, @sliding_window.size)
+      @sliding_window << 0 << 1 << 2 << 3 << 4 << 5 << 6 << 7
+      assert_sliding_window(@sliding_window, [2, 3, 4, 5, 6, 7], 6)
+      @sliding_window.shift
+      assert_sliding_window(@sliding_window, [3, 4, 5, 6, 7], 6)
+    end
+
+    def resize_to_less_than_1_raises
+      assert_raises ArgumentError do
+        @sliding_window.resize_to 0
+      end
+    end
+  end
+
+  module SlidingWindowUtilityMethods
+    def assert_sliding_window(sliding_window, array, max_size)
+      # Get private member, the sliding_window doesn't expose the entire array
+      data = sliding_window.instance_variable_get("@window")
+      assert_equal(array, data)
+      assert_equal(max_size, sliding_window.max_size)
+    end
+  end
+
+  include SlidingWindowTestCases
+
+  private
+
+  include SlidingWindowUtilityMethods
+end
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 /.bundle/
-/lib/semian/*.so
-/lib/semian/*.bundle
+/lib/**/*.so
+/lib/**/*.bundle
 /tmp/*
 *.gem
 /html/
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -78,7 +78,7 @@ module Semian
   OpenCircuitError = Class.new(BaseError)
 
   def semaphores_enabled?
-    !ENV['SEMIAN_SEMAPHORES_DISABLED']
+    !ENV['SEMIAN_SEMAPHORES_DISABLED'] && Semian.sysv_semaphores_supported?
   end
 
   module AdapterError
@@ -119,10 +119,12 @@ module Semian
   # Returns the registered resource.
   def register(name, tickets:, permissions: 0660, timeout: 0, error_threshold:, error_timeout:, success_threshold:, exceptions: [])
     circuit_breaker = CircuitBreaker.new(
+      name,
       success_threshold: success_threshold,
       error_threshold: error_threshold,
       error_timeout: error_timeout,
       exceptions: Array(exceptions) + [::Semian::BaseError],
+      implementation: ::Semian::Simple,
     )
     resource = Resource.new(name, tickets: tickets, permissions: permissions, timeout: timeout)
     resources[name] = ProtectedResource.new(resource, circuit_breaker)
@@ -154,7 +156,10 @@ require 'semian/circuit_breaker'
 require 'semian/protected_resource'
 require 'semian/unprotected_resource'
 require 'semian/platform'
-if Semian.sysv_semaphores_supported? && Semian.semaphores_enabled?
+require 'semian/simple_sliding_window'
+require 'semian/simple_integer'
+require 'semian/simple_state'
+if Semian.semaphores_enabled?
   require 'semian/semian'
 else
   Semian::MAX_TICKETS = 0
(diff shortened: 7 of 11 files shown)


lib/semian/circuit_breaker.rb
```diff
@@ -1,13 +1,15 @@
 module Semian
-  class CircuitBreaker
-    attr_reader :state
-
-    def initialize(exceptions:, success_threshold:, error_threshold:, error_timeout:)
+  class CircuitBreaker #:nodoc:
+    def initialize(name, exceptions:, success_threshold:, error_threshold:, error_timeout:, type_namespace:)
```
byroot: I'm not a fan of `type_namespace` IMO we should call this an `implementation`

kyewei: OK.

lib/semian/circuit_breaker.rb
```diff
@@ -43,70 +44,77 @@ def mark_failed(_error)
...
-      @successes += 1
+      @successes.increment
       close if success_threshold_reached?
     end
 
     def reset
-      @errors = []
-      @successes = 0
+      @errors.clear
+      @successes.value = 0
```
byroot: I'd add a reset / clear method to Successes too. If we want to expose an API sutitable for atomic operations, we shouldn't encourage to set arbitrary state, just `increment` and `reset`.

lib/semian/circuit_breaker.rb
```diff
@@ -43,70 +44,77 @@ def mark_failed(_error)
...
+      time_ms && (Time.at(time_ms / 1000) + @error_timeout < Time.now)
     end
 
-    def push_time(window, max_size, duration:, time: Time.now)
-      window.shift while window.first && window.first + duration < time
-      window.shift if window.size == max_size
-      window << time
+    def push_time(window, duration:, time: Time.now)
+      # The sliding window stores the integer amount of milliseconds since epoch as a timestamp
+      window.shift while window.first && Time.at(window.first / 1000) + duration < time
```
byroot: Do we actually need to cast to `Time` here? A pure integer timestamp comparison wouldn't do it?

lib/semian/simple_enum.rb
```diff
@@ -0,0 +1,40 @@
...
+        # Assume symbol_list[0] is mapped to 0
+        # Cannot just use #object_id since #object_id for symbols is different in every run
+        # For now, implement a C-style enum type backed by integers
+
+        @sym_to_num = Hash[symbol_list.each_with_index.to_a]
+        @num_to_sym = @sym_to_num.invert
+      end
+    end
+  end
+end
```
byroot: Again I really don't see why you'd need to make `Simple::Enum` backed by a `Simple::Integer`. For `SysV` I see it. But here it's just needless complexity.

kyewei: What sort of error would you want it to raise? I know right now if an invalid one is given `KeyError` is raised because of the `#fetch`

byroot: If we're going in the native exception list then `ArgumentError` is probably the one that makes most sense.

lib/semian/circuit_breaker.rb
kyewei: I've never tried it. But i don't think `FixNum` < `Time` works. So I suppose I'll have to convert the `time` to milliseconds.

byroot:
Just a few comments that I believe are minor. Otherwise LGTM.

lib/semian/simple_enum.rb
kyewei: Hmm. So there might be a small difference in interface then, since the SysV would throw a `KeyError` while this won't?

lib/semian/circuit_breaker.rb
byroot: `window.first / 1000 + duration < time.to_i` is simpler IMO. ut yeah whatever.

I jus think this conditional is a bit unwildly, but no huge deal.

lib/semian/circuit_breaker.rb
kyewei: Does that mean remove `attr_accessor :value`? I can see problems with this in (atleast) the SysV Enum part

```ruby
def value
  @num_to_sym.fetch(@integer.value)
end
```

lib/semian/simple_enum.rb
byroot: No, the Sysv should throw an ArgumentError as well. `KeyError` for an Enum make 0 sense.

You are leaking implementation details to the caller.

Also it doesn't matter that much because this is a programming error exception, the kind you are not supposed to rescue.

lib/semian/circuit_breaker.rb
byroot: Reading the value is fine. Assigning it is wrong.

byroot: So you want `attr_reader` instead of `attr_accessor`

kyewei: Probably should have included this as well: 
```ruby
# From the Enum
def value=(sym)
  @integer.value = @sym_to_num.fetch(sym)
end
```
They don't inherit from each other anymore, it's a `Forwardable`, and I did that to get rid of that diamond dependency. 

Although, if we get rid of it like this, the hash lookup will only be in the `SysV` so there won't be code sharing, and there won't be a diamond dependency issue, and no `def_delegators` needed. What about that? `SysV::Enum` will have to inherit from `SysV::Integer`.

kyewei: So I'm just going to add the `#reset`, but also I'm going to keep the `#value=` there because reasons ^

kyewei:
@Sirupsen any thoughts?

lib/semian/circuit_breaker.rb
byroot: Still make no sense. I'll PR your PR, it's easier to discuss with code.

lib/semian/simple_enum.rb
```diff
@@ -0,0 +1,29 @@
...
+      attr_reader :value
+
+      def value=(sym)
+        raise ArgumentError unless @allowed_symbols.include?(sym)
+        @value = sym
+      end
+
+      def increment(val = 1)
+        @value = @allowed_symbols[(@allowed_symbols.index(@value) + val) % @allowed_symbols.size]
+      end
```
Sirupsen: Is this actually used? The only reason I suggested it before was because you inherited from the integer, since you don't do that anymore, I don't think you actually use this.

lib/semian/simple_sliding_window.rb
```diff
@@ -0,0 +1,45 @@
...
+      attr_reader :max_size
+
+      def initialize(max_size:)
+        @max_size = max_size
+        @window = []
+      end
+
+      # A sliding window is a structure that stores the most @max_size recent timestamps
+      # like this: if @max_size = 4, current time is 10, @window =[5,7,9,10].
+      # Another push of (11) at 11 sec would make @window [7,9,10,11], shifting off 5.
```
Sirupsen: Move this comment to `#initialize`.

lib/semian/simple_enum.rb
kyewei: It's not used in lib code. It is being tested.

kyewei: Do you want to go back to a `undef`

lib/semian/simple_sliding_window.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      def resize_to(size)
+        raise ArgumentError.new('size must be larger than 0') if size < 1
+        @max_size = size
+        @window.shift while @window.size > @max_size
+        self
+      end
```
Sirupsen: Since this is only used in tests, I don't think you should make it part of the public interface.

lib/semian/simple_sliding_window.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      def <<(time_ms)
+        push(time_ms)
+      end
+
+      def push(time_ms)
+        @window.shift while @window.size >= @max_size
+        @window << time_ms
+        self
+      end
```
Sirupsen: `push_time` has a bunch overlap here, it's a little weird, but probably ok.

lib/semian/simple_sliding_window.rb
```diff
@@ -0,0 +1,45 @@
...
+        @max_size = size
+        @window.shift while @window.size > @max_size
+        self
+      end
+
+      def <<(time_ms)
+        push(time_ms)
+      end
+
+      def push(time_ms)
```
Sirupsen: Just call this `value` instead of `time_ms`.

lib/semian/simple_sliding_window.rb
```diff
@@ -0,0 +1,45 @@
...
+      def resize_to(size)
+        raise ArgumentError.new('size must be larger than 0') if size < 1
+        @max_size = size
+        @window.shift while @window.size > @max_size
+        self
+      end
+
+      def <<(time_ms)
+        push(time_ms)
+      end
```
Sirupsen: You can do `alias_method :<<, :push` instead.

test/simple_enum_test.rb
```diff
@@ -0,0 +1,47 @@
...
+      assert_equal :three, @enum.value
+      @enum.increment
+      assert_equal :one, @enum.value
+      @enum.increment(2)
+      assert_equal :three, @enum.value
+      @enum.increment(4)
+      assert_equal :one, @enum.value
+      @enum.increment(0)
+      assert_equal :one, @enum.value
+    end
```
Sirupsen: Again, I don't think this is actually necessary.

test/simple_enum_test.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def test_will_throw_error_when_invalid_symbol_given
+      assert_raises ArgumentError do
+        @enum.value = :four
+      end
+    end
+  end
+
+  include EnumTestCases
```
Sirupsen: Needs a test for `#reset`.

Sirupsen: There is also not a test for providing an invalid value for `#value=`.

lib/semian/simple_enum.rb
```diff
@@ -0,0 +1,29 @@
...
+        @allowed_symbols = symbol_list
+        reset
+      end
+
+      attr_reader :value
+
+      def value=(sym)
+        raise ArgumentError unless @allowed_symbols.include?(sym)
+        @value = sym
+      end
```
Sirupsen: You could just override `=` here instead of `value=`.

lib/semian/simple_enum.rb
```diff
@@ -0,0 +1,29 @@
...
+    class Enum #:nodoc:
+      def initialize(symbol_list:)
+        @allowed_symbols = symbol_list
+        reset
+      end
+
+      attr_reader :value
+
+      def value=(sym)
+        raise ArgumentError unless @allowed_symbols.include?(sym)
```
Sirupsen: Part of the definition of an enumerator is that you can only traverse forwards and/or backwards. You don't provide that guarantee for state transitions here, and they're not tested.

lib/semian/simple_enum.rb
kyewei: I might need `SysV::Enum` to inherit from `SysV::Integer` if byroot's suggestion goes through https://github.com/Shopify/semian/pull/62#discussion-diff-44469194 because thats the only way it can be done if `#value=` disappears or goes private

test/simple_integer_test.rb
```diff
@@ -0,0 +1,48 @@
+require 'test_helper'
+
+class TestSimpleInteger < MiniTest::Unit::TestCase
+  CLASS = ::Semian::Simple::Integer
+
+  def setup
+    @integer = CLASS.new
+    @integer.value = 0
```
Sirupsen: This should not be necessary since you already call `#reset` in the initializer. It's worth writing a test for this.

test/simple_enum_test.rb
kyewei: There is a test for invalid value though, `#test_will_throw_error_when_invalid_symbol_given`

lib/semian/simple_sliding_window.rb
kyewei: `push_time` does front pop until it's within the duration, it's different enough.

Sirupsen: You could make it take a block and then use it like this in the Circuit Breaker:

```ruby
@window.push(value, shift_if: -> (element) { |element| window.first / 1000 + duration < time.to_i }) 
```

That's likely overkill though. Unless you want to play with this, you don't really have to bother.

test/simple_sliding_window_test.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+  end
+
+  module SlidingWindowUtilityMethods
+    def assert_correct_first_and_last_and_size(sliding_window, first, last, size, max_size)
+      assert_equal(first, sliding_window.first)
+      assert_equal(last, sliding_window.last)
+      assert_equal(size, sliding_window.size)
+      assert_equal(max_size, sliding_window.max_size)
+    end
```
Sirupsen: These assertions are hard to read. Since your window for testing is fairly small, I'd go with something very explicit asserting on the entire window:

```ruby
assert_sliding_window [1]
```

Which is just implemented as an `assert_equal` on the underlying array.

With that, you implicitly assert on `max_size` and `size` which is IMO better and much easier to read.

test/simple_sliding_window_test.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    def assert_sliding_windows_in_sync(sliding_window_1, sliding_window_2)
+      # it only exposes ends, size, and max_size, so can only check those
+      assert_equal(sliding_window_1.first, sliding_window_2.first)
+      assert_equal(sliding_window_1.last, sliding_window_2.last)
+      assert_equal(sliding_window_1.size, sliding_window_2.size)
+      assert_equal(sliding_window_1.max_size, sliding_window_2.max_size)
+    end
```
Sirupsen: This should just be an array comparison.

Sirupsen: You also never use this assertion.

test/simple_sliding_window_test.rb
```diff
@@ -0,0 +1,67 @@
...
+      assert_equal(0, @sliding_window.size)
+      @sliding_window << 1 << 2 << 3 << 4 << 5 << 6
+      assert_correct_first_and_last_and_size(@sliding_window, 1, 6, 6, 6)
+      @sliding_window.resize_to 6
+      assert_correct_first_and_last_and_size(@sliding_window, 1, 6, 6, 6)
+      @sliding_window.resize_to 5
+      assert_correct_first_and_last_and_size(@sliding_window, 2, 6, 5, 5)
+      @sliding_window.resize_to 6
+      assert_correct_first_and_last_and_size(@sliding_window, 2, 6, 5, 6)
+    end
```
Sirupsen: This is the only case where you use `resize_to`—it's not actually used in the circuit breaker, so why not remove it altogether?

lib/semian/simple_sliding_window.rb
```diff
@@ -0,0 +1,45 @@
...
+        @max_size = max_size
+        @window = []
+      end
+
+      # A sliding window is a structure that stores the most @max_size recent timestamps
+      # like this: if @max_size = 4, current time is 10, @window =[5,7,9,10].
+      # Another push of (11) at 11 sec would make @window [7,9,10,11], shifting off 5.
+
+      def resize_to(size)
+        raise ArgumentError.new('size must be larger than 0') if size < 1
```
Sirupsen: No test fails if I remove this line of code.

lib/semian/simple_sliding_window.rb
Sirupsen: Oh, this is for when you want to do online resizing of a shared queue. Got it.

lib/semian/simple_sliding_window.rb
```diff
@@ -0,0 +1,43 @@
...
+      # like this: if @max_size = 4, current time is 10, @window =[5,7,9,10].
+      # Another push of (11) at 11 sec would make @window [7,9,10,11], shifting off 5.
+
+      def initialize(max_size:)
+        @max_size = max_size
+        @window = []
+      end
+
+      def resize_to(size)
+        raise ArgumentError.new('size must be larger than 0') if size < 1
```
Sirupsen: No test fails if I remove this.

test/simple_sliding_window_test.rb
kyewei: Do tests magically have access to private members or do I have to do some magic to get them out?

test/simple_enum_test.rb
Sirupsen: Ah, missed that one :)

lib/semian/simple_enum.rb
Sirupsen: I agree with @byroot's suggestion, which is in line with my comment above that this is not an implementation of an enumerator.

lib/semian/simple_enum.rb
kyewei: What do you mean? It already blocks invalid values, otherwise there isn't much ...

lib/semian/circuit_breaker.rb
Sirupsen: Agree with your suggestion here @byroot.

lib/semian/simple_enum.rb
kyewei: Seems weird then to have `#value` and `#=`

Sirupsen:
All my comments are smaller things, looking good Kye. After this iteration, I think it should be ready to go in. Then we'll deploy it to Shopify to make sure all is fine, and start working on the next small PR.

test/simple_sliding_window_test.rb
Sirupsen: To be honest, just use `assert_equal`:

```ruby
assert_equal @sliding_window.window, [1,2,3,4,5,6]
```

lib/semian/simple_enum.rb
Sirupsen: You'd pick one. I don't really care too much.

lib/semian/simple_enum.rb
```diff
@@ -0,0 +1,29 @@
+module Semian
+  module Simple
+    class Enum #:nodoc:
+      def initialize(symbol_list:)
```
Sirupsen: I wouldn't make this argument named.

lib/semian/simple_enum.rb
Sirupsen: Basically it's @byroot's suggestion. If I do:

```ruby
enum = Semian::Simple::Enum.new(symbol_list: [:one, :two, :three])
enum.value = :one
enum.value = :three
```

.. it works. I don't think that's what you want in this case.

lib/semian/simple_enum.rb
kyewei: I was going towards an enumerated type instead of an enumerator like Ruby's.

Sirupsen: Yeah, regardless, I think @byroot is right in thinking of this more as a Circuit Breaker state class than the type. I don't really care too much about which one you chose, as long as it validates the transitions.

kyewei: So what are the valid transitions? `:closed` to `:open` to `:half_open` back to `:closed` ?

lib/semian/simple_enum.rb
kyewei: Why not? Everything else is named already, like the `max_size:`

lib/semian/circuit_breaker.rb
kyewei: The PR doesn't mention anything about `Integer::value=` though. Isn't that the issue here?

lib/semian/simple_enum.rb
Sirupsen: I think `[:closed, :open, :half_open]`, you can only iterate forwards, but you can wrap around. Tests should confirm this.

lib/semian/simple_enum.rb
Sirupsen: yeah, but it's so key to this class and no default makes sense. It doesn't matter much. Up to you.

lib/semian/simple_enum.rb
byroot: No, from `half_open` you can go to either `open` or `closed`

Sirupsen: Ah yeah, you're right. So you can iterate forwards and backwards. I suppose the only transition that's not allowed is from `open` to `closed`.

byroot: You can also go directly from `open` to `closed` if you reset the resource.

byroot: So all transitions are valid in the end.

Sirupsen: Damnit. Maybe it doesn't matter then. :-)

kyewei: Can you go from `:closed` to `:half_open`?

kyewei: Err. So now what.

Sirupsen: I don't envision how that'd happen. You always go straight from `closed` to `half_open`.

Sirupsen: I don't think it matters too much Kye :-)

kyewei: So... get rid of validation?

Sirupsen: Yeah. That's ok with me. You guys convinced me.

byroot: > I don't envision how that'd happen. You always go straight from closed to half_open.

No you don't. That's the only transition that doesn't make sense.

byroot: Also this never have been a validation....

kyewei: Too bad I can't just make a `create_accessors_and_setters` because of english: `:closed` vs `#close`. Otherwise I'd just do something like `names.each { |name| define_method(name) { @value == name)}`.

kyewei: Do we still want an `#increment`

lib/semian/simple_sliding_window.rb
kyewei: `#resize_to_less_than_1_raises` does now.

test/simple_sliding_window_test.rb
kyewei: It won't work if it's done with the not-in-memory one though.

test/simple_sliding_window_test.rb
kyewei: Dynamic resizing, right?

kyewei:
Look over again, I suppose?

Sirupsen:
LGTM

lib/semian/simple_state.rb
```diff
@@ -0,0 +1,49 @@
...
+        @enums = [:closed, :open, :half_open]
+        reset
+      end
+
+      attr_reader :value
+
+      def value=(val)
+        raise ArgumentError unless @enums.include?(val)
+        @value = val
+      end
```
byroot: This method is unused.

byroot:
One last concern (unused method). Once it's solved :sheep:

lib/semian/simple_state.rb
kyewei: So then, if I remove this `test_will_throw_error_when_invalid_symbol_given` will be gone as well?

byroot: Yup. No need to test a method that is not used. Just remove the method and the test althogether.

kyewei:
Alright, I added @byroot 's suggestion and squashed into one commit. @Sirupsen suggested writing an appropriate PR for testing in production before merging it in here.

kyewei:
For now I'm just going to use the other branch `separate_cb_structures-incl-rollback` that is properly rebased and includes the `rollback` branch, the toxiproxy 2.0 update, and the commit from this branch. I'll use it until `rollback` branch is merged in.

Sirupsen:
I'd say merge this :)

PR #63, opened 2015-11-11
Encaspsulate Simple::Enum

byroot:
Here's what I'm trying to express since the beginning. This class should properly encapsulate it's state, and only expose meaningful state transitions.

Your version is way more complicated, just because you can't stop thinking about SysV. 

PS: I don't think Enum is a proper name. `State` or `StateMachine` is probably better. But this is just a sketch anyway.

@kyewei /cc @fw42 @Sirupsen

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 /.bundle/
-/lib/semian/*.so
-/lib/semian/*.bundle
+/lib/**/*.so
+/lib/**/*.bundle
 /tmp/*
 *.gem
 /html/
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -78,7 +78,7 @@ module Semian
   OpenCircuitError = Class.new(BaseError)
 
   def semaphores_enabled?
-    !ENV['SEMIAN_SEMAPHORES_DISABLED']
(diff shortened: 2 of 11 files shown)


kyewei:
But it's less generic.

Sirupsen:
Agree, I don't like how you can go from `closed` to `half_open` currently, or from `open` to `closed`. Those are not possible, and if they can ever happen, they should raise. I'm ok with them not being as generic if it makes them easier to understand. I don't see us using this for anything else.

kyewei:
I wrote this in the other PR, but I meant Enum as an enumerated type, not as an Enumerator. 

However if this makes it simpler I'll do this.

byroot:
> But it's less generic.

YAGNI. Don't get me wrong, I'm a huge proponent of generic code. But in this case we won't ever need a generic state machine. Generalizing this only makes thing more complicated.

> Agree, I don't like how you can go from closed to half_open currently, or from open to closed.

I missed that indeed, but it's trivial to add sanity checks, and to my defense the original code wan't preventing this either.

This is just a rough sketch of what I think would be simpler to understand.

kyewei:
(should probably close this)

byroot:
yup.

Sirupsen:
@byroot thumbs up for writing this PR, was really helpful for the review :+1:

PR #64, opened 2015-11-12
Proposed SysV structures and Mixin functionality

kyewei:
Old massive PR: https://github.com/Shopify/semian/pull/54
Small PR 1: https://github.com/Shopify/semian/pull/62

**Possibly outstanding issues**

* With the `SysV` versions of the structure needing permissions and a name, `Semian::Simple::SlidingWindow.new(max_size: 4)` and `Semian::SysV::SlidingWindow.new(max_size: 4, permissions: 0660, name: "sample_int")` have different interfaces. Ways to solve
 * do `def initialize(max_size: , **)` **current choice**
 * do `def initialize(options = {})` or something of the sort
 * not pass in `permissions` and `name` at all, but they have to come from somewhere else without leaking the implementation, possible from `Semian.register`
* Slightly extraneous: it was previously mentioned that test suite naming is a bit off, for example `TestCircuitBreaker` is `circuit_breaker_test.rb`
* Naming issue of `execute_atomically`: you know what, I changed it to `#synchronize`, similar to classes like `Mutex` and `Monitor`. It's still aliased to `#transaction`
* `#shared?` It's private, and just returns `@using_shared_memory`. Hopefully won't ruffle any feathers, or maybe just the `ivar` is enough?
* It was mentioned (not sure if still applies) that giving `name` and `permissions` to the structures was bad. Personally I disagree, but if so, start a discussion below about it.

**Content**
Alright, here's what I propose for adding in the SysV structures. I didn't include the tests yet, they will obviously fail because there's no C code to provide actual shared memory in this commit. 

The three `SysV` classes inherit directly from the `Simple` equivalents. They also mixin the `Semian::SysVSharedMemory` module, which provides the common functionality. 

From the C perspective, to get the memory allocation and things to work, it needs to do a bunch of things. Certain ways of accomplishing them are IMO worse than others, but I'll list the things that the C side needs to do, and the different ways to do them, and why/how I decided on this way.
C needs to accomplish:
* Hooking the allocation of a Ruby object so a backing C struct is made when the class is instantiated, whether or not `acquire_memory_object` is called. The C code needs this when unpacking the Ruby `VALUE self` object into a proper C `semian_shm_object *`
* Hook the implementations of the shared methods like `#acquire_shared_memory` at a shared level
* Override the specific implementations of a class, like `#value=`, `#<<`, etc 

This is about to get heavy. 

There were 3 ways of doing the above, and there is a diamond problem here. I originally had (1), but I changed to (2).

**(1)**: Have a superclass `SharedMemoryObject`, and `Simple` structures inherit from that. The `SysV` further inherit from the `Simple`
* Pros: Only hook access methods, allocation once, can replace functions at any point in time and it will work
* Cons: Weird to have `class Simple::Integer < SharedMemoryObject` even though it doesn't do sharing at all

**(2)**: Mixin a `SysVSharedMemory` module that does shares common functionality into the `SysV` variants only
* Pros: Makes sense from class organization perspective, mixin a module if you want shared-ness
* Cons: Although you can hook methods like `acquire_memory_object` in one spot, you have to replace the three different allocation methods separately. There is also a timing issue involved. You can't simply do this: 

  ```ruby
module SysVSharedMemory
  ...
  def self.included(base)
    # this doesn't work! The classes that mixin this module 
    # are defined before semian.so are loaded
    # meaning this isn't defined when self.included is called
    call_c_function_to_override_alloc()  
  end
end
  ```
Instead, as I found out, you had to do something like this in the C extension:

  ```C
void Init_semian_shm_object() {
  ...
  call_c_function_to_override_alloc_for(klass)
  ...
}
  ```

**(3)**: Make the `SysV` classes inherit from `SharedMemoryObject`. But what do we do about the common functionality between `Simple` and `SysV` versions? Put that in a module, and include it into things that need the functionality, like this:
* 

  ```ruby
class Simple::Enum
  module EnumImplementation
    ...
  end
  include EnumImplementation
end
class SysV::Enum < SharedMemoryObject
  include Simple::Enum::EnumImplementation
end
  ```
* Pros: Share things. Single point of replacing methods.
* Again, this is ugly.

If you made it this far, :+1:, it was really long.
I chose (2) because it's the most structurally clean, and I have a working solution to the timing problem, and there are no random abuses of Ruby features like Mixins beyond what is necessary.

I've isolated the locking/unlocking, and by doing that, eliminated the lock() and unlock() functions for a synchronize(), and using some delegating I wrap the methods with lock() and `ensure` unlock().

Here's the typical structure of an `acquire` that encompasses the C features as well just so we're on the same page:
```ruby
:acquire_memory_object
  # calculate byte_size, verify semaphores enabled and _acquire exists, else exit
  :_acquire (in C as semian_shm_object_acquire())
    # receive name, permissions, byte_size, check validity
    # hash the name, 
    :bind_initialize_memory_callback
      # It binds a void* function pointer to the C struct. 
         the function is a callback that initializes memory of a given size.
         this is specialized per SysV structure.
    semian_shm_object_acquire_semaphore()
      # call semget, set permissions
    semian_shm_object_synchronize()
      semian_shm_object_lock_without_gvl()
      semian_shm_object_synchronize_with_block() (in begin block)
        semian_shm_object_check_and_resize_if_needed()
          # check requested byte_size against a possible existing memory block, 
             resizing if necesssary by calling shmctl to check sizes
          semian_shm_object_acquire_memory()
            # call callback bound by :bind_initialize_memory_callback to initialize the acquired memory
            # call shmat to attach
      semian_shm_object_synchronize_restore_lock_status() (in ensure block)
  # set @using_shared_memory to true
```

Here's the structure of a synchronized access. Any method defined using `define_method_with_synchronize` is synchronized.
```ruby
:value
  :synchronize
    semian_shm_object_lock_without_gvl()
    semian_shm_object_synchronize_with_block() (in begin block)
      semian_shm_object_check_and_resize_if_needed()
      :value_inner (all the _inner methods are private)
        # access value from C struct
    semian_shm_object_synchronize_restore_lock_status() (in ensure block)
```
`define_method_with_synchronize` basically defines the original method, and then calls `do_with_sync :method_name`

**Discussion, opinions, etc would be great. I also want suggestions as to how to go forward with this.**

A version with everything working is on the branch with the large PR, https://github.com/Shopify/semian/pull/54
@Sirupsen @byroot

### DIFF
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -125,7 +125,7 @@ module Semian
       error_timeout: error_timeout,
       exceptions: Array(exceptions) + [::Semian::BaseError],
       permissions: permissions,
-      implementation: ::Semian::SysV,
+      implementation: Semian.semaphores_enabled? ? ::Semian::SysV : ::Semian::Simple,
     )
     resource = Resource.new(name, tickets: tickets, permissions: permissions, timeout: timeout)
     resources[name] = ProtectedResource.new(resource, circuit_breaker)
@@ -160,11 +160,11 @@ require 'semian/platform'
 require 'semian/simple_sliding_window'
 require 'semian/simple_integer'
 require 'semian/simple_state'
-require 'semian/sysv_shared_memory'
-require 'semian/sysv_sliding_window'
-require 'semian/sysv_integer'
-require 'semian/sysv_state'
 if Semian.semaphores_enabled?
+  require 'semian/sysv_shared_memory'
+  require 'semian/sysv_sliding_window'
+  require 'semian/sysv_integer'
+  require 'semian/sysv_state'
   require 'semian/semian'
 else
   Semian::MAX_TICKETS = 0
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -124,7 +124,8 @@ module Semian
       error_threshold: error_threshold,
       error_timeout: error_timeout,
       exceptions: Array(exceptions) + [::Semian::BaseError],
-      implementation: ::Semian::Simple,
+      permissions: permissions,
+      implementation: ::Semian::SysV,
     )
     resource = Resource.new(name, tickets: tickets, permissions: permissions, timeout: timeout)
     resources[name] = ProtectedResource.new(resource, circuit_breaker)
@@ -159,6 +160,10 @@ require 'semian/platform'
 require 'semian/simple_sliding_window'
 require 'semian/simple_integer'
 require 'semian/simple_state'
+require 'semian/sysv_shared_memory'
+require 'semian/sysv_sliding_window'
+require 'semian/sysv_integer'
+require 'semian/sysv_state'
 if Semian.semaphores_enabled?
   require 'semian/semian'
 else
--- a/lib/semian/circuit_breaker.rb
+++ b/lib/semian/circuit_breaker.rb
@@ -2,16 +2,20 @@ module Semian
   class CircuitBreaker #:nodoc:
     extend Forwardable
 
-    def initialize(name, exceptions:, success_threshold:, error_threshold:, error_timeout:, implementation:)
+    def initialize(name, exceptions:, success_threshold:, error_threshold:, error_timeout:, permissions:, implementation:)
       @name = name.to_sym
       @success_count_threshold = success_threshold
       @error_count_threshold = error_threshold
       @error_timeout = error_timeout
       @exceptions = exceptions
 
-      @errors = implementation::SlidingWindow.new(max_size: @error_count_threshold)
-      @successes = implementation::Integer.new
-      @state = implementation::State.new
+      @errors = implementation::SlidingWindow.new(max_size: @error_count_threshold,
+                                                  name: "#{name}_sysv_sliding_window",
+                                                  permissions: permissions)
+      @successes = implementation::Integer.new(name: "#{name}_sysv_integer",
+                                               permissions: permissions)
+      @state = implementation::State.new(name: "#{name}_sysv_state",
+                                         permissions: permissions)
     end
 
     def acquire
--- a/lib/semian/circuit_breaker.rb
+++ b/lib/semian/circuit_breaker.rb
@@ -10,11 +10,11 @@ module Semian
       @exceptions = exceptions
 
       @errors = implementation::SlidingWindow.new(max_size: @error_count_threshold,
-                                                  name: "#{name}_sysv_sliding_window",
+                                                  name: "#{name}_sliding_window",
                                                   permissions: permissions)
-      @successes = implementation::Integer.new(name: "#{name}_sysv_integer",
+      @successes = implementation::Integer.new(name: "#{name}_integer",
                                                permissions: permissions)
-      @state = implementation::State.new(name: "#{name}_sysv_state",
+      @state = implementation::State.new(name: "#{name}_state",
                                          permissions: permissions)
     end
 
--- a/lib/semian/simple_integer.rb
+++ b/lib/semian/simple_integer.rb
@@ -3,7 +3,7 @@ module Semian
     class Integer #:nodoc:
       attr_accessor :value
 
-      def initialize(**_)
+      def initialize(**)
         reset
       end
 
--- a/lib/semian/simple_integer.rb
+++ b/lib/semian/simple_integer.rb
@@ -3,7 +3,7 @@ module Semian
     class Integer #:nodoc:
       attr_accessor :value
 
-      def initialize
+      def initialize(**_)
         reset
       end
 
--- a/lib/semian/simple_sliding_window.rb
+++ b/lib/semian/simple_sliding_window.rb
@@ -12,7 +12,7 @@ module Semian
       # like this: if @max_size = 4, current time is 10, @window =[5,7,9,10].
       # Another push of (11) at 11 sec would make @window [7,9,10,11], shifting off 5.
 
-      def initialize(max_size:, **_)
+      def initialize(max_size:, **)
         @max_size = max_size
         @window = []
       end
--- a/lib/semian/simple_sliding_window.rb
+++ b/lib/semian/simple_sliding_window.rb
@@ -1,3 +1,5 @@
+require 'forwardable'
+
 module Semian
   module Simple
     class SlidingWindow #:nodoc:
@@ -10,7 +12,7 @@ module Semian
       # like this: if @max_size = 4, current time is 10, @window =[5,7,9,10].
       # Another push of (11) at 11 sec would make @window [7,9,10,11], shifting off 5.
 
-      def initialize(max_size:)
+      def initialize(max_size:, **_)
         @max_size = max_size
         @window = []
       end
--- a/lib/semian/sysv_integer.rb
+++ b/lib/semian/sysv_integer.rb
@@ -1,0 +1,12 @@
+module Semian
+  module SysV
+    class Integer < Semian::Simple::Integer #:nodoc:
+      include SysVSharedMemory
+
+      def initialize(name:, permissions:)
+        data_layout = [:int]
+        super() unless acquire_memory_object(name, data_layout, permissions)
+      end
+    end
+  end
+end
--- a/lib/semian/sysv_shared_memory.rb
+++ b/lib/semian/sysv_shared_memory.rb
@@ -1,0 +1,60 @@
+module Semian
+  module SysVSharedMemory #:nodoc:
+    @type_size = {}
+    def self.sizeof(type)
+      size = (@type_size[type.to_sym] ||= (respond_to?(:_sizeof) ? _sizeof(type.to_sym) : 0))
+      raise TypeError.new("#{type} is not a valid C type") if size <= 0
+      size
+    end
+
+    def semid
+      -1
+    end
+
+    def shmid
+      -1
+    end
+
+    def synchronize(&proc)
+      if respond_to?(:_synchronize) && @using_shared_memory
+        return _synchronize(&proc)
+      else
+        yield if block_given?
+      end
+    end
+
+    alias_method :transaction, :synchronize
+
+    def destroy
+      if respond_to?(:_destroy) && @using_shared_memory
+        _destroy
+      else
+        super
+      end
+    end
+
+    private
+
+    def shared?
+      @using_shared_memory
+    end
+
+    def acquire_memory_object(name, data_layout, permissions)
+      return @using_shared_memory = false unless Semian.semaphores_enabled? && respond_to?(:_acquire)
+
+      byte_size = data_layout.inject(0) { |sum, type| sum + ::Semian::SysVSharedMemory.sizeof(type) }
+      raise TypeError.new("Given data layout is 0 bytes: #{data_layout.inspect}") if byte_size <= 0
+      # Calls C layer to acquire/create a memory block, calling #bind_init_fn in the process, see below
+      _acquire(name, byte_size, permissions)
+      @using_shared_memory = true
+    end
+
+    def bind_init_fn
+      # Concrete classes must implement this in a subclass in C to bind a callback function of type
+      # void (*object_init_fn)(size_t byte_size, void *dest, void *prev_data, size_t prev_data_byte_size);
+      # to location ptr->object_init_fn, where ptr is a semian_shm_object*
+      # It is called when memory needs to be initialized or resized, possibly using previous memory
+      raise NotImplementedError
+    end
+  end
+end
--- a/lib/semian/sysv_shared_memory.rb
+++ b/lib/semian/sysv_shared_memory.rb
@@ -1,12 +1,5 @@
 module Semian
   module SysVSharedMemory #:nodoc:
-    @type_size = {}
-    def self.sizeof(type)
-      size = (@type_size[type.to_sym] ||= (respond_to?(:_sizeof) ? _sizeof(type.to_sym) : 0))
-      raise TypeError.new("#{type} is not a valid C type") if size <= 0
-      size
-    end
-
     def self.included(base)
       def base.do_with_sync(*names)
         names.each do |name|
@@ -30,38 +23,20 @@ module Semian
       -1
     end
 
-    def synchronize(&block)
-      if respond_to?(:_synchronize) && @using_shared_memory
-        return _synchronize(&block)
-      else
-        yield if block_given?
-      end
+    def synchronize
+      yield if block_given?
     end
 
-    alias_method :transaction, :synchronize
-
     def destroy
-      if respond_to?(:_destroy) && @using_shared_memory
-        _destroy
-      else
-        super
-      end
+      super
     end
 
     private
 
-    def shared?
-      @using_shared_memory
-    end
-
-    def acquire_memory_object(name, data_layout, permissions)
-      return @using_shared_memory = false unless Semian.semaphores_enabled? && respond_to?(:_acquire)
-
-      byte_size = data_layout.inject(0) { |sum, type| sum + ::Semian::SysVSharedMemory.sizeof(type) }
-      raise TypeError.new("Given data layout is 0 bytes: #{data_layout.inspect}") if byte_size <= 0
-      # Calls C layer to acquire/create a memory block, calling #bind_initialize_memory_callback in the process, see below
-      _acquire(name, byte_size, permissions)
-      @using_shared_memory = true
+    def acquire_memory_object(*)
+      # Concrete classes must call this method before accessing shared memory
+      # If SysV is enabled, a C method overrides this stub and returns true if acquiring succeeds
+      false
     end
 
     def bind_initialize_memory_callback
--- a/lib/semian/sysv_shared_memory.rb
+++ b/lib/semian/sysv_shared_memory.rb
@@ -7,6 +7,21 @@ module Semian
       size
     end
 
+    def self.included(base)
+      def base.do_with_sync(*names)
+        names.each do |name|
+          new_name = "#{name}_inner".freeze.to_sym
+          alias_method new_name, name
+          private new_name
+          define_method(name) do |*args, &block|
+            synchronize do
+              method(new_name).call(*args, &block)
+            end
+          end
+        end
+      end
+    end
+
     def semid
       -1
     end
@@ -15,9 +30,9 @@ module Semian
       -1
     end
 
-    def synchronize(&proc)
+    def synchronize(&block)
       if respond_to?(:_synchronize) && @using_shared_memory
-        return _synchronize(&proc)
+        return _synchronize(&block)
       else
         yield if block_given?
       end
--- a/lib/semian/sysv_shared_memory.rb
+++ b/lib/semian/sysv_shared_memory.rb
@@ -59,15 +59,15 @@ module Semian
 
       byte_size = data_layout.inject(0) { |sum, type| sum + ::Semian::SysVSharedMemory.sizeof(type) }
       raise TypeError.new("Given data layout is 0 bytes: #{data_layout.inspect}") if byte_size <= 0
-      # Calls C layer to acquire/create a memory block, calling #bind_init_fn in the process, see below
+      # Calls C layer to acquire/create a memory block, calling #bind_initialize_memory_callback in the process, see below
       _acquire(name, byte_size, permissions)
       @using_shared_memory = true
     end
 
-    def bind_init_fn
+    def bind_initialize_memory_callback
       # Concrete classes must implement this in a subclass in C to bind a callback function of type
-      # void (*object_init_fn)(size_t byte_size, void *dest, void *prev_data, size_t prev_data_byte_size);
-      # to location ptr->object_init_fn, where ptr is a semian_shm_object*
+      # void (*initialize_memory)(size_t byte_size, void *dest, void *prev_data, size_t prev_data_byte_size);
+      # to location ptr->initialize_memory, where ptr is a semian_shm_object*
       # It is called when memory needs to be initialized or resized, possibly using previous memory
       raise NotImplementedError
     end
--- a/lib/semian/sysv_sliding_window.rb
+++ b/lib/semian/sysv_sliding_window.rb
@@ -4,7 +4,7 @@ module Semian
       include SysVSharedMemory
 
       def initialize(max_size:, name:, permissions:)
-        data_layout = [:int, :int].concat(Array.new(max_size, :long))
+        data_layout = [:int, :int] + [:long] * max_size
         super(max_size: max_size) unless acquire_memory_object(name, data_layout, permissions)
       end
     end
--- a/lib/semian/sysv_sliding_window.rb
+++ b/lib/semian/sysv_sliding_window.rb
@@ -1,0 +1,12 @@
+module Semian
+  module SysV
+    class SlidingWindow < Semian::Simple::SlidingWindow #:nodoc:
+      include SysVSharedMemory
+
+      def initialize(max_size:, name:, permissions:)
+        data_layout = [:int, :int].concat(Array.new(max_size, :long))
+        super(max_size: max_size) unless acquire_memory_object(name, data_layout, permissions)
+      end
+    end
+  end
+end
--- a/lib/semian/sysv_state.rb
+++ b/lib/semian/sysv_state.rb
@@ -7,12 +7,12 @@ module Semian
       extend Forwardable
 
       def_delegators :@integer, :semid, :shmid, :synchronize, :transaction,
-                     :shared?, :acquire_memory_object, :bind_init_fn
-      private :shared?, :acquire_memory_object, :bind_init_fn
+                     :shared?, :acquire_memory_object, :bind_initialize_memory_callback
+      private :shared?, :acquire_memory_object, :bind_initialize_memory_callback
 
       def initialize(name:, permissions:)
         @integer = Semian::SysV::Integer.new(name: name, permissions: permissions)
-        initialize_lookup([:closed, :open, :half_open])
+        initialize_lookup
       end
 
       def open
@@ -46,12 +46,12 @@ module Semian
         @integer.value = @sym_to_num.fetch(sym) { raise ArgumentError }
       end
 
-      def initialize_lookup(symbol_list)
+      def initialize_lookup
         # Assume symbol_list[0] is mapped to 0
         # Cannot just use #object_id since #object_id for symbols is different in every run
         # For now, implement a C-style enum type backed by integers
 
-        @sym_to_num = Hash[symbol_list.each_with_index.to_a]
+        @sym_to_num = {closed: 0, open: 1, half_open: 2}
         @num_to_sym = @sym_to_num.invert
       end
     end
--- a/lib/semian/sysv_state.rb
+++ b/lib/semian/sysv_state.rb
@@ -1,0 +1,59 @@
+require 'forwardable'
+
+module Semian
+  module SysV
+    class State < Semian::Simple::State #:nodoc:
+      include SysVSharedMemory
+      extend Forwardable
+
+      def_delegators :@integer, :semid, :shmid, :synchronize, :transaction,
+                     :shared?, :acquire_memory_object, :bind_init_fn
+      private :shared?, :acquire_memory_object, :bind_init_fn
+
+      def initialize(name:, permissions:)
+        @integer = Semian::SysV::Integer.new(name: name, permissions: permissions)
+        initialize_lookup([:closed, :open, :half_open])
+      end
+
+      def open
+        self.value = :open
+      end
+
+      def close
+        self.value = :closed
+      end
+
+      def half_open
+        self.value = :half_open
+      end
+
+      def reset
+        close
+      end
+
+      def destroy
+        reset
+        @integer.destroy
+      end
+
+      def value
+        @num_to_sym.fetch(@integer.value) { raise ArgumentError }
+      end
+
+      private
+
+      def value=(sym)
+        @integer.value = @sym_to_num.fetch(sym) { raise ArgumentError }
+      end
+
+      def initialize_lookup(symbol_list)
+        # Assume symbol_list[0] is mapped to 0
+        # Cannot just use #object_id since #object_id for symbols is different in every run
+        # For now, implement a C-style enum type backed by integers
+
+        @sym_to_num = Hash[symbol_list.each_with_index.to_a]
+        @num_to_sym = @sym_to_num.invert
+      end
+    end
+  end
+end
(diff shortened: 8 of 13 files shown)


lib/semian/simple_integer.rb
```diff
@@ -3,7 +3,7 @@ module Simple
     class Integer #:nodoc:
       attr_accessor :value
 
-      def initialize
+      def initialize(**_)
```
byroot: You don't need the `_` here. If you disregard everything.

lib/semian/sysv_shared_memory.rb
```diff
@@ -0,0 +1,75 @@
...
+      return @using_shared_memory = false unless Semian.semaphores_enabled? && respond_to?(:_acquire)
+
+      byte_size = data_layout.inject(0) { |sum, type| sum + ::Semian::SysVSharedMemory.sizeof(type) }
+      raise TypeError.new("Given data layout is 0 bytes: #{data_layout.inspect}") if byte_size <= 0
+      # Calls C layer to acquire/create a memory block, calling #bind_init_fn in the process, see below
+      _acquire(name, byte_size, permissions)
+      @using_shared_memory = true
+    end
+
+    def bind_init_fn
```
byroot: Please avoid abréviations like `fn` in ruby code. It's all but idiomatic, and we're in 2015, we have wide screens and autocompletion.

lib/semian/sysv_sliding_window.rb
```diff
@@ -0,0 +1,12 @@
+module Semian
+  module SysV
+    class SlidingWindow < Semian::Simple::SlidingWindow #:nodoc:
+      include SysVSharedMemory
+
+      def initialize(max_size:, name:, permissions:)
+        data_layout = [:int, :int].concat(Array.new(max_size, :long))
```
byroot:
```ruby
data_layout = [:int, :int] + [:long] * max_size
```
Would be easier to read IMO /nitpic

lib/semian/sysv_state.rb
```diff
@@ -0,0 +1,59 @@
...
+        @integer.value = @sym_to_num.fetch(sym) { raise ArgumentError }
+      end
+
+      def initialize_lookup(symbol_list)
+        # Assume symbol_list[0] is mapped to 0
+        # Cannot just use #object_id since #object_id for symbols is different in every run
+        # For now, implement a C-style enum type backed by integers
+
+        @sym_to_num = Hash[symbol_list.each_with_index.to_a]
+        @num_to_sym = @sym_to_num.invert
```
byroot: I still don't get why you don't hardcode that mapping once and for all.

byroot:
> do def initialize(max_size: , **_) current choice

Fine by me. As said though `**` is enough.

> Naming issue of execute_atomically: you know what, I changed it to #synchronize

Synchronize is definitely much better.

> #shared? It's private, and just returns @using_shared_memory. Hopefully won't ruffle any feathers, or maybe just the ivar is enough?

Nah, an accessor is always good.


The rest of the PR is out of my my expertise.

kyewei:
@csfrancis Any suggestions or comments on the C side? A working version can be found [here](https://github.com/Shopify/semian/pull/54), but probably reading the long PR description would be enough.

lib/semian/simple_integer.rb
kyewei: So just `def initialize(**)` ? I didn't know that.

lib/semian/sysv_shared_memory.rb
```diff
@@ -0,0 +1,75 @@
...
+  module SysVSharedMemory #:nodoc:
+    @type_size = {}
+    def self.sizeof(type)
+      size = (@type_size[type.to_sym] ||= (respond_to?(:_sizeof) ? _sizeof(type.to_sym) : 0))
+      raise TypeError.new("#{type} is not a valid C type") if size <= 0
+      size
+    end
+
+    def self.included(base)
+      def base.do_with_sync(*names)
```
kyewei: @byroot probably the most interesting change that you could provide an opinion for is this function here. I call it from C to replace a regular accessor or setter that accesses shared memory and is not a atomic call, and wrap it in a `synchronize`. I wrap all the functions with this, so it's essentially a `do_with_sync :value=, :value, :reset, :increment` and `do_with_sync :size=, :max_size, :<<, :push, :pop` etc

I just want to make sure there are no disagreements here.

Relevant C code is 
https://github.com/Shopify/semian/blob/circuit-breaker-per-host/ext/semian/semian_shared_memory_object.c#L319-L324 
where `define_method_with_synchronize` calls `do_with_sync`

and 
https://github.com/Shopify/semian/pull/54/files#diff-99c3116b9c2c62cc89a4ae4cc84fadaaR262
which is where `define_method_with_synchronize` is used in place of `rb_define_method`

byroot: I'd prefer to avoid this kind of monkey patching. 

I'd rather prefer have a noop `sychronize` method in `Simple`.

kyewei: While looking over the C code base, one of the scarier things was that because I provided the lock() and unlock(), it was very difficult to track where I had to re-unlock, since errors could occur in the middle of a function, and then you'd have to insert an unlock() there. I then opted to write `ensures` in, but doing so manually would add twice the number of functions and make things very manual. 

Here's the c-ensure:
```C
return rb_ensure(semian_shm_object_synchronize_with_block, self, semian_shm_object_synchronize_restore_lock_status, (VALUE)&status);
```
Essentially each function needed a wrapper, and probably also some custom struct to overcome the one-argument limit of the function. This naturally lead to me doing it in Ruby instead since it doesn't have these limitations, which is why `synchronize` does what it does. IMO it's not a big deal since it only does it once per class, reduces code and complexity by a bit.

With regards to the noop `synchronize`, do you mean something more along the lines of (1) , with SharedMemoryObject -> Simple::Integer -> SysV::Integer ? Otherwise I don't understand what you mean by noop `synchronize`. Unless you mean just having one there to keep the interfaces the same?

lib/semian.rb
```diff
@@ -159,6 +160,10 @@ def resources
 require 'semian/simple_sliding_window'
 require 'semian/simple_integer'
 require 'semian/simple_state'
+require 'semian/sysv_shared_memory'
```
csfrancis: Should these `require`s be conditional depending on whether the target platform supports SysV?

lib/semian/circuit_breaker.rb
```diff
@@ -2,16 +2,20 @@ module Semian
...
       @success_count_threshold = success_threshold
       @error_count_threshold = error_threshold
       @error_timeout = error_timeout
       @exceptions = exceptions
 
-      @errors = implementation::SlidingWindow.new(max_size: @error_count_threshold)
-      @successes = implementation::Integer.new
-      @state = implementation::State.new
+      @errors = implementation::SlidingWindow.new(max_size: @error_count_threshold,
+                                                  name: "#{name}_sysv_sliding_window",
```
csfrancis: I don't think these names should have `sysv` in them.

lib/semian.rb
```diff
@@ -124,7 +124,8 @@ def register(name, tickets:, permissions: 0660, timeout: 0, error_threshold:, er
       error_threshold: error_threshold,
       error_timeout: error_timeout,
       exceptions: Array(exceptions) + [::Semian::BaseError],
-      implementation: ::Semian::Simple,
+      permissions: permissions,
+      implementation: ::Semian::SysV,
```
csfrancis: `implementation` should probably be passed as an argument.

lib/semian/sysv_integer.rb
```diff
@@ -0,0 +1,12 @@
+module Semian
+  module SysV
+    class Integer < Semian::Simple::Integer #:nodoc:
```
csfrancis: Why are you inheriting from `Simple::Integer`? Won't the implementation be completely different?

lib/semian/sysv_shared_memory.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    def synchronize(&block)
+      if respond_to?(:_synchronize) && @using_shared_memory
+        return _synchronize(&block)
+      else
+        yield if block_given?
+      end
+    end
+
+    alias_method :transaction, :synchronize
```
csfrancis: What's this for?

lib/semian/sysv_integer.rb
kyewei: It's to support a fallback in case SysV isn't activated for some reason. The C api will replace all the useful methods, like `#value`, `#value=`, `#increment`, `#reset`

lib/semian/sysv_shared_memory.rb
kyewei: I had another name for it originally, `#execute_atomically`, so it was suggested that `#transaction` might be better. But then I thought of `#synchronize`, which is probably the best name for that functionality. So I might remove this `alias_method` altogether.

test/sysv_sliding_window_test.rb
```diff
@@ -0,0 +1,22 @@
...
+                                name: 'TestSysVSlidingWindow',
+                                permissions: 0660)
+    @sliding_window.clear
+  end
+
+  def teardown
+    @sliding_window.destroy
+  end
+
+  include TestSimpleSlidingWindow::SlidingWindowTestCases
```
Sirupsen: Shouldn't there be a test proving it's shared and that access is syncronized where needed?

test/sysv_integer_test.rb
```diff
@@ -0,0 +1,16 @@
+require 'test_helper'
+
+class TestSysVInteger < MiniTest::Unit::TestCase
```
Sirupsen: What about proving the integer is shared?

lib/semian/sysv_state.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+
+      def reset
+        close
+      end
+
+      def destroy
+        reset
+        @integer.destroy
+      end
```
Sirupsen: `super` instead of `reset`?

lib/semian/sysv_state.rb
```diff
@@ -0,0 +1,59 @@
...
+        self.value = :closed
+      end
+
+      def half_open
+        self.value = :half_open
+      end
+
+      def reset
+        close
+      end
```
Sirupsen: This implementation is identical to `Simple::State`, why override?

lib/semian/sysv_state.rb
```diff
@@ -0,0 +1,59 @@
...
+        self.value = :open
+      end
+
+      def close
+        self.value = :closed
+      end
+
+      def half_open
+        self.value = :half_open
+      end
```
Sirupsen: In the `Simple::State` we should set `value` as an `attr_accessor` or `attr_writer` so you can just re-use the implementation from there.

lib/semian/sysv_state.rb
```diff
@@ -0,0 +1,59 @@
+require 'forwardable'
+
+module Semian
+  module SysV
+    class State < Semian::Simple::State #:nodoc:
+      include SysVSharedMemory
+      extend Forwardable
+
+      def_delegators :@integer, :semid, :shmid, :synchronize, :transaction,
+                     :shared?, :acquire_memory_object, :bind_init_fn
```
Sirupsen: Why do you need to define all of these?

lib/semian/sysv_state.rb
```diff
@@ -0,0 +1,59 @@
+require 'forwardable'
+
+module Semian
+  module SysV
+    class State < Semian::Simple::State #:nodoc:
```
Sirupsen: I made a couple of comments here, in general I think you can simplify this implementation by sharing more functionality from `Semian::Simple::State`. You may have to modify that class a little, but I think that's fine.

lib/semian/sysv_shared_memory.rb
```diff
@@ -0,0 +1,75 @@
+module Semian
+  module SysVSharedMemory #:nodoc:
```
Sirupsen: Haven't had the opportunity to review this yet, but looks quite complicated at first glance

test/sysv_sliding_window_test.rb
kyewei: I didn't add the shared tests in the commit, the tests and build would fail. Do you want me to include it?

lib/semian/sysv_state.rb
kyewei: It's not, it uses `self` instead. 

Hmm. I think the better way is to change the `Simple::State` ones into `self.value`.

lib/semian/sysv_shared_memory.rb
kyewei: Most of the complications is in the C code, so read the PR description for a rundown of what happens.

lib/semian/sysv_state.rb
kyewei: I think it was @byroot that said something about not liking a `#value=`, so I left it out and used `attr_reader`

byroot: I don't want `#value=` to be the public API. I don't give a shit about internal private APIs.

kyewei: Awesome.

kyewei:
@Sirupsen wanted to see the tests, so I'll include them now. Some of those new tests will fail because the C component isn't actually included yet.

lib/semian.rb
Sirupsen: Yeah, I don't quite understand how the implementation can be hardcoded. How does the fallback work then? It sounds like in that case there's no test to make sure it goes to the Simple fallback when SysV is not available / disabled.

lib/semian.rb
Sirupsen: Agree.

lib/semian/simple_sliding_window.rb
```diff
@@ -1,3 +1,5 @@
+require 'forwardable'
```
Sirupsen: I don't think you need this require here.

Sirupsen: Why do you need this now?

lib/semian/sysv_integer.rb
Sirupsen: I think this is a too unexpected way to provide the fallback. The fallback shouldn't be the SysV classes changing their implementation with `super` and inheritance, it should be done at boot-time. This is dangerous and would be incredibly difficult to debug.

lib/semian/sysv_state.rb
```diff
@@ -0,0 +1,43 @@
...
+        @integer.value = @sym_to_num.fetch(sym) { raise ArgumentError }
+      end
+
+      def initialize_lookup
+        # Assume symbol_list[0] is mapped to 0
+        # Cannot just use #object_id since #object_id for symbols is different in every run
+        # For now, implement a C-style enum type backed by integers
+
+        @sym_to_num = {closed: 0, open: 1, half_open: 2}
+        @num_to_sym = @sym_to_num.invert
```
Sirupsen: These never change, so why not just make them constants?

test/sysv_sliding_window_test.rb
Sirupsen: No, you make a good point, let's wait with them—but make sure to not forget them. We can't merge this unless it's green.

Sirupsen:
I think that *2* is the right approach as well.

lib/semian/sysv_shared_memory.rb
```diff
@@ -0,0 +1,75 @@
...
+    @type_size = {}
+    def self.sizeof(type)
+      size = (@type_size[type.to_sym] ||= (respond_to?(:_sizeof) ? _sizeof(type.to_sym) : 0))
```
Sirupsen: Why not just use `@type_size[type.to_sym]` and make this easier to read? I assume the only reason you need the `respond_to?` is because you don't have the C implementation just yet?

lib/semian/sysv_shared_memory.rb
```diff
@@ -0,0 +1,75 @@
...
+    private
+
+    def shared?
+      @using_shared_memory
+    end
+
+    def acquire_memory_object(name, data_layout, permissions)
+      return @using_shared_memory = false unless Semian.semaphores_enabled? && respond_to?(:_acquire)
+
+      byte_size = data_layout.inject(0) { |sum, type| sum + ::Semian::SysVSharedMemory.sizeof(type) }
```
Sirupsen: Why does `sizeof` need to be a class method?

lib/semian/sysv_shared_memory.rb
```diff
@@ -0,0 +1,75 @@
...
+    end
+
+    private
+
+    def shared?
+      @using_shared_memory
+    end
+
+    def acquire_memory_object(name, data_layout, permissions)
+      return @using_shared_memory = false unless Semian.semaphores_enabled? && respond_to?(:_acquire)
```
Sirupsen: I don't think this logic should be done at runtime, but always at boot time.

lib/semian/sysv_shared_memory.rb
```diff
@@ -0,0 +1,75 @@
...
+      else
+        super
+      end
+    end
+
+    private
+
+    def shared?
+      @using_shared_memory
+    end
```
Sirupsen: I still don't understand why you want this method (at least not if you do the boot time selection of the driver).

lib/semian/sysv_shared_memory.rb
```diff
@@ -0,0 +1,75 @@
...
+      # Calls C layer to acquire/create a memory block, calling #bind_initialize_memory_callback in the process, see below
+      _acquire(name, byte_size, permissions)
+      @using_shared_memory = true
+    end
+
+    def bind_initialize_memory_callback
+      # Concrete classes must implement this in a subclass in C to bind a callback function of type
+      # void (*initialize_memory)(size_t byte_size, void *dest, void *prev_data, size_t prev_data_byte_size);
+      # to location ptr->initialize_memory, where ptr is a semian_shm_object*
+      # It is called when memory needs to be initialized or resized, possibly using previous memory
```
Sirupsen: I feel like the name could be better then.

lib/semian/sysv_shared_memory.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    alias_method :transaction, :synchronize
+
+    def destroy
+      if respond_to?(:_destroy) && @using_shared_memory
+        _destroy
+      else
+        super
+      end
+    end
```
Sirupsen: Why wouldn't the C implementation just override this (given the driver check is done at boot time).

Sirupsen:
Fundamentally I think what makes this PR more complicated than it needs to be is that the fallback is done within the SysV driver, instead of at boot time as @csfrancis also pointed out. This will simplify I bunch of things because you can make more assumptions.

test/sysv_sliding_window_test.rb
```diff
@@ -0,0 +1,160 @@
+require 'test_helper'
+
+class TestSysVSlidingWindow < MiniTest::Unit::TestCase
+  CLASS = ::Semian::SysV::SlidingWindow
```
Sirupsen: This is a little bit of a nitpick, but I'd call this `KLASS` instead of make it a method. `CLASS` is too close to the reserved keyword.

lib/semian/sysv_shared_memory.rb
Sirupsen: @kyewei I think what he means with noop syncronize is that the SysV implementations continue to inherit from `Simple::Integer` and friends, which then look like this:

```ruby
class Semian::Simple::Integer
  # .. stuff
  def increment(delta)
    synchronize { value += delta }
  end

  private
  def synchronize
    yield
  end
  # more stuff
end
```

When this mixin is included, it just overrides it. The implementation can then be simplified when all the fallback checking is done at boot-time because you can just override it in C, and not have it in Ruby.

lib/semian/sysv_shared_memory.rb
```diff
@@ -0,0 +1,75 @@
...
+      -1
+    end
+
+    def synchronize(&block)
+      if respond_to?(:_synchronize) && @using_shared_memory
+        return _synchronize(&block)
+      else
+        yield if block_given?
+      end
+    end
```
Sirupsen: Example of a function that could probably just be implemented in C and would benefit from the simplification of not doing run-time fallbacks.

lib/semian/sysv_shared_memory.rb
kyewei: It's for memoizing the sizes after it's read once from C. I don't want to hardcode sizes, they differ by platform.

Sirupsen: I agree with that, but I don't see why you need the local variable when you have the class instance variable.

kyewei: Oh. For raising the TypeError. Maybe that should be moved to C instead?

lib/semian/sysv_shared_memory.rb
kyewei: I see what you're saying, but that also doesn't work in protecting lets say.. `Semian::SysV::Integer#value`. If the `#value` is not protected in C code, and it directly replaces the `SysV::Integer`, there is no entry point to override and replace with `synchronize { ... }` except through what `do_with_sync` does. If you do protect in C, well, that goes back to the problem of making 2x the functions to call `rb_ensure(callback, arg1, ensure_fn, arg2)`

lib/semian.rb
kyewei: How I did this before was that the `SysV::` classes can themselves function as the fallback, since they inherit from the `Simple::` classes. Distinguishing between the two states was the reason I had `@using_shared_memory` as a variable.
I guess what you're imagining is more of a 
```ruby
implementation: semaphores_enabled? ? ::Semian::SysV : ::Semian::Simple
```

lib/semian/sysv_state.rb
```diff
@@ -0,0 +1,31 @@
...
+      def_delegators :@integer, :semid, :shmid, :synchronize, :acquire_memory_object,
+                     :bind_initialize_memory_callback, :destroy
+      private :acquire_memory_object, :bind_initialize_memory_callback
+
+      def initialize(name:, permissions:)
+        @integer = Semian::SysV::Integer.new(name: name, permissions: permissions)
+      end
+
+      def value
+        NUM_TO_SYM.fetch(@integer.value) { raise ArgumentError }
```
Sirupsen: If you did `alias_method :to_i, :value` you wouldn't need to override these. Duck typing hard!

lib/semian/sysv_state.rb
```diff
@@ -0,0 +1,31 @@
+require 'forwardable'
+
+module Semian
+  module SysV
+    class State < Semian::Simple::State #:nodoc:
+      include SysVSharedMemory
+      extend Forwardable
+
+      SYM_TO_NUM = {closed: 0, open: 1, half_open: 2}.freeze
+      NUM_TO_SYM = SYM_TO_NUM.invert.freeze
```
Sirupsen: Declaring these in `SIMPLE` and validating against them might be more appropriate, as other drivers would likely want these.

lib/semian.rb
```diff
@@ -159,6 +160,10 @@ def resources
 require 'semian/simple_sliding_window'
 require 'semian/simple_integer'
 require 'semian/simple_state'
+require 'semian/sysv_shared_memory'
+require 'semian/sysv_sliding_window'
+require 'semian/sysv_integer'
+require 'semian/sysv_state'
```
Sirupsen: You should put these in the block below.

lib/semian.rb
```diff
@@ -159,6 +160,10 @@ def resources
...
 if Semian.semaphores_enabled?
   require 'semian/semian'
```
Sirupsen: Huh? Only require `semian/semian` is semaphores are enabled?

lib/semian/sysv_integer.rb
Sirupsen: At this point, isn't it easier to write this entire thing in C? You don't need the `unless` below anymore.

lib/semian/sysv_shared_memory.rb
```diff
@@ -0,0 +1,50 @@
...
+          alias_method new_name, name
+          private new_name
+          define_method(name) do |*args, &block|
+            synchronize do
+              method(new_name).call(*args, &block)
+            end
+          end
+        end
+      end
+    end
```
Sirupsen: You can still kill this meta-programming by the no-op `synchronize` in the `Simple` implementation, as showed with code in another comment.

lib/semian/sysv_shared_memory.rb
```diff
@@ -0,0 +1,50 @@
...
+    def semid
+      -1
+    end
+
+    def shmid
+      -1
+    end
+
+    def synchronize
+      yield if block_given?
```
Sirupsen: This is not quite what I meant by a no-op synchronize, see my code-example in the other comment. This code should be in the `Simple` classes, not here where it should do the right thing.

lib/semian/sysv_shared_memory.rb
Sirupsen: As they get increasingly simpler as we iterate them closer to the Simple implementation, do we even need these exposed to Ruby?

lib/semian/sysv_sliding_window.rb
```diff
@@ -0,0 +1,12 @@
+module Semian
+  module SysV
+    class SlidingWindow < Semian::Simple::SlidingWindow #:nodoc:
```
Sirupsen: Ditto on: Wouldn't this be better to just have implemented in C?

lib/semian/sysv_shared_memory.rb
kyewei: https://github.com/Shopify/semian/pull/64#discussion_r45502636

You probably didn't see it, I responded saying how a no-op synchronize doesn't work in all the cases.

lib/semian/sysv_integer.rb
kyewei: I mean, the issue is whether the `SysV` structure should be able to itself fallback to a `Simple`'s functionality. That's what I did here to make that work. 
I don't see a justification for writing it in C if it's doing exactly what it's doing here, except in more verbose C.

lib/semian/sysv_shared_memory.rb
kyewei: If nothing else, this needs to be clearly stated to be overriden, etc. Also, isn't this an old diff? That `acquire_memory_object` is old. The current one just returns `false`

lib/semian/sysv_state.rb
kyewei: Simple has the lookup hard-coded, as requested earlier. If other drivers want them, why wouldn't then just use `Semian::SysV::State::SYM_TO_NUM`, etc?

lib/semian/sysv_integer.rb
Sirupsen: Yeah, I meant with inheritance from C, but that probably doesn't give you anything and gives more poor visibility.

lib/semian/sysv_shared_memory.rb
Sirupsen: I understand your concern, but can't you just make `synchronize` detect whether it's being recursed and in that case not grab the lock again?

kyewei: The issue is that if you define `#value` to return `@value` in `Simple`, your ability to share that with `SysV` is out since the `SysV` ones don't define `@value`. If you define `#value` to return `self.value`, well, then you have infinite looping.

I don't think the no-op synchronize works at all in this situation, is what I'm saying

Sirupsen: I'll try to put what I meant into code, I think all you need to do is make sure you don't try to grab the mutex again if you nest (with the mutex here being the shared memory one, I'm just using a pure Ruby example):

```ruby
class A
  def value
    synchronize { @value }
  end

  private
  def synchronize
    yield
  end
end

class B < A
  def initialize
    @mutex = Mutex.new
  end

  def increment(val = 1)
    synchronize { self.value += val }
  end 

  private
  def synchronize
    return yield if @nested
    @mutex.synchronize {
       @nested = true
       yield 
     }
  end
end 
```

kyewei: Yeah but how do you make sure the `SysV::Integer#value` is called, if all you do is `Simple::Integer#value`? The thing is you can't reuse `@value` at all since it doesn't exist in `SysV::Integer`. 

The C version and the Ruby version has two different ways of storing the value, and their sole accessor is through their `#value`. You can't both override it and call the original, i.e. you can't have your cake and eat it too.

Sirupsen: Ah. What about exposing it under a `raw_value` alias?

kyewei: That's what the `do_with_sync` is doing, it aliases the unprotected to `#{method_name}_inner` and makes the protected `#{method_name}`

Sirupsen: But why do you have to do it for any other method than `#value`?

kyewei: `#increment`, gets and sets value
`#reset` delegates to `#value=`
`#value=` should definitely be atomic

For the SlidingWindow ones: 
`"size", "max_size", "resize_to", "<<", "push", "pop", "shift", "unshift", "clear", "first", "last"`, either they are accessors or they modify the array, so they are necessary. 

You **could** make the case to remove the synchronize from the accessors, but why would you make it less safe, for example.
It makes sense to just make it uniform from then on.

lib/semian/sysv_state.rb
Sirupsen: Because `Simple` is the canonical implementation everything inherits from, not SysV.

lib/semian/sysv_shared_memory.rb
Sirupsen: The things I still don't understand why you can't just do this for `#value` and reuse the implementation for the other ones, so you don't need this ugly meta programming that makes it hard to grep for and grok what's going on.

kyewei: Because you'd still have the same problem for each of those `<<`, `pop`, `shift`,`clear` etc. UNLESS you provide some uniform `[]` and make everything delegate to that, but I distinctly remember the motivation for not having `[]` that was to make it a `SlidingWindow`, not an `Array`

Sirupsen: Let's discuss this IRL after lunch

lib/semian/sysv_state.rb
kyewei: Then there's going to be the question for why it's there, since it seems out of place in a class that hardcodes them. Then it leads to `why is it hardcoded` and stuff happens.

PR #71, opened 2015-11-27
Net::HTTP resource invalidation fix

kyewei:
From investigating the issue with the flakey test, I found that if something like this occured:
```ruby
Net::HTTP.semian_configuration = proc { DEFAULT_SEMIAN_OPTIONS.merge(name: 'localhost_80') }
http = Net::HTTP.new('localhost', 80)
http.semian_resource.acquire {}
Net::HTTP.semian_configuration = proc { DEFAULT_SEMIAN_OPTIONS.merge(name: 'default') }
assert_equal 'nethttp_default', http.semian_resource.name #FAILS
```
This actually occurs when using the Toxiproxy ruby library for example. For communicating with the daemon, the ruby client does something to this effect: `Toxiproxy.http ||= Net::HTTP.new('localhost', 8474)`

The above fails because `@semian_resource` is cached. So I uncached it by checking to see if the `proc` changes and setting it to `nil` if it does. However, it then occurred to me that since the `@semian_options` are passed in when the `Resource` is made, if the `proc` produces different options each time, it would still not take effect. IMO it is *very difficult* to have a `Resource` that keeps up with a varying `semian_option` if both `host` and `port` stay the same. We can only justify re-creating a `Resource` when the proc changes, since that means the settings have been changed as a whole.

Therefore, it makes sense to instead make clear to the user that `proc`s that produce different `semian_options` should be an invalid input.

With that in mind then, we can cache the `@semian_options` again. 

EDIT: stance has changed to not allowing `semian_configuration` to be changed once it is set. So it can only be set once, otherwise it raises an error. Reasoning is [here](https://github.com/Shopify/semian/pull/71#discussion_r46702782)

Thoughts? @Sirupsen

### DIFF
--- a/lib/semian/net_http.rb
+++ b/lib/semian/net_http.rb
@@ -22,12 +22,14 @@ module Semian
     ResourceBusyError = ::Net::ResourceBusyError
     CircuitOpenError = ::Net::CircuitOpenError
 
-    def semian_configuration
-      Semian::NetHTTP.retrieve_semian_configuration(address, port)
+    class SemianConfigurationChangedError < RuntimeError
+      def initialize(msg = "Cannot re-initialize semian_configuration")
+        super
+      end
     end
 
     def semian_identifier
-      "nethttp_#{semian_configuration[:name]}"
+      "nethttp_#{raw_semian_options[:name]}"
     end
 
     DEFAULT_ERRORS = [
@@ -52,8 +54,13 @@ module Semian
     end
 
     class << self
-      attr_accessor :semian_configuration
       attr_accessor :exceptions
+      attr_reader :semian_configuration
+
+      def semian_configuration=(configuration)
+        raise Semian::NetHTTP::SemianConfigurationChangedError unless @semian_configuration.nil?
+        @semian_configuration = configuration
+      end
 
       def retrieve_semian_configuration(host, port)
         @semian_configuration.call(host, port) if @semian_configuration.respond_to?(:call)
@@ -67,9 +74,10 @@ module Semian
     Semian::NetHTTP.reset_exceptions
 
     def raw_semian_options
-      options = semian_configuration
-      options = options.dup unless options.nil?
-      options
+      @raw_semian_options ||= begin
+        @raw_semian_options = Semian::NetHTTP.retrieve_semian_configuration(address, port)
+        @raw_semian_options = @raw_semian_options.dup unless @raw_semian_options.nil?
+      end
     end
 
     def resource_exceptions
@@ -77,7 +85,7 @@ module Semian
     end
 
     def disabled?
-      semian_configuration.nil?
+      raw_semian_options.nil?
     end
 
     def connect
--- a/test/net_http_test.rb
+++ b/test/net_http_test.rb
@@ -192,6 +192,17 @@ class TestNetHTTP < MiniTest::Unit::TestCase
     end
   end
 
+  def test_custom_raw_semian_options_can_only_assign_once
+    semian_configuration_proc = proc do |host, port|
+      DEFAULT_SEMIAN_OPTIONS.merge(name: "#{host}_#{port}")
+    end
+    with_semian_configuration(semian_configuration_proc) do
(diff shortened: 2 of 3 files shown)


lib/semian/net_http.rb
```diff
@@ -67,9 +76,11 @@ def reset_exceptions
...
 
     def raw_semian_options
-      options = semian_configuration
-      options = options.dup unless options.nil?
-      options
+      check_semian_configuration_changed
+      @raw_semian_options ||= begin
+        @raw_semian_options = Semian::NetHTTP.retrieve_semian_configuration(address, port)
+        @raw_semian_options = @raw_semian_options.dup unless @raw_semian_options.nil?
+      end
```
Sirupsen: I don't understand why we can justify caching these values at all, since the options can change due to change to the function that could use time, random, .. or something else to spit out different values at different times:

```ruby
Semian::NetHTTP.semian_configuration = proc { |host, port|
  if Time.now.hour > 8 && Time.now.hour < 18 # during peak hours
    SEMIAN_PARAMETERS.merge(tickets: 10, name: "#{host}_#{port}")
  else # during the night
   SEMIAN_PARAMETERS.merge(tickets: 5, name: "#{host}_#{port}")
  end
}
```

Why do we need to cache it in the first place?

kyewei: As I said above in the PR description, it is pretty difficult to keep the inner resource updated with the different values ... that I think it's not worth it to do so. 

So then I took a chance of stance and said that the function should produce the same value each time. With that it mind then it makes sense to cache the values.

Sirupsen: My question is, why cache in the first place?

kyewei: With the assumption in mind, the value you get from calling the closure doesn't change, so cache to eliminate the closure call overhead.

Sirupsen: My point is that I think the performance gain is negligible, for less predictability and what I'd essentially call an intentional bug.

kyewei: But there is no intentional bug? What's the bug? The requirement of having the `proc` return the same value each time isn't a limitation ... if you don't do that and somehow make the `Resource` object call the `proc` each time it wants a `ticket` size, a `error_threshold` or any of the other parameters, it will lead to race conditions.

kyewei: I think my main point here is that it is awkward to argue for resetting the cached `@resource` based on condition `@configuration_id != Semian::NetHTTP.semian_configuration.object_id` but not have a similar cache for `@raw_semian_options`

Sirupsen: The bug is that subsequent invocations of a function can change their return value. That's the expected behaviour. Anything else is magic.

kyewei: (This issue was discussed IRL, it was decided that we make `semian_configuration` only allow initialization once. Otherwise checking whether the `proc` was return different values would either (1) have a `Hash` check overhead, not good, or (2) inconsistent behaviour between assigning a `proc` that gives different values at runtime and booting different workers)

kyewei:
@Sirupsen I got rid of the `test_custom_configuration_invalidates_old_resource` because that isn't needed anymore, and replaced it with `test_custom_raw_semian_options_can_only_assign_once`. 

I also fixed `with_semian_configuration` to bypass the 'assign only once' limitation for testing purposes and (since reassigning the `proc` may change the state) restore the state of any `Net::HTTP` instances affected by a changed `proc`

test/net_http_test.rb
```diff
@@ -346,10 +357,35 @@ def test_persistent_state_after_server_restart
 
   def with_semian_configuration(options = DEFAULT_SEMIAN_CONFIGURATION)
     orig_semian_options = Semian::NetHTTP.semian_configuration
+    Semian::NetHTTP.class_eval "@semian_configuration = nil"
```
Sirupsen: `instance_variable_set(:"@semian_configuration", nil)`

Sirupsen:
Go for it

kyewei:
Rebased and squashed

PR #72, opened 2016-01-19
Rescue DNS errors

fw42:
We had problems in the past where due to a network outage in our datacenter, Redis clients would randomly show DNS resolution errors for their hostnames. Semian doesn't rescue those. The problem gets even more shitty since Redis and HiRedis handle those cases differently (see https://github.com/redis/hiredis-rb/issues/40).

This is a super quick-and-dirty fix. I'm sure there is a better way, but I want to get the discussion going.

Is this useful?

@byroot @csfrancis @Sirupsen

### DIFF
--- a/lib/semian/redis.rb
+++ b/lib/semian/redis.rb
@@ -30,6 +30,17 @@ end
 
 module Semian
   module Redis
+    module DnsResolutionErrorPatch
+      def raw_connect
+        super
+      rescue ::SocketError, RuntimeError => exception
+        raise unless exception.message =~ /Name or service not known/
+        new_exception = ::Redis::BaseConnectionError.new(exception.message)
+        new_exception.set_backtrace(exception.backtrace)
+        raise new_exception
+      end
+    end
+
     include Semian::Adapter
 
     ResourceBusyError = ::Redis::ResourceBusyError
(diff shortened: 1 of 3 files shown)


fw42:
What I don't like about this implementation is that it changes the behaviour of the redis/hiredis gem, which might have unexpected side-effects in the application that uses Semian and depends on this behaviour. The alternative would be to just rescue (and re-raise) all SocketError and RuntimeError. I'm fine with doing that for SocketError, but blindly rescueing ALL RuntimeError that happen in the driver seems kinda dangerous to me.

Alternatively, I could rescue the RuntimeError, check for the DNS message, reraise a SocketError, and then make semian rescue the SocketError (as part of `resource_exceptions`).

Thoughts?

byroot:
We already handle all of those concerns in Semian, see https://github.com/Shopify/semian/blob/834d556a5c54bf372237123c97b0506105ff5819/lib/semian/redis.rb#L7 for example.

When you add `AdapterError` to an exception class, it adds a `semian_identifier` attribute to it. When the exception is raised under a cricuit breaker, the name of the resource is added.

And then in `with_fallback`, we ignore then error unless the semian_identifier matches our expectations: https://github.com/Shopify/shopify/blob/04aea13307920bef4f2a7f4ea785710652fdb4e8/lib/resiliency.rb#L12

So no need to convert the exception, and the fix should be in semian itself.

byroot:
IMO this https://github.com/Shopify/semian/pull/73 should be the implementation.

fw42:
Closing in favor of https://github.com/Shopify/semian/pull/73

PR #73, opened 2016-01-19
Handle Hiredis silly RuntimeError on DNS resolution issues

byroot:
WIP, need to add tests.

cc @fw42

### DIFF
--- a/test/redis_test.rb
+++ b/test/redis_test.rb
@@ -52,7 +52,7 @@ class TestRedis < MiniTest::Unit::TestCase
   end
 
   def test_dns_resolution_errors_in_ruby_driver_open_the_circuit
-    redis_options = { host: "invalid-dns-name", semian: SEMIAN_OPTIONS, reconnect_attempts: 0 }
+    redis_options = {host: "invalid-dns-name", semian: SEMIAN_OPTIONS, reconnect_attempts: 0}
 
     clients = [
       Redis.new(driver: 'ruby', **redis_options),
--- a/test/redis_test.rb
+++ b/test/redis_test.rb
@@ -51,6 +51,29 @@ class TestRedis < MiniTest::Unit::TestCase
     end
   end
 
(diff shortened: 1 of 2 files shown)


fw42:
Added a test and removed the special case for hiredis. This is stupid. We should fix that one upstream (in hiredis-rb). Will look into it. For now, this PR solves the problem for the ruby redis driver, which I think is an improvement.

@byroot, cool?

Sirupsen:
LGTM

test/redis_test.rb
```diff
@@ -51,6 +51,29 @@ def test_connection_errors_open_the_circuit
     end
   end
 
+  def test_dns_resolution_errors_in_ruby_driver_open_the_circuit
+    redis_options = { host: "invalid-dns-name", semian: SEMIAN_OPTIONS, reconnect_attempts: 0 }
+
+    clients = [
+      Redis.new(driver: 'ruby', **redis_options),
+      # Redis.new(driver: 'hiredis', **redis_options)
```
byroot: So you don't want to test hiredis? Even though I agree it should be fixed upstream, we have no idea what the ETA might be.

byroot:
Not sure about the ignoring hiredis situation. But I'll let you decide.

Regular Redis LGTM.

:shipit:

test/redis_test.rb
fw42: well the test is gonna fail :) and I don't like that ugly hack where we look at the message :/

byroot:
Oh, you have a rubocop violation :/ Sorry.

fw42:
Do you remember why this never shipped?

fw42:
Scott is trying to do something similar in https://github.com/Shopify/semian/pull/184. Which one should we ship?

casperisfine:
Sorry can't remember why this one fall through the cracks.

Scott's look cleaner anyway. Closing.

fw42:
The test here (assert the circuit was opened) seems useful, maybe we should port that to the other PR?

PR #74, opened 2016-01-22
NetHTTP response code handling

j-mutter:
Getting an unwanted response code, say, a `500`, doesn't actually raise anything in Net::HTTP, and therefore doesn't trigger anything for Semian to detect.

This provides that via an optional `raise_on` config key :)

@Sirupsen @byroot - can you guys take a peek please?
cc @Mavvie - relevant to your interests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@
 /html/
 Gemfile.lock
 vendor/
+
+test/test_config.yml
--- a/lib/semian/net_http.rb
+++ b/lib/semian/net_http.rb
@@ -3,6 +3,7 @@ require 'semian/adapter'
 
 module Net
   ProtocolError.include(::Semian::AdapterError)
+  HTTPBadResponse.include(::Semian::AdapterError)
 
   class SemianError < ::Net::ProtocolError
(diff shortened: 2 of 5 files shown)


byroot:
I'm on the fence about this.

  - `HTTPBadResponse` is an existing exception class that already has a meaning etc, I think it's a bad idea to re-use it
  - This totally change Net::HTTP behavior, many libraries use it, so you might break a bunch of libraries with this.

j-mutter:
Yeah, I thought about making a new exception class, so that works for me. And this is a totally optional opt-in feature that you can scope to specific domains and ports. It should be transparent to anyone who doesn't want to use it

byroot:
Well, if you create a new exception class (that can inherit from Net::HTTP::ProtocolError), then yeah maybe.

Sorry I don't want to be dismissive, it's just that there is many many things to consider before I can say if it has a chance to work. So I think we should try that out for a while before actually merging it.

j-mutter:
Alright, going to do this a little higher up the food chain in HTTParty as it already has an opt-in `raise_on` API. I'll close this for now

PR #75, opened 2016-03-08
Automatically convert T_FLOAT ticket count

csfrancis:
@Sirupsen, @byroot 

This will automatically convert floating point ticket counts to integer values instead of raising `TypeError`.

### DIFF
--- a/test/resource_test.rb
+++ b/test/resource_test.rb
@@ -26,6 +26,12 @@ class TestResource < MiniTest::Unit::TestCase
     end
   end
 
+  def test_initialize_with_float
+    resource = create_resource :testing, tickets: 1.0
+    assert resource
+    assert_equal 1, resource.tickets
+  end
+
   def test_max_tickets
     assert Semian::MAX_TICKETS > 0
   end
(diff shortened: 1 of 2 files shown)


test/resource_test.rb
```diff
@@ -26,6 +26,12 @@ def test_initialize_invalid_args
     end
   end
 
+  def test_initialize_with_float
+    r = create_resource :testing, tickets: 1.0
+    assert r
+    assert r.tickets == 1
```
fw42: `assert_equal`?

fw42:
:+1:

test/resource_test.rb
```diff
@@ -26,6 +26,12 @@ def test_initialize_invalid_args
     end
   end
 
+  def test_initialize_with_float
+    r = create_resource :testing, tickets: 1.0
+    assert r
```
byroot: single letter variables makes me :crying_cat_face:

byroot:
Agree with Flo on the `assert_equal`.

Other than that :+1:

PR #78, opened 2016-03-22
Remove deprecated TimeoutError const

createdbypete:
Ruby 2.3.0 has deprecated the `TimeoutError` const causing Ruby to issue some warnings.

```
lib/semian/net_http.rb:37: warning: constant ::TimeoutError is deprecated
```

Since this was listed as an alias for `Timeout::Error` seems safe(?) to propose it's removal.

Added Ruby 2.3.0 to the Travis builds to confirm.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ before_install:
 
 rvm:
   - 2.1.1
+  - 2.3.0
 
 services:
   - redis-server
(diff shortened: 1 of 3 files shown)


byroot:
```ruby
>> TimeoutError
=> Timeout::Error
```

Yup, that was silly. Will merge on green. Thanks a lot!

byroot:
Look like the tests don't run on 2.3. I'll take a look tomorrow.

createdbypete:
@byroot we're back in green! Was simply a Minitest inconsistency between the Ruby versions. Resolved by including the gem.

byroot:
Thanks again for looking into it.

byroot:
Released as `0.4.2`.

PR #80, opened 2016-05-09
Fix lazy aliasing of Redis#semian_resource

casperispoop:
### Context

The redis client is a bit special because:
  - `Redis.new` returns some kind of wrapper.
  - The "real" client is in `Redis.new.client`
  - So we have an delegator on `Redis#semian_resource` to `Redis::Client#semian_resource` to respect's semian contract.

The problem is that the redis "wrapper", during `pipelined` and similar block calls, replaces the `Redis::Client` instance by a `Redis::Pipeline` instance, and there is no way to access the real client anymore because it's just in a local variable. See: https://github.com/redis/redis-rb/blob/fd77358708920b3b7b76955eda22e8c89647aae5/lib/redis.rb#L2248-L2258

### Problem

We knew about this error, I even put a descriptive comment about why we were memoizing the resource.

Where I dropped the ball is that if the first call ever to `semian_resource` is from inside a `pipelined` block, then you end up with

```ruby
NoMethodError·undefined method `semian_resource' for #<Redis::Pipeline:0x005631a026c398>
```

Any of @fw42 @pushrax @Sirupsen @csfrancis for review please.

### DIFF
--- a/lib/semian/redis.rb
+++ b/lib/semian/redis.rb
@@ -16,11 +16,17 @@ class Redis
   ResourceBusyError = Class.new(SemianError)
   CircuitOpenError = Class.new(SemianError)
 
-  # This memoized alias is necessary because during a `pipelined` block
-  # the client is replaced by an instance of `Redis::Pipeline` and there is
-  # no way to access the original client.
-  def semian_resource
-    @semian_resource ||= @client.semian_resource
+  attr_reader :semian_resource
+
+  alias_method :_original_initialize, :initialize
+
+  def initialize(*args, &block)
+    _original_initialize(*args, &block)
+
+    # This alias is necessary because during a `pipelined` block
+    # the client is replaced by an instance of `Redis::Pipeline` and there is
+    # no way to access the original client.
+    @semian_resource = client.semian_resource
   end
 
   def semian_identifier
--- a/test/redis_test.rb
+++ b/test/redis_test.rb
@@ -19,9 +19,9 @@ class TestRedis < MiniTest::Unit::TestCase
   end
 
   def test_semian_identifier
-    assert_equal :redis_foo, Redis.new(semian: {name: 'foo'}).client.semian_identifier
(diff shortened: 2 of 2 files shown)


Sirupsen:
Subtle. LGTM

Did you find this in production?

byroot:
Nope ci, but it could have happened in prod

lib/semian/redis.rb
```diff
@@ -16,11 +16,17 @@ def initialize(semian_identifier, *args)
...
-  # the client is replaced by an instance of `Redis::Pipeline` and there is
-  # no way to access the original client.
-  def semian_resource
-    @semian_resource ||= @client.semian_resource
+  attr_reader :semian_resource
+
+  alias_method :_original_initialize, :initialize
+
+  def initialize(*args, &block)
+    _original_initialize(*args, &block)
```
fw42: blergh, this is super ugly. Can't we `prepend`?

test/redis_test.rb
```diff
@@ -19,9 +19,9 @@ def setup
   end
 
   def test_semian_identifier
-    assert_equal :redis_foo, Redis.new(semian: {name: 'foo'}).client.semian_identifier
-    assert_equal :'redis_127.0.0.1:6379/0', Redis.new.client.semian_identifier
-    assert_equal :'redis_example.com:42/0', Redis.new(host: 'example.com', port: 42).client.semian_identifier
```
fw42: are those changes necessary for this test to pass? not sure I understand

casperispoop: The semian resource used to be lazily accessed, so this test was able to work without providing semian config.

Now that it's eagerly accessed, it's not possible anymore, so I had to fix that test.

lib/semian/redis.rb
casperispoop: I sticked with the file stile. I'm not 100% sure why we didn't use prepend in semian. IIRC it's because prepeding a method defined in C and calling super used to make MRI segfault.

I'd be happy to convert semian to prepend, but not in this PR though.

fw42:
👍

PR #81, opened 2016-05-10
Convert to prepend

casperispoop:
@fw42 as you requested.

cc @Sirupsen


casperispoop:
Oh actually this doesn't make sense. We alias the methods as `raw_<original>` to you can wrap them with things like `statsd-instrument`. If we use `super`, then your decorator won't be called...

So yeah closing.

PR #82, opened 2016-05-11
Only issue unsupported or disabled semaphores warnings when the first resource is instanciated

casperispoop:
That will allow users to set the logger level and set another logger.


Sirupsen:
LGTM

lib/semian/platform.rb
```diff
@@ -1,6 +1,16 @@
 module Semian
+  extend self
```
kmcphillips: 👌

lib/semian/protected_resource.rb
```diff
@@ -1,5 +1,3 @@
-require 'forwardable'
```
kmcphillips: 👌

kmcphillips:
This is tops. Thanks for taking ownership of this so quickly.

Issue #86, opened 2016-08-09
Possible concurrency issues


eapache:
I was poking around and found what I think are a few (hypothetical) issues when some semian methods are called concurrently. It's possible I'm missing something in the architecture (e.g are adapter methods expected to always be mutexed by the parent driver the way redis is?) but if not then here are some things I noticed:

- multiple concurrent calls to a new adapter can end up racing in `retrieve_or_register` creating multiple instances of the same circuit-breaker and protected resource
- various issues in the circuit-breaker, e.g. concurrent calls to `request_allowed?` can "lose" successful responses by triggering multiple transitions to the half-open state

casperisfine:
> are adapter methods expected to always be mutexed by the parent driver the way redis is

Yes. I can't think of any datastore driver that would support being used concurrently from multiple threads.

eapache:
I am thinking e.g. of mysql, where you have a connection pool; the individual connections are synchronous but the pool itself is concurrent. I was assuming the driver as a whole got a single semian resource, not one per connection?

casperisfine:
> I was assuming the driver as a whole got a single semian resource, not one per connection?

It's one per connection. MySQL and other relational DB drivers don't handle pooling, you have to handle pooling higher in the stack.

If one driver were to implement pooling, the semian adapter would have to take care of this concern indeed.

eapache:
OK, adapter mutexing takes care of the first issue (`retrieve_or_register`) but I believe the circuit-breaker point is still valid since `request_allowed?` and other such methods are expected to be publicly called outside the scope of the adapter.

casperisfine:
> request_allowed? and other such methods are expected to be publicly called outside the scope of the adapter

By whom? I can't see a reason to call those methods except to simulate a test example, in which case threading isn't an issue.

casperisfine:
Ok, so as discussed offline. `request_allowed?` can make sense to be alled higher on the start to be able to bail out of a code path early if you know it depends on a datastore.

Because of this it should be thread safe. I'm working on a fix as we speak.

Sirupsen:
Yeah, it was not designed to be `thread_safe` (we haven't needed it), but it definitely makes sense for it to be.

👍  to @casperisfine's direction.

PR #87, opened 2016-08-09
Make request_allowed? thread safe

casperisfine:
First draft for https://github.com/Shopify/semian/issues/86

The idea here is to remove any side effect `request_allowed?` might have, so that it's safe to call it from a threaded environment.

NB: The other circuit breaker methods stays unsafe and it's the adapter's responsibility to call them safely.

### DIFF
--- a/lib/semian/circuit_breaker.rb
+++ b/lib/semian/circuit_breaker.rb
@@ -2,6 +2,8 @@ module Semian
   class CircuitBreaker #:nodoc:
     extend Forwardable
 
+    def_delegators :@state, :closed?, :open?, :half_open?
+
     def initialize(name, exceptions:, success_threshold:, error_threshold:, error_timeout:, implementation:)
       @name = name.to_sym
       @success_count_threshold = success_threshold
@@ -15,6 +17,8 @@ module Semian
     end
 
     def acquire
+      half_open if open? && error_timeout_expired?
+
       raise OpenCircuitError unless request_allowed?
 
       result = nil
@@ -30,9 +34,10 @@ module Semian
     end
 
     def request_allowed?
-      return true if closed?
-      half_open if error_timeout_expired?
-      !open?
+      closed? ||
+        half_open? ||
+        # The circuit breaker is officially open, but it will transition to half-open on the next attempt.
+        (open? && error_timeout_expired?)
     end
 
     def mark_failed(_error)
@@ -63,9 +68,6 @@ module Semian
     end
 
     private
-
-    def_delegators :@state, :closed?, :open?, :half_open?
-    private :closed?, :open?, :half_open?
 
     def close
       log_state_transition(:closed)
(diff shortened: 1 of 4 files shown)


lib/semian/circuit_breaker.rb
```diff
@@ -30,9 +32,7 @@ def acquire
     end
 
     def request_allowed?
-      return true if closed?
-      half_open if error_timeout_expired?
-      !open?
+      closed? || half_open? || (open? && error_timeout_expired?)
```
dylanahsmith: `error_timeout_expired?` uses `@errors.last` which is calling `.last` on an array.  That is probably safe to do in ruby due to the vm lock, but we might want to have a `@last` instance variable in the SlidingWindow class if you want it to be accessed in a thread-safe way for jruby.

dylanahsmith:
Seems like the right approach.

It just looks like CI isn't setup correctly anymore, since it is getting connect errors and retrying a CI run on master produced errors now when it was passing 29 days ago.

lib/semian/circuit_breaker.rb
casperisfine: Indeed. But JRuby isn't really a concern at this point since we rely on C extensions.

casperisfine:
Ok, CI is green now (thx Dylan). 

@dylanahsmith @csfrancis @Sirupsen @eapache mind reviewing?

dylanahsmith:
:+1:

lib/semian/circuit_breaker.rb
eapache: Might want to leave a comment that `open? && error_timeout_expired?` is intended as a non-mutating equivalent check to "should transition to half-open"?

Or should that condition even just go in `half_open?` to begin with?

eapache:
👍

Sirupsen:
👍

dylanahsmith:
This looks like it could be tested by using `request_allowed?` in a circuit breaker test and asserting that it doesn't change the state when the error timeout expires.

lib/semian/circuit_breaker.rb
casperisfine: > Or should that condition even just go in half_open? to begin with?

It can't really go in `half_open?` because then you wouldn't know if you have to make a transition or not. So yeah that piece isn't necessarily trivial to understand. I'm usually not a comment guy but it's likely worth it here.

dylanahsmith:
It looks like only UnprotectedResource#request_allowed? is tested in the test suite.  It doesn't look like there are any tests for CircuitBreaker#request_allowed? or ProtectedResource#request_allowed? which delegates to the circuit breaker.

casperisfine:
I added a test, but since the state query methods were all private I had to publicize them. 

And yes CircuitBreaker is mostly tested in an integration way.

dylanahsmith:
I guess thread safety is hard to test in an integration way.  Thanks for adding the test though

casperisfine:
Released as `0.5.2`

PR #88, opened 2016-08-09
Fix CI connect errors on HTTP requests by using 127.0.0.1 for host

dylanahsmith:
@casperisfine for review

After seeing CI failures on https://github.com/Shopify/semian/pull/87 which didn't seem relevant to that PR, I retried CI on master (https://travis-ci.org/Shopify/semian/jobs/143832816) and it failed the same way.

It looks like `TCPSocket.new(hostname, port)` was raising an exception when hostname is "localhost", so tried "127.0.0.1" and it seems to have fixed the error.

### DIFF
--- a/test/net_http_test.rb
+++ b/test/net_http_test.rb
@@ -11,7 +11,7 @@ class TestNetHTTP < Minitest::Test
     end
   end
 
-  HOSTNAME = "localhost"
+  HOSTNAME = "127.0.0.1"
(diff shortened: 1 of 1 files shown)


Sirupsen:
LGTM

It may have been binding to `localhost` in Vagrant, but on OS X it only binds to 127.0.0.1 IIRC

byroot:
👍 :shipit:

PR #89, opened 2016-08-17
Change to allow tickets: false to not have bulkhead required

BoGs:
Change to allow `tickets: false` to give the functionality of circuit breaking but not bulk heading using semaphores. 

@Sirupsen and @csfrancis please take a look and let me know what you think. Since tickets can be `Integer` or `Boolean` using `!!` seemed best - thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -407,6 +407,15 @@ queried frequently has worked well for us.
   We recommend this to be `0` unless you have very few workers running (i.e.
   less than ~5).
 
+Note that there are system-wide limitations on how many tickets can be allocated
+on a system. `cat /proc/sys/kernel/sem` will tell you.
+
+> System-wide limit on the number of semaphore sets.  On Linux
+  systems before version 3.19, the default value for this limit
+  was 128.  Since Linux 3.19, the default value is 32,000.  On
+  Linux, this limit can be read and modified via the fourth
+  field of `/proc/sys/kernel/sem`.
+
 ## Defense line
 
 The finished defense line for resource access with circuit breakers and


lib/semian/protected_resource.rb
```diff
@@ -12,15 +12,19 @@ def initialize(resource, circuit_breaker)
     end
 
     def destroy
-      @resource.destroy
+      @resource.destroy if !!tickets
```
Sirupsen: No need for double-bang, just do `if tickets`. It'll be `falsy` if there are any. Even if there's `0`, we'll allocate things.

lib/semian/protected_resource.rb
```diff
@@ -12,15 +12,19 @@ def initialize(resource, circuit_breaker)
...
       @circuit_breaker.destroy
     end
 
     def acquire(timeout: nil, scope: nil, adapter: nil)
       @circuit_breaker.acquire do
         begin
-          @resource.acquire(timeout: timeout) do
-            Semian.notify(:success, self, scope, adapter)
+          if !!tickets
+            @resource.acquire(timeout: timeout) do
```
Sirupsen: Why don't you do the `tickets` check inside of the resource?

test/net_http_test.rb
```diff
@@ -105,6 +105,49 @@ def test_get_is_protected
     end
   end
 
+  def test_get_with_no_bulkhead
```
Sirupsen: I'm not sure what this test really tests.

BoGs: this tests that a success response happens if the ticket count is `false` - given the tickets get passed to the resource - which was crashing because not initialized before hand. Was trying not to initialize semaphore and skip acquiring it.

Sirupsen: OK

Sirupsen: Is there an inspect API in Semian where you can see that no tickets are acquired and no resource was created perhaps?

Sirupsen:
What about the creation of the semaphore set? I think that happens in the initialization code somewhere, that would also have to be disabled.

Sirupsen:
Bump

PR #90, opened 2016-08-27
mysql: circuit break pings

Sirupsen:
Currently pings are not behind circuit breakers. They respect the default `read_timeout`, **which means that in stock Rails each request will take at least `read_timeout`, even if the data store is behind a circuit breaker**. This is because a connection is always pinged when it's checked out of the pool, see [`ActiveRecord::ConnectionPool#checkout_and_verify`](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb#L776) and [`ActiveRecord::AbstractAdapter#verify!`](https://github.com/rails/rails/blob/4-2-stable/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L339).

Tagging Shopify people who've dabbled in AR:

@dylanahsmith @csfrancis @rafaelfranca @casperisfine

### DIFF
--- a/lib/semian/mysql2.rb
+++ b/lib/semian/mysql2.rb
@@ -27,6 +27,7 @@ module Semian
 
     ResourceBusyError = ::Mysql2::ResourceBusyError
     CircuitOpenError = ::Mysql2::CircuitOpenError
+    PingFailure = Class.new(::Mysql2::Error)
 
     DEFAULT_HOST = 'localhost'
     DEFAULT_PORT = 3306
@@ -44,6 +45,9 @@ module Semian
 
       base.send(:alias_method, :raw_connect, :connect)
       base.send(:remove_method, :connect)
+
+      base.send(:alias_method, :raw_ping, :ping)
+      base.send(:remove_method, :ping)
     end
 
     def semian_identifier
@@ -55,6 +59,17 @@ module Semian
         end
         :"mysql_#{name}"
       end
+    end
+
+    def ping
+      result = nil
+      acquire_semian_resource(adapter: :mysql, scope: :ping) do
+        result = raw_ping
+        raise PingFailure.new(result.to_s) unless result
+      end
+      result
+    rescue ResourceBusyError, CircuitOpenError, PingFailure
+      false
     end
 
     def query(*args)
@@ -84,7 +99,7 @@ module Semian
     def acquire_semian_resource(*)
       super
     rescue ::Mysql2::Error => error
-      if error.message =~ CONNECTION_ERROR
+      if error.message =~ CONNECTION_ERROR || error.is_a?(PingFailure)
         semian_resource.mark_failed(error)
         error.semian_identifier = semian_identifier
       end
--- a/lib/semian/mysql2.rb
+++ b/lib/semian/mysql2.rb
@@ -21,7 +21,7 @@ module Semian
 
     CONNECTION_ERROR = Regexp.union(
       /Can't connect to MySQL server on/i,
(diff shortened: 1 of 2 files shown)


lib/semian/mysql2.rb
```diff
@@ -21,12 +21,13 @@ module Mysql2
...
     CONNECTION_ERROR = Regexp.union(
       /Can't connect to MySQL server on/i,
-      /Lost connection to MySQL server during query/i,
+      /Lost connection to MySQL server/i,
       /MySQL server has gone away/i,
     )
 
     ResourceBusyError = ::Mysql2::ResourceBusyError
     CircuitOpenError = ::Mysql2::CircuitOpenError
+    PingFailure = ::Mysql2::Error
```
casperisfine: I'd so a subclass here: `PingFailure = Class.new(::Mysql2::Error)`

lib/semian/mysql2.rb
```diff
@@ -84,7 +99,7 @@ def connect(*args)
     def acquire_semian_resource(*)
       super
     rescue ::Mysql2::Error => error
-      if error.message =~ CONNECTION_ERROR
+      if error.message =~ CONNECTION_ERROR || error.class == PingFailure
```
casperisfine: Related to the previous comment, if you don't make `PingFailure` a subclass, this condition will always be true.

Also I'd use `is_a?` since it's way more idiomatic.

casperisfine:
The CI failures seem legit.

Sirupsen:
> The CI failures seem legit.

It wasn't failing locally. Strange. I'll take a look.

lib/semian/mysql2.rb
```diff
@@ -57,6 +61,17 @@ def semian_identifier
...
 
+    def ping
+      result = nil
+      acquire_semian_resource(adapter: :mysql, scope: :ping) do
+        result = raw_ping
+        raise PingFailure.new(result.to_s) unless result
+      end
+      result
+    rescue ResourceBusyError, CircuitOpenError, PingFailure
+      false
```
casperisfine: Ok, so I could probably answer this myself if I checked (which I might do a little later), but I think there is a possible important concern here.

If `ping` is used by connection the pool to check if a connection is healthy, and create a new one if it isn't. We should make sure failing pings when the circuit is open won't make AR's connection pool do crazy things, and generate tons of connections attempts. (I'm pretty sure the connection is also protected by the circuit breaker, but I prefer to be on the safe side)

Sirupsen: Since establishing a new connection is also behind a circuit, do you think this is problematic? In vanilla AR, what would happen is that the connection is checked out of the pool, on `verify!` `#ping` returns false and a new connection is established. This would raise right away. Given that Rails doesn't check out the connection by default anymore in 4.2 (only when it's required), this seems sensible.

Internally, we only reconnect on `#execute` (.. and `#quote`), and we'd only do so when a `#ping` fails (when it's circuit broken), in which case it'd often just raise straight away.

I can't think of a case where things would go nuts in AR, and given that this doesn't change behaviour but just enforces it more strictly and faster during certain conditions, I'm not super concerned. Is there anything in particular you'd like me to investigate?

casperisfine: Nope, sounds good! Thanks for thinking about it :)

csfrancis:
👍 LGTM!

dylanahsmith:
LGTM

rafaelfranca:
:+1:

PR #91, opened 2016-09-02
Also let "Too many connections" be a first class conn error

methodmissing:
Observed that this isn't handled by the adapter when lowering MySQL `max_connections` counts during testing.

@Sirupsen @camilo @csfrancis @byroot

### DIFF
--- a/lib/semian/mysql2.rb
+++ b/lib/semian/mysql2.rb
@@ -23,6 +23,7 @@ module Semian
       /Can't connect to MySQL server on/i,
       /Lost connection to MySQL server/i,
       /MySQL server has gone away/i,
+      /Too many connections/i,
     )
 
     ResourceBusyError = ::Mysql2::ResourceBusyError
--- a/lib/semian/version.rb
+++ b/lib/semian/version.rb
@@ -1,3 +1,3 @@
 module Semian
-  VERSION = '0.5.3'
+  VERSION = '0.5.4'
 end


Sirupsen:
Ugh, thanks @methodmissing.. I also ran into another one the other day. Should we consider calling `#ping` instead to see if the connection is alive (if it's not, it must be a connection error, if it is, it must be an execution error)? That may be much more reliable than this string parsing.

csfrancis:
👍

byroot:
:shipit:

methodmissing:
I'm on the fence with changing the behaviour to a ping check as it might introduce new edges, especially in production whereas string parsing has been through a great deal of runtime.

I agree it's not perfect as there's a catch up game to play, but still workable given general MySQL error messages and codes has been mostly static since the 3.x releases (libmysqlclient specifically) and the fact that we basically would only ever add to the set and not modify it's contents means:

* No fiddling with know good production cases as there's no modification to the list elements directly
* It's OK to add an edge to the tail and handle a new error state

Thoughts?

byroot:
Agree with what you said. Also since https://github.com/Shopify/semian/pull/90 `ping` is behind the circuit breaker as well.

methodmissing:
@Sirupsen if that's good with you, can we cut a new release?

Sirupsen:
@methodmissing go ahead with merging and bumping the release. You can push the gem with ShipIt once you've bummed the version file.

camilo:
👍 

on a side note we should see if we can change the adapter at some point to avoid that ungodly regexp matching.

Sirupsen:
> on a side note we should see if we can change the adapter at some point to avoid that ungodly regexp matching.

Suggestions?

camilo:
> Suggestions?

Naively we can patch (upstream) `mysql2` to have typed exceptions based on the err code so something like`Too many connections/` would be `Mysql::Error::TooManyConnections`.

That way if the string changes for some reason mysql2 would translate to the right exception. Or we could also use the errornumber/sql state instead of the regexp.

TBH I don't think is the hugest deal, just in general pattern matching exceptions messages is  a bit of a 👃

Sirupsen:
It's the same problem as in `Net::HTTP`, where we need a [list of all the exceptions](https://github.com/Shopify/semian/blob/master/lib/semian/net_http.rb#L35). Given that we've run this for so long without problems, I think this is OK and not worth spending much engineering on.

PR #92, opened 2016-10-17
Implement AR::AbstractAdapter#semian_resource

casperisfine:
Allows for things like:

```ruby
if ActiveRecord::Base.connection.semian_resource.request_allowed?
  # ...
end
```

It's easily done with most semian adapters, but if you use Rails, the actual MySQL2 client is hard to get access to.

@fw42 @Sirupsen

### DIFF
--- a/lib/semian/rails.rb
+++ b/lib/semian/rails.rb
@@ -1,0 +1,7 @@
+require 'active_record/connection_adapters/abstract_adapter'
+
+class ActiveRecord::ConnectionAdapters::AbstractAdapter
+  def semian_resource
+    @connection.semian_resource
+  end
+end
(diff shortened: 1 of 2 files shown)


Sirupsen:
I've wanted this before, 👍

lib/semian/rails.rb
```diff
@@ -0,0 +1,7 @@
+require 'active_record/connection_adapters/abstract_adapter'
+
+class ActiveRecord::ConnectionAdapters::AbstractAdapter
```
Sirupsen: Could we just automagically include this from the MySQL module if `ActiveRecord` is defined? Instead of manually having to require this?

dylanahsmith: Won't that make it order dependant such that it won't work if semiannual is loaded first by bundler

casperisfine: No because if you are in the context of a Rails app, the `Bundler.require` is triggered by Rails itself early it's boot process, so effectively when semian is required ActiveSupport should be there.

PR #99, opened 2017-02-09
Switch to enumerated type for tracking semaphore indices

dalehamel:
# What

Switch from tracking our semaphore indices manually, with static const ints, to using an enumerated type.

# Why

In my opinion, this is a C best-practice as this is what enumerated types are for. I hope this doesn't come off as a bike shed, as I think that there are legitimate technical reasons for using enumerated types in this instance.

First off, I think that the code becomes more readable as the concept of an index is abstracted - you're able to give the name of the semaphore you expect, and the fact that it is an index happens implicitly because of how enumerated types work.

This also somewhat future proofs this, as there's no need to update the "kNumSemaphores" variable manually, the convention of using the last value of an enum as the count will allow us to always know how many semaphores are in the set.

WDYT?

### DIFF
--- a/ext/semian/semian.c
+++ b/ext/semian/semian.c
@@ -35,11 +35,13 @@ static ID id_timeout;
 static VALUE eSyscall, eTimeout, eInternal;
 static int system_max_semaphore_count;
 
-static const int kIndexTickets = 0;
-static const int kIndexTicketMax = 1;
-static const int kIndexLock = 2;
-
-static const int kNumSemaphores = 3;
+typedef enum
+{
+  SI_SEM_TICKETS,            // semaphore for the tickets currently issued
+  SI_SEM_CONFIGURED_TICKETS, // semaphore to track the desired number of tickets available for issue
+  SI_SEM_LOCK,               // metadata lock to act as a mutex, ensuring thread-safety for updating other semaphores
+  SI_NUM_SEMAPHORES          // always leave this as last entry for count to be accurate
+} semaphore_indices;
 
 typedef struct {
   int sem_id;
@@ -135,7 +137,7 @@ static const int kInternalTimeout = 5; /* seconds */
 static int
 get_max_tickets(int sem_id)
 {
-  int ret = semctl(sem_id, kIndexTicketMax, GETVAL);
+  int ret = semctl(sem_id, SI_SEM_CONFIGURED_TICKETS, GETVAL);
   if (ret == -1) {
     rb_raise(eInternal, "error getting max ticket count, errno: %d (%s)", errno, strerror(errno));
   }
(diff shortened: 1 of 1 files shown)


ext/semian/semian.c
```diff
@@ -35,11 +35,13 @@ static ID id_timeout;
...
 
-static const int kIndexTickets = 0;
-static const int kIndexTicketMax = 1;
-static const int kIndexLock = 2;
-
-static const int kNumSemaphores = 3;
+typedef enum
+{
+  sem_tickets,     // semaphore for the tickets currently issued
+  sem_max_tickets, // semaphore to track the maximum number of tickets available for issue
```
sirupsen: I would almost consider renaming this one to `sem_configured_tickets`. I see it more as the value to indicate what the amount of currently configured tickets are, so we can accurately adjust.

dalehamel: @sirupsen hadn't pinged you yet because still working on this, but yeah, my first question was going to be "what should we name these things"

sirupsen: Sorry, watching the repo so I saw it when you made the PR. 😬

dalehamel:
@sirupsen @csfrancis for review.

In particular, what do you think are the most appropriate, descriptive names we can use for these semaphores?

dalehamel:
FWIW CI seems to actually pass, but be flakey? Some upstream gem is failing to compile native extensions.

ext/semian/semian.c
```diff
@@ -35,11 +35,13 @@ static ID id_timeout;
...
 static int system_max_semaphore_count;
 
-static const int kIndexTickets = 0;
-static const int kIndexTicketMax = 1;
-static const int kIndexLock = 2;
-
-static const int kNumSemaphores = 3;
+typedef enum
+{
+  sem_tickets,     // semaphore for the tickets currently issued
```
csfrancis: Could you upper case the enum values?

ext/semian/semian.c
```diff
@@ -35,11 +35,13 @@ static ID id_timeout;
...
+  sem_configured_tickets, // semaphore to track the desired number of tickets available for issue
+  sem_lock,        // metadata lock to act as a mutex, ensuring thread-safety for updating other semaphores
+  NUM_SEMAPHORES   // always leave this as last entry for count to be accurate
+} semaphore_set;
```
csfrancis: `semaphore_indeces` ?

ext/semian/semian.c
csfrancis: And maybe prefix with `SI` (for `semaphore_indeces`) ?

dalehamel: @csfrancis like SISemTickets?

dalehamel: or SI_SEM_TICKETS?

ext/semian/semian.c
dalehamel: agreed

dalehamel:
@csfrancis names adjusted per your request

PR #102, opened 2017-02-12
Refactor semian.c includes and types into header files

dalehamel:
# What

Part of https://github.com/Shopify/semian/pull/101
- Refactor semian includes into semian.h
- Refactor semian types into semian_types.h

# Why

Splitting larger omnibus refactor into more reviewable PRs

### DIFF
--- a/ext/semian/semian.h
+++ b/ext/semian/semian.h
@@ -1,0 +1,30 @@
+/*
+System, 3rd party, and project includes
+
+Implements Init_semian, which is used as C/Ruby entrypoint.
+*/
+
+#ifndef SEMIAN_H
+#define SEMIAN_H
+
+// System includes
+#include <sys/types.h>
+#include <sys/ipc.h>
+#include <sys/sem.h>
+#include <sys/time.h>
+#include <errno.h>
+#include <string.h>
+#include <stdio.h>
+
+// 3rd party includes
+#include <openssl/sha.h>
+#include <ruby.h>
+#include <ruby/util.h>
+#include <ruby/io.h>
+
+//semian includes
+#include <semian_types.h>
+
+void Init_semian();
+
+#endif //SEMIAN_H
--- a/ext/semian/semian.h
+++ b/ext/semian/semian.h
@@ -8,10 +8,6 @@ Implements Init_semian, which is used as C/Ruby entrypoint.
 #define SEMIAN_H
 
 // System includes
-#include <sys/types.h>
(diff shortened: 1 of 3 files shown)


ext/semian/semian_types.h
```diff
@@ -0,0 +1,42 @@
...
+  struct seminfo  *__buf;  /* Buffer for IPC_INFO
+                             (Linux-specific) */
+};
+
+// FIXME
+// Why is this needed? (i don't think it is)
+typedef struct {
+  int sem_id;
+  int tickets;
+} update_ticket_count_t;
```
jacobbednarz: only references I can see are https://github.com/Shopify/semian/pull/102/files#diff-7340ca03324f1cf7420c68d866f42818R128

jacobbednarz: and https://github.com/Shopify/semian/blob/fd3604e1269b7b18d6f5771b2345a03d753098ce/ext/semian/semian.c#L150

jacobbednarz:
This seems fine to me and my hacky C👍

ext/semian/semian_types.h
dalehamel: Yeah i changed my mind on this.

I know it's in use, but I was questioning the value of a struct with only two fields vs just passing two params instead of one.

But, i've reversed this opinion.

dalehamel:
For the record, an omnibus header file like this is something I consider an antipattern.

For simplicity while refactoring, it makes sense to lump all the includes together so you don't have to worry about what function needs what.

There are some includes here that are, more or less, entirely globally needed, but many aren't.

A future refactor may bring some of these headers down to the units which need them (especially utility includes like "time", "ipc", etc).

ext/semian/semian.h
```diff
@@ -0,0 +1,30 @@
...
+Implements Init_semian, which is used as C/Ruby entrypoint.
+*/
+
+#ifndef SEMIAN_H
+#define SEMIAN_H
+
+// System includes
+#include <sys/types.h>
+#include <sys/ipc.h>
+#include <sys/sem.h>
```
csfrancis: I would `#include` these in `semian_types.h` since that's where they're being referenced.

dalehamel:
@csfrancis thanks for noticing that, I moved those includes into semian_types.h

In general, i'd like to have semian.h only include other semian includes, or stuff that's needed by *everything* so I appreciate this suggestion.

PR #103, opened 2017-02-12
Extract semian ruby memory management from semian.c

dalehamel:
# What

Extract ruby GC related functions from semian.c

# Why

Part of omnibus refactor

### DIFF
--- a/ext/semian/semian.c
+++ b/ext/semian/semian.c
@@ -40,48 +40,6 @@ raise_semian_syscall_error(const char *syscall, int error_num)
 }
 
 static void
-semian_resource_mark(void *ptr)
-{
-  /* noop */
-}
-
-static void
-semian_resource_free(void *ptr)
-{
-  semian_resource_t *res = (semian_resource_t *) ptr;
-  if (res->name) {
-    free(res->name);
-    res->name = NULL;
-  }
-  xfree(res);
-}
-
-static size_t
-semian_resource_memsize(const void *ptr)
-{
-  return sizeof(semian_resource_t);
-}
-
-static const rb_data_type_t
-semian_resource_type = {
-  "semian_resource",
-  {
-    semian_resource_mark,
-    semian_resource_free,
-    semian_resource_memsize
-  },
-  NULL, NULL, RUBY_TYPED_FREE_IMMEDIATELY
-};
-
-static VALUE
-semian_resource_alloc(VALUE klass)
-{
-  semian_resource_t *res;
-  VALUE obj = TypedData_Make_Struct(klass, semian_resource_t, &semian_resource_type, res);
-  return obj;
-}
-
-static void
 set_semaphore_permissions(int sem_id, int permissions)
 {
   union semun sem_opts;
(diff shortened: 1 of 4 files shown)


dalehamel:
Depends on https://github.com/Shopify/semian/pull/102

ext/semian/semian.c
```diff
@@ -1,25 +1,4 @@
...
-#include <stdio.h>
-
-union semun {
-  int              val;    /* Value for SETVAL */
-  struct semid_ds *buf;    /* Buffer for IPC_STAT, IPC_SET */
-  unsigned short  *array;  /* Array for GETALL, SETALL */
-  struct seminfo  *__buf;  /* Buffer for IPC_INFO
-                             (Linux-specific) */
-};
+#include <semian.h>
```
dalehamel: note should be pulled in by #102

ext/semian/semian_types.h
```diff
@@ -0,0 +1,32 @@
...
+// see man semop
+union semun {
+  int              val;    /* Value for SETVAL */
+  struct semid_ds *buf;    /* Buffer for IPC_STAT, IPC_SET */
+  unsigned short  *array;  /* Array for GETALL, SETALL */
+  struct seminfo  *__buf;  /* Buffer for IPC_INFO
+                             (Linux-specific) */
+};
+
+// FIXME
```
dalehamel: ignore this

csfrancis:
Why don't you base this on #102 to make it easier to review?

dalehamel:
@csfrancis done

ext/semian/semian_resource_alloc.h
```diff
@@ -0,0 +1,47 @@
+/*
+For memory management operations of semian resources.
+*/
+#ifndef SEMIAN_RESOURCE_ALLOC_H
+#define SEMIAN_RESOURCE_ALLOC_H
+
+#include <semian.h>
```
csfrancis: You have a circular reference here. You can probably be more specific in what you need to include.

csfrancis:
> I don't have much experience organizing C, but is this worth it? 

I have similar concerns. I understand your intend, but also don't like the idea of having to navigate through a bunch of files to figure out where something's located.

dalehamel:
> I have similar concerns. I understand your intent, but also don't like the idea of having to navigate through a bunch of files to figure out where something's located.

These could possible be shared with semian_resource, but they don't have much to do with it.

I'd say of files here this is probably the set that is most suitable for merging (semian_resource and semian_resource_alloc).

I think that there's a clear separation between at least the following 3 concerns:

* the semian resource, that implements ruby operations
* semian ticket management
* direct interactions with sysV semaphores.

The function names also already lend themselves to this separation, as all semian_resource functions already include this in their name, likewise the ticket management stuff mentions tickets, and the semaphore stuff directly mentions semaphores.

By splitting things out into separate files, it also makes it easier to refactor complicated functions by pulling out the logic into statics, scoped only to that file.

If you don't agree, then we should probably close this entire set of PRs and revert the previous merge.

dalehamel:
Decided offline that this belongs with the semian_resource code

PR #107, opened 2017-02-13
Refactor semian_resource related functions

dalehamel:
# What

This pulls out semian_resource related functions (stuff exposed to ruby) into resource.c/h.

# How

Moving the code.

I had to drop a few forward declarations of prototypes into sysv_semaphores.h and semian.h. In subsequent PRs this will get cleaned up.

Since there is still a pile of stuff in semian.c (and the definitions for the above prototypes) these forward declarations are a temporary measure to allow everything to compile.


dalehamel:
@sirupsen once this gets merged, I'd like to roll up the existing changes into a gem bump and deploy before merging any more changes

ext/semian/resource.c
```diff
@@ -0,0 +1,262 @@
+#include "resource.h"
+
+// "Private" function forward declarations
```
csfrancis: You don't really need to declare these IMO. Just define them before the the public functions. The fact that they're `static` conveys that they are "private".

ext/semian/resource.h
```diff
@@ -0,0 +1,75 @@
...
+*/
+#ifndef SEMIAN_RESOURCE_H
+#define SEMIAN_RESOURCE_H
+
+#include "semian.h" // FIXME remove this once configure_tickets has been refactored
+#include "types.h"
+#include "sysv_semaphores.h"
+
+// Ruby variables
+ID id_timeout;
```
csfrancis: I think these can go in the .c file.

ext/semian/sysv_semaphores.h
```diff
@@ -0,0 +1,83 @@
...
+VALUE eSyscall, eTimeout, eInternal;
+
+// Here we define an enum value and string representation of each semaphore
+// This allows us to key the sem value and string rep in sync easily
+// utilizing pre-processor macros.
+//   SI_SEM_TICKETS             semaphore for the tickets currently issued
+//   SI_SEM_CONFIGURED_TICKETS  semaphore to track the desired number of tickets available for issue
+//   SI_SEM_LOCK                metadata lock to act as a mutex, ensuring thread-safety for updating other semaphores
+//   SI_NUM_SEMAPHORES          always leave this as last entry for count to be accurate
+#define FOREACH_SEMINDEX(SEMINDEX) \
```
csfrancis: Oh god.

ext/semian/sysv_semaphores.h
```diff
@@ -0,0 +1,83 @@
...
+#define GENERATE_ENUM(ENUM) ENUM,
+#define GENERATE_STRING(STRING) #STRING,
+
+// Generate enum for sem indices
+enum SEMINDEX_ENUM {
+    FOREACH_SEMINDEX(GENERATE_ENUM)
+};
+
+// Generate string rep for sem indices for debugging puproses
+extern const char *SEMINDEX_STRING[];
```
csfrancis: YAGNI?

dalehamel: Since i switched from get_max_tickets to a more generic get_sem_val, i wanted to be able to provide the string representation

ext/semian/sysv_semaphores.h
dalehamel: can you be more specific :)

ext/semian/resource.h
dalehamel: I think you're right about that

ext/semian/sysv_semaphores.h
csfrancis: Just my initial reaction when I see preprocessor meta-programming :-)

ext/semian/sysv_semaphores.h
csfrancis: But I don't see you using it anywhere in this PR?

dalehamel: That is correct, it doesn't come until a later PR. Since this is contentious I'll just remove it.

dalehamel:
> It looks like you're doing more in this PR than just pulling out the resource functions. I think it would be cleaner (and easier to review) if you just did that.

@csfrancis I believe I've done this:

* I've removed sysv_semaphores.h
 * The declarations that were there have been temporarily moved to semian.h, because we need a central place for the declarations to live.

I also believe I've made all the other modifications you've requested, but I have a difference of opinion on static forward declarations

ext/semian/resource.h
```diff
@@ -0,0 +1,75 @@
...
+*/
+
+#ifndef SEMIAN_RESOURCE_H
+#define SEMIAN_RESOURCE_H
+
+#include "semian.h" // FIXME remove this once temporary declarations are removed
+#include "types.h"
+
+// Ruby variables
+ID id_timeout;
```
dalehamel: @csfrancis note that this must be in a header because both semian.c and resource.c need to access it.

ext/semian/semian.c
```diff
@@ -221,79 +161,7 @@ create_semaphore(int key, int permissions, int *created)
...
- */
-static VALUE
-semian_resource_initialize(VALUE self, VALUE id, VALUE tickets, VALUE permissions, VALUE default_timeout)
-{
-  key_t key;
-  int created = 0;
-  semian_resource_t *res = NULL;
-  const char *id_str = NULL;
-
-  if (TYPE(id) != T_SYMBOL && TYPE(id) != T_STRING) {
```
dalehamel: Note that I refactored these type checks into static functions within resource.c.

They now also cast to the correct type. In the existing code there is actually causing truncation in some places where a long is being downcast to an int.

ext/semian/semian.c
```diff
@@ -202,14 +142,14 @@ configure_tickets(int sem_id, int tickets, int should_initialize)
   }
 }
 
-static int
-create_semaphore(int key, int permissions, int *created)
```
dalehamel: This is where it was getting downcasted

ext/semian/resource.c
```diff
@@ -0,0 +1,253 @@
...
+  key_t key;
+  int c_permissions;
+  double c_timeout;
+  int c_tickets;
+  int created = 0;
+  semian_resource_t *res = NULL;
+  const char *c_id_str = NULL;
+
+  // Check and cast arguments
+  c_tickets = check_tickets_arg(tickets);
```
dalehamel: Note that this is where I've moved the argument checks into helper functions.

ext/semian/resource.h
```diff
@@ -0,0 +1,75 @@
...
+int system_max_semaphore_count;
```
csfrancis: Can't this move to the .c file?

ext/semian/semian.c
```diff
@@ -202,14 +142,14 @@ configure_tickets(int sem_id, int tickets, int should_initialize)
...
+int
+create_semaphore(int key, long permissions, int *created)
 {
   int semid = 0;
   int flags = 0;
 
   *created = 0;
-  flags = IPC_EXCL | IPC_CREAT | FIX2LONG(permissions);
```
dalehamel: This check was happening in the wrong spot.

We were truncating long -> int and then going back to long. I've fixed this in the correct spot.

ext/semian/resource.h
dalehamel: it is required by Init_semian and semian_resource_initialize, which are in semian.c and resource.c respectively, so I don't think so.

PR #111, opened 2017-02-14
Check arguments for initialize_resource

dalehamel:
# What

Here we refactor argument checks for resource initialization out into static methods.
The static methods also perform type casting to the desired type.

A potential bug was discovered, where permissions were being downcasted from long to int.
This has been repaired, and signatures have been adjusted to use long instead of int where required.
The type conversion is now happening in exactly one place.

# Why

It's easier to read the initialize_resource function with this checking out of the way.

This also asserts a contract that each argument should have a function that checks it for validity, and then casts it to the correct C type.

### DIFF
--- a/ext/semian/resource.c
+++ b/ext/semian/resource.c
@@ -3,6 +3,18 @@
 static VALUE
 cleanup_semian_resource_acquire(VALUE self);
 
+static int
+check_tickets_arg(VALUE tickets);
+
+static long
+check_permissions_arg(VALUE permissions);
+
+static const
+char *check_id_arg(VALUE id);
+
+static double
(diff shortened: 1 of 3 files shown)


ext/semian/resource.c
```diff
@@ -152,6 +155,62 @@ cleanup_semian_resource_acquire(VALUE self)
   return Qnil;
 }
 
+static long check_permissions_arg(VALUE permissions)
```
sirupsen: It's custom in Semian (and Rubby C-extensions, since Ruby follows this convention) to have the type on a line before the function name

dalehamel: thanks

dalehamel: missed this one

PR #116, opened 2017-02-15
Refactor semian ticket management into its own files

dalehamel:
# What

This separates out ticket management functionality into their own files

# How

The implementation is moved from semian.c to tickets.c

A new file, tickets.h is created with the prototypes for these functions.

A few headers are cleaned up as they have includes that are no longer needed.

*Note:* this is the last of the major refactors that move code around. There will be a few follow ups that reorganize the code within these files that will follow.


ext/semian/semian.c
```diff
@@ -1,83 +1,5 @@
 #include "semian.h"
 
-static VALUE
```
dalehamel: All the red here moves to tickets.c without modification

PR #117, opened 2017-02-16
Refactor get_max_tickets to be a more generic get_sem_val

dalehamel:
# What

replace get_max_tickets with a more generic get_sem_val, to be used to retrieve the value of an arbitrary semaphore in the set.

# How

The function has been refactored to accept a second parameter - the index to use.

I also moved the semian indices to be generated via a C pre-processor technique called "x macros".

**Before you run away screaming**, keep in mind that the x macro is a fairly elegant solution to keeping to data structures in sync in C, and is a standard pattern for **exactly** this type of application. 

The x macro is utilized to keep the enum symbols and strings in sync. This allows for us to easily print which semaphore is being operated on for debugging purposes, or any time we would like to "pretty print" a semaphore by ID.

### DIFF
--- a/ext/semian/sysv_semaphores.h
+++ b/ext/semian/sysv_semaphores.h
@@ -54,10 +54,6 @@ enum SEMINDEX_ENUM {
     FOREACH_SEMINDEX(GENERATE_ENUM)
 };
 
-// Generate string rep for sem indices for debugging puproses
-extern const char *SEMINDEX_STRING[];
-
-
 VALUE eSyscall, eTimeout, eInternal;
 
 // Helper for syscall verbose debugging
--- a/ext/semian/sysv_semaphores.h
+++ b/ext/semian/sysv_semaphores.h
@@ -32,6 +32,32 @@ typedef VALUE (*my_blocking_fn_t)(void*);
 // Time to wait for timed ops to complete
 #define INTERNAL_TIMEOUT 5 /* seconds */
 
+// Here we define an enum value and string representation of each semaphore
+// This allows us to key the sem value and string rep in sync easily
+// utilizing pre-processor macros.
+// If you're unfamiliar with this pattern, this is using "x macros"
+//   SI_SEM_TICKETS             semaphore for the tickets currently issued
+//   SI_SEM_CONFIGURED_TICKETS  semaphore to track the desired number of tickets available for issue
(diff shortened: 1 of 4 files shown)


dalehamel:
I'm going to leave this for @csfrancis to give his 2 cents on then, because we balked at it the first time we saw it.

ext/semian/sysv_semaphores.h
```diff
@@ -32,6 +32,32 @@ typedef VALUE (*my_blocking_fn_t)(void*);
...
+#define GENERATE_ENUM(ENUM) ENUM,
+#define GENERATE_STRING(STRING) #STRING,
+
+// Generate enum for sem indices
+enum SEMINDEX_ENUM {
+    FOREACH_SEMINDEX(GENERATE_ENUM)
+};
+
+// Generate string rep for sem indices for debugging puproses
+extern const char *SEMINDEX_STRING[];
```
csfrancis: Does this need to be `extern`?

dalehamel: @csfrancis as far as I know, because if it's defined in the header it would be redefined if multiple things include it, which is why I pulled the definition into the C file.

I'm open to suggestions, but afaict this is the way it has to go

dalehamel:
```
linking shared-object semian/semian.so
semian.o:(.data.rel.local+0x0): multiple definition of `SEMINDEX_STRING'
resource.o:(.data.rel.local+0x0): first defined here
sysv_semaphores.o:(.data.rel.local+0x0): multiple definition of `SEMINDEX_STRING'
resource.o:(.data.rel.local+0x0): first defined here
tickets.o:(.data.rel.local+0x0): multiple definition of `SEMINDEX_STRING'
resource.o:(.data.rel.local+0x0): first defined here
collect2: error: ld returned 1 exit status
Makefile:254: recipe for target 'semian.so' failed
```

It makes the linker sad :'(

csfrancis: Though you're only referencing it from `sysv_semaphores.c`. I guess my real question is whether it actually needs to be in the .h.

dalehamel: That is a good point.

I'll move it to the .c and make it static.

PR #119, opened 2017-02-16
Refactor tickets.c to make it easier to implement additional ticket strategies

dalehamel:
# What

tickets.c is refactored to make it easier to implement another ticket strategy (quota based).

# How

- configure_tickets is gutted and now just initializes tickets if they need to be initialized, and then initiates the correct strategy (right now there is only one strategy)
- The initialization logic for semaphore tickets has been put into its own function, "initialize_tickets"
- The logic that was previously part of configure_tickets is now a part of "configure_static_tickets".

### DIFF
--- a/ext/semian/tickets.c
+++ b/ext/semian/tickets.c
@@ -1,5 +1,11 @@
 #include <tickets.h>
 
+static void
+initialize_tickets(int sem_id, int tickets);
+
+static void
+configure_static_tickets(int sem_id, int tickets);
+
 VALUE
 update_ticket_count(update_ticket_count_t *tc)
 {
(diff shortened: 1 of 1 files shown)


ext/semian/tickets.c
```diff
@@ -15,58 +21,75 @@ update_ticket_count(update_ticket_count_t *tc)
     }
 
     if (semctl(tc->sem_id, SI_SEM_CONFIGURED_TICKETS, SETVAL, tc->tickets) == -1) {
-      rb_raise(eInternal, "error updating max ticket count, errno: %d (%s)", errno, strerror(errno));
+      rb_raise(eInternal, "error updating configured ticket count, errno: %d (%s)", errno, strerror(errno));
```
sirupsen: `updating configured` sounds a bit redundant, `error configuring ticket count`?

ext/semian/tickets.c
```diff
@@ -15,58 +21,75 @@ update_ticket_count(update_ticket_count_t *tc)
...
 void
 configure_tickets(int sem_id, int tickets, int should_initialize)
 {
+  if (should_initialize) {
+    initialize_tickets(sem_id, tickets);
+  }
+
+  if (tickets > 0) {
+    configure_static_tickets(sem_id, tickets);
+  }
```
dalehamel: A second strategy would get called from right here ... ;)

ext/semian/tickets.c
dalehamel: yeah that's better english

ext/semian/tickets.c
```diff
@@ -15,58 +21,75 @@ update_ticket_count(update_ticket_count_t *tc)
...
+}
+
+static void
+initialize_tickets(int sem_id, int tickets)
+{
   unsigned short init_vals[SI_NUM_SEMAPHORES];
+
+  if (tickets > 0) {
+    init_vals[SI_SEM_TICKETS] = init_vals[SI_SEM_CONFIGURED_TICKETS] = tickets;
```
csfrancis: Why are you only initializing this if `tickets > 0`? With `init_vals` declared on the stack, these elements will contain garbage if not initialized here.

dalehamel: That's a good catch, this conditional doesn't need to be here, yet.

Issue #98, opened 2017-02-08
Dynamic semaphore initialization from ticket quota


dalehamel:
The gist of what is proposed here is to allow us to eliminate the assumption that there are a fixed number of workers (resource consumers) on a particular host. In a more dynamic scheduling environment (think: kubernetes), we cannot be certain of the number of resource consumers on a given host.

This is problematic, because under the current model we would have a ticket quota of a fixed size for a fixed number of workers. For illustration:

* Assume we have a resource that permits 5 tickets (T) -> T = 5
* Assume we have 10 workers (W) -> W = 10
* In this case, a quota  (Q) of only half of the workers may access the resource at a time - > Q = 0.5

So, since W is no longer static, we need T to be able to react to it, such to preserve Q at 0.5.

The proposed implementation is as follows:

* Have a new semaphore based on the maximum number of tickets that tracks the tickets per worker (needs to be unique per process or thread  -probably parent pid or pid_threadid). This would be the "quota semaphore", or "worker quota semaphore", tracking the number of worker tickets that have been issued. As new, unique, workers are added, we decrement this value. As they are removed, increment it.
* The difference between the quota semaphore and the configured global maximum is the number of workers participating in the quota. **This allows us to keep track of W**.
* As we update the quota semaphore, we can dynamically update the number of available RPC tickets (T), in order to maintain the desired quota (Q).

A nuance to this is:

> When workers unregister themselves (they're killed or stop and SEM_UNDO does its thing), the worker count needs to be adjusted by something. We can cache the worker count in a semaphore in the semaphore set for the resource. On #acquire if it's different, we call update_ticket_count. It seems better for this reason to do it at #acquire time rather than #register time.

dalehamel:
Had a chat with @sirupsen and I think that the approach we've settled on boils down to:

```
Have a new semaphore based on the maximum number of tickets that tracks the tickets per worker (needs to be unique per process or thread  -probably parent pid or pid_threadid).

The difference between this and the configured global maximum is the number of tickets currently available for that resource.

Based on the updates to this, we can dynamically update the number of available RPC tickets, based on some fraction. If the floor of this is different from the previous floor, do a thread-safe update of the ticket count. 
```

sirupsen:
@csfrancis

sirupsen:
@casperisfine

dalehamel:
@csfrancis asked me:

> in k8s are sysv semaphores shared across the entire physical host? that seems strange

[...]

PR #120, opened 2017-02-17
Support quota based allocation strategy

dalehamel:
**Note:** This is still a work in progress - particularly I'm missing test coverage, but I'd like a preliminary review (low time investment) on the approach here. I want to make sure that it passes a basic "kick-the-tires" test before devoting more time to something I may potentially have to drastically change.

Specifically, I'm looking for:

* Things I've blatantly overlooked or misunderstood
* Design issues or shortcomings
* Edge cases that I've missed during testing
* Tests cases that I should add but haven't

This still contains some debug prints / reminders to myself, please ignore those.

For deploying this, I'd like to:

* Do a first round of review on the approach / tests as described above
* Ship existing semian changes as 0.6.3
* Finalize the review / acceptance of these changes
* Release this as 0.7.0 due to the new functionality it adds

# What

This implements a new strategy for ticket allocation that is based on the number of workers per the function:

```
ceil(workers * quota)
```

Such that the number of configured tickets shall always be proportional to the number of registered workers.

# How

* If a quota > 0 is specified, we toggle to using the 'quota' strategy instead of the 'static ticket count' strategy.
* Ticket counts provided by resource.ticket come directly from the number of configured tickets instead of a ruby instance variable, since this can now be variable
* Whenever a worker registers using the quota strategy, it increments SI_SEM_REGISTERED_WORKER
* Whenever the quota is updated based on SI_SEM_REGISTERED_WORKER, we set SI_SEM_CONFIGURED_WORKER to match SI_SEM_REGISTERED_WORKER, and configure SI_SEM_CONFIGURED_TICKETS according to the configured quota
* We updated the quotas on each call to acquire, in case there is a mismatch between the configured and registered workers

# Testing

The most difficult part of this PR so far has been developing a testing strategy.

I lost a fair bit of time trying to use the existing create_resource helper along with a forking strategy. This is doomed to failure, and will cause resources under tests to be spontaneously deleted, which is quite aggravating.

I tried a few other strategies, including spawning new ruby VMs to run a fixture script that would simply acquire a lock.

I ended up going back to a forking strategy, but manually keeping track of the forked PIDs.

The key here is the TRAP and SIGNAL approach I am using. Any forked child that manages to acquire the ticket will be sleep forever until it is signaled. This allow us to avoid guesswork and race conditions while reasoning about workers and ticket counts.

# Outstanding issues

* [ ] How should small ticket counts be address during, for instance, rolling deploys? Should a minimum ticket count be specified?
* [ ] How should 1 registration per worker be enforced?
* [ ] What should the bounds of a quota value be? Should a quota of 1 be allowed?

Fixes https://github.com/Shopify/semian/issues/98 (when done)

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,67 @@ client = Redis.new(semian: {
 })
 ```
 
+#### Warning: Thread unsafe
+
+Semian is currently known to be thread-unsafe, and is intended to be used by *separate processes*.
+
+Internally, semian uses `SEM_UNDO` for several sysv semaphore operations:
+
+* Acquire
+* Worker registration
+* Semaphore metadata state lock
+
+The intention behind `SEM_UNDO` is that a semaphore operation is automatically undone when the process exits. This
+is true even if the process exits abnormally - crashes, receives a `SIG_KILL`, etc, because it is handled by
+the operating system and not the process itself.
+
+If, however, a thread performs a semop, the `SEM_UNDO` is on its parent process. This means that the operation
+*will not* be undone when the thread exits. This can result in the following unfavorable behavior when using
+threads:
+
+* Threads acquire a resource, but are killed and the resource ticket is never released. For a process, the
+ticket would be released by `SEM_UNDO`, but since it's a thread there is the potential for ticket starvation.
+This can result in deadlock on the resource.
+* Threads that register workers on a resource but are killed and never unregistered. For a process, the worker
+count would be automatically decremented by `SEM_UNDO`, but for threads the worker count will continue to increment,
+only being undone when the parent process dies. This can cause the number of tickets to dramatically exceed the quota.
+* If a thread acquires the semaphore metadata lock and dies before releasing it, semian will deadlock on anything
+attempting to acquire the metadata lock until the thread's parent process exits. This can prevent the ticket count
+from being updated.
+
+Moreover, a strategy that utilizes `SEM_UNDO` is not compatible with a strategy that attempts to the semaphores tickets manually.
+In order to support threads, operations that currently use `SEM_UNDO` would need to use no semaphore flag, and the calling process
+will be responsible for ensuring that threads are appropriately cleaned up. It is still possible to implement this, but
+it would likely require an in-memory semaphore managed by the parent process of the threads. PRs welcome for this functionality.
+
+#### Quotas
+
+You may now set quotas per worker:
+
+```ruby
+client = Redis.new(semian: {
+  name: "inventory",
+  quota: 0.5,
+  success_threshold: 2,
+  error_threshold: 4,
+  error_timeout: 20
+})
+
+```
+
+Per the above example, you no longer need to care about the number of tickets.
+
+Rather, the tickets shall be computed as a proportion of the number of active workers.
+
+In this case, we'd allow 50% of the workers on a particular host to connect to this redis resource.
+
+**Note**:
+
+- You must pass **exactly** one of ticket or quota.
+- Tickets available will be the ceiling of the quota ratio to the number of workers
+ - So, with one worker, there will always be a minimum of 1 ticket
+- Workers in different processes will automatically unregister when the process exits.
+
 #### Net::HTTP
 For the `Net::HTTP` specific Semian adapter, since many external libraries may create
 HTTP connections on the user's behalf, the parameters are instead provided
--- a/ext/semian/resource.c
+++ b/ext/semian/resource.c
@@ -3,6 +3,12 @@
 static VALUE
 cleanup_semian_resource_acquire(VALUE self);
 
+static void
+check_tickets_xor_quota_arg(VALUE tickets, VALUE quota);
+
+static double
+check_quota_arg(VALUE quota);
+
 static int
 check_tickets_arg(VALUE tickets);
 
@@ -63,9 +69,20 @@ semian_resource_acquire(int argc, VALUE *argv, VALUE self)
 VALUE
 semian_resource_destroy(VALUE self)
 {
+  struct timespec ts = { 0 };
+  ts.tv_sec = INTERNAL_TIMEOUT;
+
   semian_resource_t *res = NULL;
 
   TypedData_Get_Struct(self, semian_resource_t, &semian_resource_type, res);
+
+  // Prevent a race to deletion
+  if (perform_semop(res->sem_id, SI_SEM_LOCK, -1, 0, &ts) == -1) {
+    if (errno == EINVAL || errno == EIDRM) {
+      return Qtrue;
+    }
+  }
+
   if (semctl(res->sem_id, SI_NUM_SEMAPHORES, IPC_RMID) == -1) {
     raise_semian_syscall_error("semctl()", errno);
   }
@@ -89,6 +106,21 @@ semian_resource_count(VALUE self)
 }
 
 VALUE
+semian_resource_tickets(VALUE self)
+{
+  int ret;
+  semian_resource_t *res = NULL;
+
+  TypedData_Get_Struct(self, semian_resource_t, &semian_resource_type, res);
+  ret = semctl(res->sem_id, SI_SEM_CONFIGURED_TICKETS, GETVAL);
+  if (ret == -1) {
+    raise_semian_syscall_error("semctl()", errno);
+  }
+
+  return LONG2FIX(ret);
+}
+
+VALUE
 semian_resource_id(VALUE self)
 {
   semian_resource_t *res = NULL;
@@ -97,17 +129,18 @@ semian_resource_id(VALUE self)
 }
 
 VALUE
-semian_resource_initialize(VALUE self, VALUE id, VALUE tickets, VALUE permissions, VALUE default_timeout)
+semian_resource_initialize(VALUE self, VALUE id, VALUE tickets, VALUE quota, VALUE permissions, VALUE default_timeout)
 {
-  key_t key;
-  int c_permissions;
+  long c_permissions;
   double c_timeout;
+  double c_quota;
   int c_tickets;
-  int created = 0;
   semian_resource_t *res = NULL;
   const char *c_id_str = NULL;
 
   // Check and cast arguments
+  check_tickets_xor_quota_arg(tickets, quota);
+  c_quota = check_quota_arg(quota);
   c_tickets = check_tickets_arg(tickets);
   c_permissions = check_permissions_arg(permissions);
   c_id_str = check_id_arg(id);
@@ -119,19 +152,8 @@ semian_resource_initialize(VALUE self, VALUE id, VALUE tickets, VALUE permission
   // Populate struct fields
   ms_to_timespec(c_timeout * 1000, &res->timeout);
   res->name = strdup(c_id_str);
-
-  // Get or create semaphore set
-  //  note that tickets = 0 will be used to acquire a semaphore set after it's been created elswhere
-  key = generate_key(c_id_str);
-  res->sem_id = c_tickets == 0 ? get_semaphore(key) : create_semaphore(key, c_permissions, &created);
-  if (res->sem_id == -1) {
-    raise_semian_syscall_error("semget()", errno);
-  }
-
-  set_semaphore_permissions(res->sem_id, c_permissions);
-
-  // Configure semaphore ticket counts
-  configure_tickets(res->sem_id, c_tickets, created);
+  res->quota = c_quota;
+  res->sem_id = initialize_semaphore_set(c_id_str, c_permissions, c_tickets, c_quota);
 
   return self;
 }
@@ -160,6 +182,33 @@ check_permissions_arg(VALUE permissions)
 {
   Check_Type(permissions, T_FIXNUM);
   return FIX2LONG(permissions);
+}
+
+static void
+check_tickets_xor_quota_arg(VALUE tickets, VALUE quota)
+{
+  if ((TYPE(tickets) == T_NIL && TYPE(quota) == T_NIL) ||(TYPE(tickets) != T_NIL && TYPE(quota) != T_NIL)){
+    rb_raise(rb_eArgError, "Must pass exactly one of ticket or quota");
+  }
+}
+
+static double
+check_quota_arg(VALUE quota)
+{
+  double c_quota;
+
+  if (TYPE(quota) != T_NIL) {
+    if (TYPE(quota) != T_FIXNUM && TYPE(quota) != T_FLOAT) {
+      rb_raise(rb_eTypeError, "expected decimal type for quota");
+    }
+    if (NUM2DBL(quota) <= 0 || NUM2DBL(quota) > 1) {
+      rb_raise(rb_eArgError, "quota must be a decimal between 0 and 1");
+    }
+    c_quota = NUM2DBL(quota);
+  } else {
+    c_quota = -1.0f;
+  }
+  return c_quota;
 }
 
 static int
--- a/ext/semian/resource.h
+++ b/ext/semian/resource.h
@@ -7,7 +7,6 @@ Functions here are associated with rubyland operations.
 #define SEMIAN_RESOURCE_H
 
 #include "types.h"
-#include "tickets.h"
 #include "sysv_semaphores.h"
 
 // Ruby variables
@@ -16,12 +15,12 @@ int system_max_semaphore_count;
 
 /*
  * call-seq:
- *    Semian::Resource.new(id, tickets, permissions, default_timeout) -> resource
+ *    Semian::Resource.new(id, tickets, quota, permissions, default_timeout) -> resource
  *
  * Creates a new Resource. Do not create resources directly. Use Semian.register.
  */
 VALUE
-semian_resource_initialize(VALUE self, VALUE id, VALUE tickets, VALUE permissions, VALUE default_timeout);
+semian_resource_initialize(VALUE self, VALUE id, VALUE tickets, VALUE quota, VALUE permissions, VALUE default_timeout);
 
 /*
  * call-seq:
@@ -58,6 +57,15 @@ semian_resource_destroy(VALUE self);
  */
 VALUE
 semian_resource_count(VALUE self);
+
+/*
+ * call-seq:
+ *    resource.tickets -> count
+ *
+ * Returns the configured number of tickets for a resource.
+ */
+VALUE
+semian_resource_tickets(VALUE self);
 
 /*
  * call-seq:
--- a/ext/semian/sysv_semaphores.c
+++ b/ext/semian/sysv_semaphores.c
@@ -1,7 +1,16 @@
 #include "sysv_semaphores.h"
 
+static key_t
+generate_key(const char *name);
+
 static void *
 acquire_semaphore(void *p);
+
+static int
+wait_for_new_semaphore_set(key_t key, long permissions);
+
+static void
+initialize_new_semaphore_values(int sem_id, long permissions);
 
 // Generate string rep for sem indices for debugging puproses
 static const char *SEMINDEX_STRING[] = {
@@ -14,28 +23,39 @@ raise_semian_syscall_error(const char *syscall, int error_num)
   rb_raise(eSyscall, "%s failed, errno: %d (%s)", syscall, error_num, strerror(error_num));
 }
 
-key_t
-generate_key(const char *name)
+int
+initialize_semaphore_set(const char* id_str, long permissions, int tickets, double quota)
 {
-  char semset_size_key[20];
-  char *uniq_id_str;
+  int sem_id;
+  key_t key;
 
-  // It is necessary for the cardinatily of the semaphore set to be part of the key
-  // or else sem_get will complain that we have requested an incorrect number of sems
-  // for the desired key, and have changed the number of semaphores for a given key
-  sprintf(semset_size_key, "_NUM_SEMS_%d", SI_NUM_SEMAPHORES);
-  uniq_id_str = malloc(strlen(name)+strlen(semset_size_key)+1);
-  strcpy(uniq_id_str, name);
-  strcat(uniq_id_str, semset_size_key);
+  key = generate_key(id_str);
+  sem_id = semget(key, SI_NUM_SEMAPHORES, IPC_CREAT | IPC_EXCL | permissions);
 
-  union {
-    unsigned char str[SHA_DIGEST_LENGTH];
-    key_t key;
-  } digest;
-  SHA1((const unsigned char *) uniq_id_str, strlen(uniq_id_str), digest.str);
-  free(uniq_id_str);
-  /* TODO: compile-time assertion that sizeof(key_t) > SHA_DIGEST_LENGTH */
-  return digest.key;
+  /*
+  This approach is based on http://man7.org/tlpi/code/online/dist/svsem/svsem_good_init.c.html
+  which avoids race conditions when initializing semaphore sets.
+  */
+  if (sem_id != -1) {
+    // Happy path - we are the first worker, initialize the semaphore set.
+    initialize_new_semaphore_values(sem_id, permissions);
+  } else {
+    // Something went wrong
+    if (errno != EEXIST) {
+      raise_semian_syscall_error("semget() failed to initialize semaphore values", errno);
+    } else {
+      // The semaphore set already exists, ensure it is initialized
+      sem_id = wait_for_new_semaphore_set(key, permissions);
+    }
+  }
+
+  set_semaphore_permissions(sem_id, permissions);
+
+  sem_meta_lock(sem_id); // Sets otime for the first time by acquiring the sem lock
+  configure_tickets(sem_id, tickets,  quota);
+  sem_meta_unlock(sem_id);
+
+  return sem_id;
 }
 
 void
@@ -52,26 +72,6 @@ set_semaphore_permissions(int sem_id, long permissions)
     semctl(sem_id, 0, IPC_SET, sem_opts);
   }
 }
-
-int
-create_semaphore(int key, long permissions, int *created)
-{
-  int semid = 0;
-  int flags = 0;
-
-  *created = 0;
-  flags = IPC_EXCL | IPC_CREAT | permissions;
-
-  semid = semget(key, SI_NUM_SEMAPHORES, flags);
-  if (semid >= 0) {
-    *created = 1;
-  } else if (semid == -1 && errno == EEXIST) {
-    flags &= ~IPC_EXCL;
-    semid = semget(key, SI_NUM_SEMAPHORES, flags);
-  }
-  return semid;
-}
-
 
 int
 perform_semop(int sem_id, short index, short op, short flags, struct timespec *ts)
@@ -94,7 +94,7 @@ get_sem_val(int sem_id, int sem_index)
 {
   int ret = semctl(sem_id, sem_index, GETVAL);
   if (ret == -1) {
-    rb_raise(eInternal, "error getting value of %s, errno: %d (%s)", SEMINDEX_STRING[sem_index], errno, strerror(errno));
+    rb_raise(eInternal, "error getting value of %s for sem %d, errno: %d (%s)", SEMINDEX_STRING[sem_index], sem_id, errno, strerror(errno));
   }
   return ret;
 }
@@ -136,8 +136,92 @@ acquire_semaphore(void *p)
 {
   semian_resource_t *res = (semian_resource_t *) p;
   res->error = 0;
+#ifdef DEBUG
+  print_sem_vals(res->sem_id);
+#endif
   if (perform_semop(res->sem_id, SI_SEM_TICKETS, -1, SEM_UNDO, &res->timeout) == -1) {
     res->error = errno;
   }
   return NULL;
+}
+
+static key_t
+generate_key(const char *name)
+{
+  char semset_size_key[20];
+  char *uniq_id_str;
+
+  // It is necessary for the cardinatily of the semaphore set to be part of the key
+  // or else sem_get will complain that we have requested an incorrect number of sems
+  // for the desired key, and have changed the number of semaphores for a given key
+  sprintf(semset_size_key, "_NUM_SEMS_%d", SI_NUM_SEMAPHORES);
+  uniq_id_str = malloc(strlen(name)+strlen(semset_size_key)+1);
+  strcpy(uniq_id_str, name);
+  strcat(uniq_id_str, semset_size_key);
+
+  union {
+    unsigned char str[SHA_DIGEST_LENGTH];
+    key_t key;
+  } digest;
+  SHA1((const unsigned char *) uniq_id_str, strlen(uniq_id_str), digest.str);
+  free(uniq_id_str);
+  /* TODO: compile-time assertion that sizeof(key_t) > SHA_DIGEST_LENGTH */
+  return digest.key;
+}
+
+
+static void
+initialize_new_semaphore_values(int sem_id, long permissions)
+{
+  unsigned short init_vals[SI_NUM_SEMAPHORES];
+
+  init_vals[SI_SEM_TICKETS] = init_vals[SI_SEM_CONFIGURED_TICKETS] = 0;
+  init_vals[SI_SEM_REGISTERED_WORKERS] = 0;
+  init_vals[SI_SEM_LOCK] = 1;
+
+  if (semctl(sem_id, 0, SETALL, init_vals) == -1) {
+    raise_semian_syscall_error("semctl()", errno);
+  }
+#ifdef DEBUG
+  print_sem_vals(sem_id);
+#endif
+}
+
+static int
+wait_for_new_semaphore_set(key_t key, long permissions)
+{
+  int i;
+  int sem_id = -1;
+  union semun sem_opts;
+  struct semid_ds sem_ds;
+
+  sem_opts.buf = &sem_ds;
+  sem_id = semget(key, 1, permissions);
+
+  if (sem_id == -1){
+      raise_semian_syscall_error("semget()", errno);
+  }
+
+  for (i = 0; i < ((INTERNAL_TIMEOUT * MICROSECONDS_IN_SECOND) / INIT_WAIT); i++) {
+
+    if (semctl(sem_id, 0, IPC_STAT, sem_opts) == -1) {
+      raise_semian_syscall_error("semget()", errno);
+    }
+
+    // If a semop has been performed by someone else, the values must be initialized
+    if (sem_ds.sem_otime != 0) {
+      break;
+    }
+
+#ifdef DEBUG
+    printf("Waiting for another process to initialize semaphore values, checked: %d times\n", i);
+#endif
+    usleep(INIT_WAIT);
+  }
+
+  if (sem_ds.sem_otime == 0) {
+    rb_raise(eTimeout, "error: timeout waiting for semaphore values to initialize after %d checks", INTERNAL_TIMEOUT);
+  }
+
+  return sem_id;
 }
--- a/ext/semian/sysv_semaphores.h
+++ b/ext/semian/sysv_semaphores.h
@@ -17,6 +17,7 @@ and functions associated directly weth semops.
 #include <ruby/io.h>
 
 #include "types.h"
+#include "tickets.h"
 
 // Defines for ruby threading primitives
 #if defined(HAVE_RB_THREAD_CALL_WITHOUT_GVL) && defined(HAVE_RUBY_THREAD_H)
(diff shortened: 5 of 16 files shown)


ext/semian/resource.c
```diff
@@ -34,6 +40,11 @@ semian_resource_acquire(int argc, VALUE *argv, VALUE self)
   TypedData_Get_Struct(self, semian_resource_t, &semian_resource_type, self_res);
   res = *self_res;
 
+  if (res.quota > 0) {
+    // Ensure that configured tickets match the quota before acquiring
+    update_tickets_from_quota(res.sem_id, res.quota);
```
dalehamel: It is *extremely* important that all code in this path block for the least amount of time possible.

ext/semian/resource.c
```diff
@@ -53,7 +64,9 @@ semian_resource_acquire(int argc, VALUE *argv, VALUE self)
     if (res.error == EAGAIN) {
       rb_raise(eTimeout, "timed out waiting for resource '%s'", res.name);
     } else {
-      raise_semian_syscall_error("semop()", res.error);
+      char semerr[40];
+      sprintf(semerr, "semop() %d", res.sem_id);
```
dalehamel: This is for debugging (saying what semaphore fucked up), will be removed before merging.

ext/semian/resource.c
```diff
@@ -66,7 +79,7 @@ semian_resource_destroy(VALUE self)
   semian_resource_t *res = NULL;
 
   TypedData_Get_Struct(self, semian_resource_t, &semian_resource_type, res);
-  if (semctl(res->sem_id, SI_NUM_SEMAPHORES, IPC_RMID) == -1) {
```
dalehamel: revert me

ext/semian/resource.c
```diff
@@ -89,6 +102,21 @@ semian_resource_count(VALUE self)
 }
 
 VALUE
+semian_resource_tickets(VALUE self)
```
dalehamel: Added because tickets are no longer a static count.

ext/semian/sysv_semaphores.c
```diff
@@ -136,6 +136,7 @@ acquire_semaphore(void *p)
 {
   semian_resource_t *res = (semian_resource_t *) p;
   res->error = 0;
+  //printf("Tickets %d\n", get_sem_val(res->sem_id, SI_SEM_TICKETS));
```
dalehamel: revert me

ext/semian/sysv_semaphores.h
```diff
@@ -39,11 +39,15 @@ typedef VALUE (*my_blocking_fn_t)(void*);
...
 //   SI_SEM_CONFIGURED_TICKETS  semaphore to track the desired number of tickets available for issue
 //   SI_SEM_LOCK                metadata lock to act as a mutex, ensuring thread-safety for updating other semaphores
+//   SI_SEM_REGISTERED_WORKERS  semaphore for the number of workers currently registered
+//   SI_SEM_CONFIGURED_WORKERS  semaphore for the number of workers that our quota is using for configured tickets
 //   SI_NUM_SEMAPHORES          always leave this as last entry for count to be accurate
 #define FOREACH_SEMINDEX(SEMINDEX) \
         SEMINDEX(SI_SEM_TICKETS)   \
         SEMINDEX(SI_SEM_CONFIGURED_TICKETS)  \
         SEMINDEX(SI_SEM_LOCK)   \
+        SEMINDEX(SI_SEM_REGISTERED_WORKERS)  \
```
dalehamel: adds our new semaphores

ext/semian/tickets.c
```diff
@@ -27,31 +30,101 @@ update_ticket_count(update_ticket_count_t *tc)
...
+int
+update_tickets_from_quota(int sem_id, double quota)
+{
+  int delta = 0;
+  int tickets = 0;
+  int state = 0;
+  update_ticket_count_t tc;
+
+  // DEBUG PRINT - delete me
+  //printf("%d %d %d %d\n",
```
dalehamel: revert me

ext/semian/tickets.c
```diff
@@ -27,31 +30,101 @@ update_ticket_count(update_ticket_count_t *tc)
...
+  //  get_sem_val(sem_id, SI_SEM_TICKETS),
+  //  get_sem_val(sem_id, SI_SEM_CONFIGURED_TICKETS),
+  //  get_sem_val(sem_id, SI_SEM_REGISTERED_WORKERS),
+  //  get_sem_val(sem_id, SI_SEM_CONFIGURED_WORKERS)
+  //);
+  // If the configured worker count doesn't match the registered worker count, adjust it.
+  // and adjust the underlying tickets available to match.
+  sem_meta_lock(sem_id);
```
dalehamel: Lock placed here as this is the only critical section.

It's important that the lock is *before* the delta is computed, so that multiple processes don't waste their time here if they don't need to on the perform_semop.

ext/semian/tickets.c
```diff
@@ -27,31 +30,101 @@ update_ticket_count(update_ticket_count_t *tc)
...
+  delta = get_sem_val(sem_id, SI_SEM_REGISTERED_WORKERS) - get_sem_val(sem_id, SI_SEM_CONFIGURED_WORKERS);
+  if (delta != 0) {
+    if (perform_semop(sem_id, SI_SEM_CONFIGURED_WORKERS, delta, 0, NULL) == -1) {
+      rb_raise(eInternal, "error setting configured workers, errno: %d (%s)", errno, strerror(errno));
+    }
+
+    // Compute the ticket count
+    tickets = (int) ceil(get_sem_val(sem_id, SI_SEM_CONFIGURED_WORKERS) * quota);
+    // DEBUG PRINT - delete me
+    //printf("Configured ticket count %d with quota %f and workers %d\n", tickets, quota, get_sem_val(sem_id, SI_SEM_CONFIGURED_WORKERS));
```
dalehamel: REVERT ME

ext/semian/tickets.c
```diff
@@ -27,31 +30,101 @@ update_ticket_count(update_ticket_count_t *tc)
...
 }
 
+
+static void
+configure_quota_tickets(int sem_id, double quota)
+{
+  int state;
+
+  // Ensure that a worker for this process is registered
+  if (perform_semop(sem_id, SI_SEM_REGISTERED_WORKERS, 1, SEM_UNDO | IPC_NOWAIT, NULL) == -1) {
```
dalehamel: this doesn't require a meta lock, we want to have the worker register itself immediately. IPC_NOWAIT is safe with positive integers.

ext/semian/tickets.c
```diff
@@ -27,31 +30,101 @@ update_ticket_count(update_ticket_count_t *tc)
...
+    rb_raise(eInternal, "error incrementing registered workers, errno: %d (%s)", errno, strerror(errno));
+  }
+  state = update_tickets_from_quota(sem_id, quota);
+
+  if (state) {
```
dalehamel: In case we need to handle an exception in rubyland

test/resource_test.rb
```diff
@@ -319,4 +398,21 @@ def destroy_resources
     end
     @resources = []
   end
+
+  def fork_workers(count:, resource: :testing, quota:, timeout: 0.1, sleep_seconds: 1)
+    (1..count).each do
+      fork do
+        begin
+          resource = Semian::Resource.new(resource.to_sym, quota: quota, timeout: timeout)
```
dalehamel: Note that using create_resource doesn't work as expected from a threaded context, as the forked workers appear to try and run the "teardown" handler on their own / out of sync with the test runner.

This leads to a race that can cause forked processes to fail with EIDRM.

To avoid this, we spawn the workers ourselves and must use ensure on our own that they are killed / terminate. This is done with Process.waitall usually.

dalehamel:
@csfrancis wanted to get this in front of your :eyes: while you're still available, even though it's not 100% complete.

@sirupsen @csfrancis see remarks in description

dalehamel:
FYI ran the current test suite in a continuous loop for the resource_test.rb for almost 2 hours, running several hundred tests. Not a single failure.

ext/semian/tickets.c
```diff
@@ -27,31 +34,93 @@ update_ticket_count(update_ticket_count_t *tc)
...
     raise_semian_syscall_error("semctl()", errno);
   }
 }
 
+
+static void
+configure_quota_tickets(int sem_id, double quota)
+{
+  // Ensure that a worker for this process is registered
+  if (perform_semop(sem_id, SI_SEM_REGISTERED_WORKERS, 1, SEM_UNDO | IPC_NOWAIT, NULL) == -1) {
```
dalehamel: I'm not sure that it is possible to guarantee that a worker only registers once, which is why you have to increment the worker count whenever a call to configure the tickets is requested.

I recall talking about this exact subject with you, and that was the conclusion we arrived at. 

Perhaps you have another idea?

ext/semian/tickets.c
```diff
@@ -27,31 +34,93 @@ update_ticket_count(update_ticket_count_t *tc)
...
+    rb_raise(eInternal, "error incrementing registered workers, errno: %d (%s)", errno, strerror(errno));
+  }
+  update_tickets_from_quota(sem_id, quota);
```
dalehamel: Perhaps a minimum quota is needed to protect against this?

ext/semian/tickets.c
```diff
@@ -27,31 +34,93 @@ update_ticket_count(update_ticket_count_t *tc)
...
+
   if (should_initialize) {
-    initialize_tickets(sem_id, tickets);
+    initialize_tickets(sem_id, tickets, quota);
   }
 
   if (tickets > 0) {
     configure_static_tickets(sem_id, tickets);
+  } else if (quota > 0) {
+    configure_quota_tickets(sem_id, quota);
```
dalehamel: Static tickets just sets it directly.

The quota tickets must compute what it *should* be set to first, then calls the same function to update the ticket count.

I'm not quite sure I understand your concern via text

ext/semian/resource.c
```diff
@@ -162,6 +196,33 @@ check_permissions_arg(VALUE permissions)
...
+  }
+}
+
+static double
+check_quota_arg(VALUE quota)
+{
+  double c_quota;
+
+  if (TYPE(quota) != T_NIL) {
+    if (TYPE(quota) != T_FIXNUM && TYPE(quota) != T_FLOAT) {
```
dalehamel: 0 is definitely invalid and discarded, but integer 1 is "technicall" valid.

ext/semian/resource.c
```diff
@@ -162,6 +196,33 @@ check_permissions_arg(VALUE permissions)
...
+      rb_raise(rb_eTypeError, "expected decimal type for quota");
+    }
+    if (NUM2DBL(quota) <= 0 || NUM2DBL(quota) > 1) {
```
dalehamel: I couldn't think of why you'd ever want 1, but you might want exactly as many tickets as you have workers so I figured it was valid to leave it as an option.

dalehamel:
@sirupsen wouldn't let me reply to you inline about your remark:

> Why would you ever need to configure it from here? Is this to deal with workers that die and release the semaphore? It's configured from configure_tickets which is called every time we register—if a worker dies, I think we can be fairly certain a worker will go through the configure_tickets code-paths soon (e.g. killed by the Unicorn master for memory hogging or timeout).

Yes, this was for in case a worker died or was removed for any reason since the resource was configured.

dalehamel:
@sirupsen thanks for the review

> I've given this a first pass. I haven't dug into the tests yet. I don't see any overarching problems here, but I think we can simplify by not separating configuring quotas from static tickets. That would give me more confidence in the code.

I'm not sure i fully understand your meaning, would probably be best if we had a chat in person to clear up what you're thinking.

ext/semian/resource.c
dalehamel: oh, here it's letting me reply, weird.

Yes, this is because workers die the registered workers will be out of since with the configured workers.

ext/semian/tickets.c
dalehamel: The only alternate approach I can think of would be to somehow use shared memory to "lock" per some ID, but that adds a bit more scope here as there's nothing else using shared memory so it would be a novel case.

sirupsen: Yeah, I think this is OK. Can we store this in a global class variable or something to _make sure_ people never register twice?

ext/semian/tickets.c
sirupsen: I think addressing https://github.com/Shopify/semian/pull/120#discussion_r102572522 will address that.

ext/semian/tickets.c
```diff
@@ -1,12 +1,82 @@
...
+  update_ticket_count_t tc;
+
+  if (should_initialize) {
+    initialize_tickets(sem_id, tickets, quota);
+  }
+
+  if (quota > 0) {
+    tickets = calculate_quota_tickets(sem_id, quota);
+  } else {
+    wait_for_init(sem_id);
```
dalehamel: the wait_for_init logic was extracted from the static ticket configuration.

It's only valid for static tickets, which is why it's separated out here. With quota it will cause an infinite wait (well, a timeout).

ext/semian/tickets.c
```diff
@@ -1,12 +1,73 @@
...
+calculate_quota_tickets(int sem_id, double quota);
+
 static void
-configure_static_tickets(int sem_id, int tickets);
+wait_for_init(int sem_id);
+
+void
+configure_tickets(int sem_id, int tickets, double quota, int should_initialize)
+{
+  int state;
```
csfrancis: I would initialize all the locals here.

ext/semian/tickets.c
```diff
@@ -27,40 +95,42 @@ update_ticket_count(update_ticket_count_t *tc)
...
-static void
-initialize_tickets(int sem_id, int tickets)
-{
-  unsigned short init_vals[SI_NUM_SEMAPHORES];
+  // If the configured worker count doesn't match the registered worker count, adjust it.
+  // and adjust the underlying tickets available to match.
+  delta = get_sem_val(sem_id, SI_SEM_REGISTERED_WORKERS) - get_sem_val(sem_id, SI_SEM_CONFIGURED_WORKERS);
+  if (delta != 0) {
+    sem_meta_lock(sem_id);
+    if (perform_semop(sem_id, SI_SEM_CONFIGURED_WORKERS, delta, 0, NULL) == -1) {
```
dalehamel: ah of course, because it's potentially racy - nice catch.

ext/semian/tickets.c
dalehamel: Yeah, wouldn't want to crash the VM with a null pointer deref... again ;)

ext/semian/resource.h
```diff
@@ -21,7 +21,7 @@ int system_max_semaphore_count;
  * Creates a new Resource. Do not create resources directly. Use Semian.register.
  */
 VALUE
-semian_resource_initialize(VALUE self, VALUE id, VALUE tickets, VALUE permissions, VALUE default_timeout);
+semian_resource_initialize(VALUE self, VALUE id, VALUE tickets, VALUE quota, VALUE permissions, VALUE default_timeout);
```
dalehamel: Yeah i think I had a fixme for that, at some point.

Regarding the name, initialize_semaphore I'm on the fence as this handles more than *just* semaphore initialization. The rest of the functions use semian_resource as a prefix, and they are all associated with the rubyland semian resource object. Then again, as far as rubyland is concerned this is all just the magic semaphore code... so I don't really care what we call it.

csfrancis: Yeah, I don't think the name matters _that_ much since it's really internal to the Semian resource. In fact, this should probably be defined with `rb_define_protected_method`.

dalehamel: I'll look into that

ext/semian/sysv_semaphores.c
```diff
@@ -14,28 +23,35 @@ raise_semian_syscall_error(const char *syscall, int error_num)
...
 
-  union {
-    unsigned char str[SHA_DIGEST_LENGTH];
-    key_t key;
-  } digest;
-  SHA1((const unsigned char *) uniq_id_str, strlen(uniq_id_str), digest.str);
-  free(uniq_id_str);
-  /* TODO: compile-time assertion that sizeof(key_t) > SHA_DIGEST_LENGTH */
-  return digest.key;
+  if (sem_id != -1) {
```
dalehamel: Note: This has been based off of http://man7.org/tlpi/code/online/dist/svsem/svsem_good_init.c.html which avoids a race condition when initializing the semaphore set. Our previous implementation was pretty literally [how to do it the wrong way](http://man7.org/tlpi/code/online/diff/svsem/svsem_bad_init.c.html).

It's still possible that the process that tries to init will die, but in that case we will throw an error.

It's possible that we may want to fallback to the "Bad" approach, but I think an explicit error is better.

ext/semian/tickets.c
```diff
@@ -1,12 +1,51 @@
...
+  if (quota > 0) {
+    tickets = calculate_quota_tickets(sem_id, quota);
+  }
+
+  /*
+    A manually specified ticket count of 0 is special, meaning "don't set"
+    We need to throw an error if we set it to 0 during initialization.
+    Otherwise, we back out of here completely.
+  */
+  if (get_sem_val(sem_id, SI_SEM_CONFIGURED_TICKETS) == 0 && tickets == 0) {
```
dalehamel: this is a bit ugly, but I think it's the best spot to put this check in @sirupsen

lib/semian.rb
```diff
@@ -143,7 +148,7 @@ def register(name, tickets:, permissions: 0660, timeout: 0, error_threshold:, er
       exceptions: Array(exceptions) + [::Semian::BaseError],
       implementation: ::Semian::Simple,
     )
-    resource = Resource.new(name, tickets: tickets, permissions: permissions, timeout: timeout)
+    resource = Resource.instance(name, tickets: tickets, quota: quota, permissions: permissions, timeout: timeout)
```
dalehamel: We now call resource.instance instead of resource.new directly, to ensure that only one resource can be registered per name. This is essential to ensure for calculating the tickets from quota accurately.

dalehamel:
@sirupsen this should be green again, and implements the changes we discussed

ext/semian/sysv_semaphores.c
```diff
@@ -54,26 +70,6 @@ set_semaphore_permissions(int sem_id, long permissions)
...
 int
-create_semaphore(int key, long permissions, int *created)
-{
-  int semid = 0;
-  int flags = 0;
-
-  *created = 0;
-  flags = IPC_EXCL | IPC_CREAT | permissions;
-
-  semid = semget(key, SI_NUM_SEMAPHORES, flags);
```
dalehamel: note that this is essentially the example of the [bad way to do it](http://man7.org/tlpi/code/online/diff/svsem/svsem_bad_init.c.html)

ext/semian/sysv_semaphores.c
```diff
@@ -136,8 +132,92 @@ acquire_semaphore(void *p)
...
+
+    if (semctl(sem_id, 0, IPC_STAT, sem_opts) == -1) {
+      raise_semian_syscall_error("semget()", errno);
+    }
+
+    // If a semop has been performed by someone else, the values must be initialized
+    if (sem_ds.sem_otime != 0) {
+      break;
+    }
+    sleep(1);
```
sirupsen: I think you should change this to something more like 10ms. 1 full second added to boot on deploys in a new namespace seems like a bit too much. Most of the time I bet the first worker has been done for a couple of milliseconds before anyone reaches this, but that sucks to prove—so let's just use `usleep(3)`.

ext/semian/sysv_semaphores.c
```diff
@@ -136,8 +132,92 @@ acquire_semaphore(void *p)
...
+
+#ifdef DEBUG
+    printf("Waiting for another process to initialize semaphore values, checked: %d times\n", i);
+#endif
+
+    if (semctl(sem_id, 0, IPC_STAT, sem_opts) == -1) {
+      raise_semian_syscall_error("semget()", errno);
+    }
+
+    // If a semop has been performed by someone else, the values must be initialized
```
sirupsen: Consider linking to the Linux API bible here, since it was useful to come up with this approach.

lib/semian/resource.rb
```diff
@@ -2,14 +2,20 @@ module Semian
   class Resource #:nodoc:
     attr_reader :tickets, :name
 
-    def initialize(name, tickets:, permissions: 0660, timeout: 0)
+    class << Semian::Resource
+      # Ensure that there can only be one resource of a given type
+      def instance(*args)
+        Semian.resources[args.first] ||= new(*args)
+      end
+    end
```
sirupsen: Can you explain how this makes sure there's ever only one instance?

ext/semian/tickets.c
```diff
@@ -27,68 +73,17 @@ update_ticket_count(update_ticket_count_t *tc)
...
 {
-  unsigned short init_vals[SI_NUM_SEMAPHORES];
+  int tickets = 0;
 
-  init_vals[SI_SEM_TICKETS] = init_vals[SI_SEM_CONFIGURED_TICKETS] = tickets;
-  init_vals[SI_SEM_LOCK] = 1;
-
-  if (semctl(sem_id, 0, SETALL, init_vals) == -1) {
-    raise_semian_syscall_error("semctl()", errno);
+  // Ensure that a worker for this process is registered
```
sirupsen: It's worth adding to the comment here that we from Ruby-land guarantee that a worker only goes through this code once.

ext/semian/tickets.c
```diff
@@ -1,12 +1,51 @@
...
+update_ticket_count(update_ticket_count_t *tc);
 
-static void
-configure_static_tickets(int sem_id, int tickets);
+static int
+calculate_quota_tickets(int sem_id, double quota);
 
-VALUE
+void
+configure_tickets(int sem_id, int tickets, double quota)
```
sirupsen: A comment here indicating this _must_ be called with the meta lock might be helpful. Perhaps this should even acquire the meta lock? No strong opinions on this.

ext/semian/tickets.c
```diff
@@ -16,8 +55,15 @@ update_ticket_count(update_ticket_count_t *tc)
   if (get_sem_val(tc->sem_id, SI_SEM_CONFIGURED_TICKETS) != tc->tickets) {
```
sirupsen: You already have the lock, so I don't think you need the second check. I think this is a relic from the time when the logic was: `check if wrong -> grab lock -> check if still wrong -> adjust if still wrong`. Now we just always grab the lock and only do adjustments at register-time, which is simpler.

test/resource_test.rb
```diff
@@ -323,4 +494,56 @@ def destroy_resources
...
     @resources = []
   end
+
+  # Utility function to test with multiple processes
+  # In particular, this is necessary to ensure handling of unique PIDs is correct,
+  # which is key to functionality that is utilized like SEM_UNDO
+  # Active workers are accumulated in the instance variable @workers,
+  # and workers must be cleaned up between tests by the teardown script
+  # An exit value of 100 is to keep track of timeouts, 0 for success.
+  def fork_workers(count:, resource: :testing, quota: nil, tickets: nil, timeout: 0.1, wait_for_timeout: false)
```
sirupsen: Some of the existing tests use `Thread.start` and not fork, we need to settle on one pattern I think unless we're testing specific edge-cases. This somewhat relates to my comment about this not working for threads. I see you already consolidated some of the tests, which is great—but I think we need to do it for the thread tests too.. maybe?

ext/semian/tickets.c
```diff
@@ -27,68 +73,17 @@ update_ticket_count(update_ticket_count_t *tc)
...
-  unsigned short init_vals[SI_NUM_SEMAPHORES];
+  int tickets = 0;
 
-  init_vals[SI_SEM_TICKETS] = init_vals[SI_SEM_CONFIGURED_TICKETS] = tickets;
-  init_vals[SI_SEM_LOCK] = 1;
-
-  if (semctl(sem_id, 0, SETALL, init_vals) == -1) {
-    raise_semian_syscall_error("semctl()", errno);
+  // Ensure that a worker for this process is registered
+  if (perform_semop(sem_id, SI_SEM_REGISTERED_WORKERS, 1, SEM_UNDO, NULL) == -1) {
```
sirupsen: We need to add a way to decrement this without `SEM_UNDO`, e.g. as part of destroying the resource—otherwise this will never work in threaded environments. We need to add a comment about this too, that if you're in a threaded implementation you _must_ unregister yourself.

test/resource_test.rb
sirupsen: Overall, I really like this approach for testing. Makes a ton of sense.

test/resource_test.rb
```diff
@@ -323,4 +494,56 @@ def destroy_resources
...
+            # Hold the resource until signalled
+            # This helps to avoid race conditions in testing.
+            Signal.trap('TERM') do
+              yield if block_given?
+              exit! 0
+            end
+            sleep
+          end
+        rescue Semian::TimeoutError
+          Signal.trap('TERM') do
```
sirupsen: I don't see any reason why you can't just `exit!` here. The test needs to `wait` for it anyway.

test/resource_test.rb
```diff
@@ -133,6 +176,119 @@ def test_acquire_with_fork
...
+
+    fork_workers(count: workers, quota: quota, wait_for_timeout: true)
+    fork_workers(count: added_workers, quota: new_quota, wait_for_timeout: true)
+
+    # Allow all processes to terminate
+    signal_workers('TERM')
+    es = Process.waitall
+
+    # Ensure the correct number of processes timed out, based on the *new* quota
+    actual_timeouts = es.count { |s| s.last.exitstatus == 100 }
```
sirupsen: It might be worth writing a helper for the `Process.waitall` and checking the status. I don't think it needs comments.

test/resource_test.rb
```diff
@@ -133,6 +176,119 @@ def test_acquire_with_fork
...
+    expected_timeouts = ((1 - new_quota) * (workers + added_workers)).floor
+    assert_equal expected_timeouts, actual_timeouts
+  end
+
+  def test_quota_decrease
```
sirupsen: Great test.

test/resource_test.rb
```diff
@@ -133,6 +176,119 @@ def test_acquire_with_fork
...
+
+    # Ensure the correct number of processes timed out
+    timeouts = Process.waitall.count { |s| s.last.exitstatus == 100 }
+    assert_equal workers - ((1 - quota) * workers).ceil, timeouts
+
+    # Ensure that the tickets were released
+    assert_equal resource.tickets, resource.count
+  end
+
+  def test_quota_increase
```
sirupsen: I would expect this test and the decrease test to be much more alike, why aren't they? It seems to me it's because the `decrease` takes the shortcut of only being about 1 worker, but I think they should be pretty much mirrors of each other for simplicity when reading the test suite.

test/resource_test.rb
```diff
@@ -133,6 +176,119 @@ def test_acquire_with_fork
...
+
+    assert_equal((workers * new_quota).ceil, resource.tickets)
+  end
+
+  def test_quota_sets_tickets_from_workers
+    quota = 0.5
+    workers = 50
+
+    resource = create_resource :testing, quota: quota, timeout: 0.1
+    fork_workers(count: workers - 1, quota: quota, wait_for_timeout: true)
```
sirupsen: Why `- 1` in this test? This should probably be the first test in the `quota` family as it checks the most basic functionality as far as I can see.

test/resource_test.rb
```diff
@@ -133,6 +176,119 @@ def test_acquire_with_fork
...
+  end
+
+  def test_quota_adjust_tickets_on_new_workers
+    quota = 0.5
+    workers = 50
+
+    resource = create_resource :testing, quota: quota, timeout: 0.1
+
+    # Spawn some workers to get a basis for the quota
+    fork_workers(count: workers - 1, quota: quota, wait_for_timeout: true)
```
sirupsen: Again, why `-1`?

test/resource_test.rb
```diff
@@ -133,6 +176,119 @@ def test_acquire_with_fork
...
+
+    # Signal and wait for the workers to quit
+    signal_workers('KILL')
+    Process.waitall
+
+    # Number of tickets should be unchanged until next acquire
+    assert_equal((workers * quota).ceil, resource.tickets)
+
+    resource = create_resource :testing, quota: quota, timeout: 0.1
+    assert_equal 1, resource.tickets
```
sirupsen: GREAT test!

test/resource_test.rb
```diff
@@ -9,6 +9,8 @@ def setup
 
   def teardown
```
sirupsen: Overall, I'm pretty happy with the test suite here! Good job. I honestly can't really think of other test cases to do... that's rare :)

sirupsen:
> How should small ticket counts be address during, for instance, rolling deploys? Should a minimum ticket count be specified?

My concern here is that we may be doing some more I/O at the beginning, but on the other hand, I can't really see why that would be either... if we're careful about the rollout and test a siege on sb-perf while deploying, I think that's enough to prove this is not a problem.

> How should 1 registration per worker be enforced?

It seems you took a stab at this, I left a comment on you explaining it. Then we can take it from there.

> What should the bounds of a quota value be? Should a quota of 1 be allowed?

I think we should keep the quota of 1. It's a way to no-op it to roll it out to production, which I think is a good thing.

ext/semian/tickets.c
dalehamel: @sirupsen true, but I don't think there's any way to enforce this from C, and this is the same situation we have right now.

I suppose we should add a warning to the README that the quota workers with threads won't cleanly handle unregistering.

lib/semian/resource.rb
dalehamel: Did you mean in the comments, or here?

I'll do it here for now.

Because we call instance instead of new, we can detect if a worker for a given resource has already been registered, and provide that instead of creating a new semian resource.

If that resource is deleted or never existed,  a new one would be created instead.

test/resource_test.rb
dalehamel: This is because of the worker that gets registered for the test process itself (the line above). To make the math simpler, we subtract that from the total number of workers to fork.

Is there another way you can think of to do this more intuitively?

test/resource_test.rb
dalehamel: Above^

test/resource_test.rb
```diff
@@ -133,6 +176,119 @@ def test_acquire_with_fork
...
+
+    # Spawn some workers to get a basis for the quota
+    fork_workers(count: workers - 1, quota: quota, wait_for_timeout: true)
+    assert_equal((workers * quota).ceil, resource.tickets)
+
+    # Signal and wait for the workers to quit
+    signal_workers('KILL')
+    Process.waitall
+
+    # Number of tickets should be unchanged until next acquire
```
dalehamel: FIXME - comment is wrong

test/resource_test.rb
dalehamel: @sirupsen I only consolidated the test(s) that were doing forking to use this approach to be more coherent. 

I left anything doing threading as-is.

test/resource_test.rb
dalehamel: I believe I had a reason for this but I can't remember what it is or if it is still valid...

ext/semian/tickets.c
dalehamel: yep good catch, 100% correct here.

test/resource_test.rb
sirupsen: Don't worry about it then :)

ext/semian/sysv_semaphores.c
dalehamel: Done earlier in the code path

ext/semian/sysv_semaphores.h
```diff
@@ -97,4 +106,17 @@ get_semaphore(int key);
 void *
 acquire_semaphore_without_gvl(void *p);
 
+#ifdef DEBUG
```
dalehamel: I kind of want to keep this definition after removing the debug prints, as its very useful for debugging.

test/resource_test.rb
dalehamel: @sirupsen I remember why now. This is needed in order to reason about the number of *registered* workers in a reasonable way.

test/resource_test.rb
dalehamel: @sirupsen you're quite right.

I think these were really testing two different things here, so I added additional tests and clarified the names:

* test_quota_increase / test_quota_decrease - just spawn a single worker to mutate the quota and make assertions about the ticket count changes. Added the test_quota_increase, to mirror test_quota_decrease.
* test_quota_acquire_increase_quota / test_quota_acquire_decrease_quota - these also take into account how the quota adjustment should affect acquire and timeouts. Renamed test_quota_increase to test_quota_acquire_increase_quota, and wrote test_quota_acquire_decrease_quota to mirror it.

ext/semian/resource.c
```diff
@@ -63,9 +69,20 @@ semian_resource_acquire(int argc, VALUE *argv, VALUE self)
...
 {
+  struct timespec ts = { 0 };
+  ts.tv_sec = INTERNAL_TIMEOUT;
+
   semian_resource_t *res = NULL;
 
   TypedData_Get_Struct(self, semian_resource_t, &semian_resource_type, res);
+
+  // Prevent a race to deletion
+  if (perform_semop(res->sem_id, SI_SEM_LOCK, -1, 0, &ts) == -1) {
```
dalehamel: This is equivalent to sem_meta_lock, but we want to allow EINVAL as that will show us that the resource has already been deleted.

test/resource_test.rb
```diff
@@ -198,18 +358,25 @@ def test_destroy
     end
   end
 
+  def test_destroy_already_destroyed
```
dalehamel: @sirupsen this is the regression test I added to ensure that there is no race to destroy the resource.

dalehamel:
@sirupsen all comments have been addressed, PTAL again :eyes: 

Also, I found and fixed a bug related to multiple deletes of the same resource.

README.md
```diff
@@ -127,6 +127,64 @@ client = Redis.new(semian: {
 })
 ```
 
+#### Warning: Thread unsafe
+
+Semian is currently known to be thread-unsafe, and is intended to be used by *separate processes*.
+
+Internally, semian uses ```SEM_UNDO``` for several sysv semaphore operations:
```
sirupsen: Single backticks are adequate here, could you use single backticks around other mentions of `SEM_UNDO` for completeness?

README.md
```diff
@@ -127,6 +127,64 @@ client = Redis.new(semian: {
...
+
+* Acquire
+* Worker registration
+* Semaphore metadata state lock
+
+The intention behind SEM_UNDO is that a semaphore operation is automatically undone when the process exits.
```
sirupsen: When a process exits abnormally, may be worth mentioning, i.e. `SIGKILL`.

README.md
```diff
@@ -127,6 +127,64 @@ client = Redis.new(semian: {
...
+
+If, however, a thread performs a semop, the SEM_UNDO is on its parent process. This means that the operation
```
sirupsen: `semop` is somewhat internal vocabulary.

README.md
```diff
@@ -127,6 +127,64 @@ client = Redis.new(semian: {
...
+* Threads that register workers on a resource but are killed and never unregistered. For a process, the worker
+count would be automatically decremented by SEM_UNDO, but for threads the worker count will continue to increment,
+only being undone when the parent process dies. This can cause the number of tickets to dramatically exceed the quota.
+* If a thread acquires the semaphore metadata lock and dies before releasing it, semian will deadlock on anything
+attempting to acquire the metadata lock until the thread's parent process exits. This can prevent the ticket count
+from being updated.
+
+Moreover, a strategy that utilizes SEM_UNDO is not compatible with a strategy that attempts to the semaphores tickets manually.
+In order to support threads, operations that currently use SEM_UNDO would need to use no semaphore flag, and the calling process
+will be responsible for ensuring that threads are appropriately cleaned up.
```
sirupsen: I'd add a note on supporting threads through an in-memory semaphore, or, that it's possibly by creating another API—and that we'll welcome PRs.

ext/semian/sysv_semaphores.c
```diff
@@ -136,8 +136,92 @@ acquire_semaphore(void *p)
...
+generate_key(const char *name)
+{
+  char semset_size_key[20];
+  char *uniq_id_str;
+
+  // It is necessary for the cardinatily of the semaphore set to be part of the key
+  // or else sem_get will complain that we have requested an incorrect number of sems
+  // for the desired key, and have changed the number of semaphores for a given key
+  sprintf(semset_size_key, "_NUM_SEMS_%d", SI_NUM_SEMAPHORES);
+  uniq_id_str = malloc(strlen(name)+strlen(semset_size_key)+1);
```
sirupsen: Why do you need to allocate this on the heap instead of using the stack for this temporary string?

dalehamel: @sirupsen this was because I don't know for certain what the size of the key string is going to be.

We could make the assumption about the size of *name, but then we have the potential risk of buffer overflow.

Since we delete it here, and it's only done once during init, I think the malloc is safer than hardcoded assumptions about the buffer.

README.md
dalehamel: So is SEM_UNDO, and semop directs them to the correct library call.

README.md
dalehamel: agree

README.md
dalehamel: done

README.md
dalehamel: agree, added.

PR #123, opened 2017-02-23
Break the coupling between the protected resource and circuit breaker

### LINKED ISSUE #95: Opt out of ticket count feature
### What
It would be good to be able to configure semian without having to specify a ticket count, or to be able to explicitly disable the ticket count feature altogether.

### Why
You might not want to limit the number of concurrent requests to a resource. This is the case in shopify-app-store, where we don’t want to limit the number of concurrent requests to the Shopify API.

BoGs:
# What
Breaking the coupling of ProtectedResource from circuit breaker and bulkhead. To allow the ability to request a circuit_breaker or bulkhead independent of each other without breaking the path of processing. 

Fixes #95 

# How

The protected resource will be registered but will only contain the resources that have been requested. Given you do not want a resource the code path will not execute and will allow the ability to interact based on the properties given in `acquire` or `register`.

### DIFF
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -116,7 +116,11 @@ module Semian
 
   # Registers a resource.
   #
-  # +name+: Name of the resource - this can be either a string or symbol.
+  # +name+: Name of the resource - this can be either a string or symbol. (required)
+  #
+  # +circuit_breaker+: The boolean if you want a circuit breaker acquired for your resource. Default true.
+  #
+  # +bulkhead+: The boolean if you want a bulkhead to be acquired for your resource. Default true.
   #
   # +tickets+: Number of tickets. If this value is 0, the ticket count will not be set,
   # but the resource must have been previously registered otherwise an error will be raised.
@@ -126,31 +130,29 @@ module Semian
   # Must be greater than 0, less than or equal to 1. There will always be at least 1 ticket, as it
   # is calculated as (workers * quota).ceil
   # Mutually exclusive with the 'ticket' argument.
+  # but the resource must have been previously registered otherwise an error will be raised. (bulkhead)
   #
-  # +permissions+: Octal permissions of the resource.
+  # +permissions+: Octal permissions of the resource. Default 0660. (bulkhead)
   #
-  # +timeout+: Default timeout in seconds.
+  # +timeout+: Default timeout in seconds. Default 0. (bulkhead)
   #
-  # +error_threshold+: The number of errors that will trigger the circuit opening.
+  # +error_threshold+: The number of errors that will trigger the circuit opening. (circuit breaker required)
   #
   # +error_timeout+: The duration in seconds since the last error after which the error count is reset to 0.
+  # (circuit breaker required)
   #
   # +success_threshold+: The number of consecutive success after which an half-open circuit will be fully closed.
+  # (circuit breaker required)
   #
-  # +exceptions+: An array of exception classes that should be accounted as resource errors.
+  # +exceptions+: An array of exception classes that should be accounted as resource errors. Default [].
+  # (circuit breaker)
   #
   # Returns the registered resource.
-  def register(name, tickets: nil, quota: nil, permissions: 0660, timeout: 0, error_threshold:, error_timeout:, success_threshold:, exceptions: [])
-    circuit_breaker = CircuitBreaker.new(
-      name,
-      success_threshold: success_threshold,
-      error_threshold: error_threshold,
-      error_timeout: error_timeout,
-      exceptions: Array(exceptions) + [::Semian::BaseError],
-      implementation: ::Semian::Simple,
-    )
-    resource = Resource.instance(name, tickets: tickets, quota: quota, permissions: permissions, timeout: timeout)
-    resources[name] = ProtectedResource.new(resource, circuit_breaker)
+  def register(name, **options)
+    circuit_breaker = create_circuit_breaker(name, **options)
+    bulkhead = create_bulkhead(name, **options)
+
+    resources[name] = ProtectedResource.new(bulkhead, circuit_breaker)
   end
 
   def retrieve_or_register(name, **args)
@@ -171,6 +173,37 @@ module Semian
   # Retrieves a hash of all registered resources.
   def resources
     @resources ||= {}
+  end
+
+  private
+
+  def create_circuit_breaker(name, **options)
+    circuit_breaker = options.fetch(:circuit_breaker, true)
+    return unless circuit_breaker
+    raise ArgumentError unless required_keys?([:success_threshold, :error_threshold, :error_timeout], options)
+
+    exceptions = options[:exceptions] || []
+    CircuitBreaker.new(
+      name,
+      success_threshold: options[:success_threshold],
+      error_threshold: options[:error_threshold],
+      error_timeout: options[:error_timeout],
+      exceptions: Array(exceptions) + [::Semian::BaseError],
+      implementation: ::Semian::Simple,
+    )
+  end
+
+  def create_bulkhead(name, **options)
+    bulkhead = options.fetch(:bulkhead, true)
+    return unless bulkhead
+
+    permissions = options[:permissions] || 0660
+    timeout = options[:timeout] || 0
+    Resource.new(name, tickets: options[:tickets], quota: options[:quota], permissions: permissions, timeout: timeout)
+  end
+
+  def required_keys?(required = [], **options)
+    required.all? { |key| options.key? key }
   end
 end
 
--- a/lib/semian/protected_resource.rb
+++ b/lib/semian/protected_resource.rb
@@ -2,34 +2,56 @@ module Semian
   class ProtectedResource
     extend Forwardable
 
-    def_delegators :@resource, :destroy, :count, :semid, :tickets, :name
+    def_delegators :@bulkhead, :destroy, :count, :semid, :tickets, :name
     def_delegators :@circuit_breaker, :reset, :mark_failed, :mark_success, :request_allowed?,
                    :open?, :closed?, :half_open?
 
-    def initialize(resource, circuit_breaker)
-      @resource = resource
+    attr_reader :bulkhead, :circuit_breaker
+
+    def initialize(bulkhead, circuit_breaker)
+      @bulkhead = bulkhead
       @circuit_breaker = circuit_breaker
     end
 
     def destroy
-      @resource.destroy
-      @circuit_breaker.destroy
+      @bulkhead.destroy unless @bulkhead.nil?
+      @circuit_breaker.destroy unless @circuit_breaker.nil?
     end
 
     def acquire(timeout: nil, scope: nil, adapter: nil)
-      @circuit_breaker.acquire do
-        begin
-          @resource.acquire(timeout: timeout) do
-            Semian.notify(:success, self, scope, adapter)
-            yield self
-          end
-        rescue ::Semian::TimeoutError
-          Semian.notify(:busy, self, scope, adapter)
-          raise
+      acquire_circuit_breaker(scope, adapter) do
+        acquire_bulkhead(timeout, scope, adapter) do
+          yield self
+        end
+      end
+    end
+
+    private
+
+    def acquire_circuit_breaker(scope, adapter)
+      if @circuit_breaker.nil?
+        yield self
+      else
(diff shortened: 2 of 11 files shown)


lib/semian/protected_resource.rb
```diff
@@ -2,35 +2,59 @@ module Semian
...
     extend Forwardable
 
-    def_delegators :@resource, :destroy, :count, :semid, :tickets, :name
+    def_delegators :@bulkhead, :destroy, :count, :semid, :tickets, :name
     def_delegators :@circuit_breaker, :reset, :mark_failed, :mark_success, :request_allowed?,
                    :open?, :closed?, :half_open?
 
-    def initialize(resource, circuit_breaker)
-      @resource = resource
+    attr_reader :bulkhead, :circuit_breaker
```
BoGs: I am not _super_ happy with this not sure how to expose and support the nil assertions. Thoughts?

lib/semian/protected_resource.rb
```diff
@@ -2,35 +2,59 @@ module Semian
   class ProtectedResource
     extend Forwardable
 
-    def_delegators :@resource, :destroy, :count, :semid, :tickets, :name
+    def_delegators :@bulkhead, :destroy, :count, :semid, :tickets, :name
```
BoGs: Changed `resource` -> `bulkhead` the resource resource was confusing for me.

lib/semian/unprotected_resource.rb
```diff
@@ -6,6 +6,8 @@ class UnprotectedResource
 
     def initialize(name)
       @name = name
+      @bulkhead = nil
```
jpittis: I believe both of these initializations are unnecessary? Are `@name` and `@bulkhead` `nil` by default?

lib/semian/protected_resource.rb
jpittis: I personally like this. I found this confusing too. The only issue is that in the C code, it's still referenced as `Resource`. It seems unnecessary to change both though.

test/protected_resource_test.rb
```diff
@@ -0,0 +1,90 @@
...
+    Semian.destroy(:testing)
+  rescue
+    nil
+  end
+
+  def teardown
+    destroy_resources
+  end
+
+  def test_acquire_without_bulkhead
```
jpittis: The test says it acquires without bulkhead but I don't see an acquire call in the test.

lib/semian/protected_resource.rb
```diff
@@ -2,35 +2,59 @@ module Semian
...
     end
 
     def destroy
-      @resource.destroy
-      @circuit_breaker.destroy
+      @bulkhead.destroy unless @bulkhead.nil?
+      @circuit_breaker.destroy unless @circuit_breaker.nil?
     end
 
     def acquire(timeout: nil, scope: nil, adapter: nil)
```
jpittis: Is there some way you could put the check for `@circuit_breaker` and `@bulkhead` being nil in the respective `acquire_<blank>` helpers. I'm not sure if that would be cleaner or not.

lib/semian.rb
```diff
@@ -166,6 +167,32 @@ def destroy(name)
...
+  def create_circuit_breaker(name, **options)
+    circuit_breaker = options.fetch(:circuit_breaker, true)
+    return unless circuit_breaker
+
+    exceptions = options[:exceptions] || []
+    CircuitBreaker.new(
+      name,
+      success_threshold: options[:success_threshold],
+      error_threshold: options[:error_threshold],
+      error_timeout: options[:error_timeout],
```
sirupsen: Previously we'd raise from `#register` if these options weren't given (they have an empty value). Can we raise an `ArgumentError` if `circuit_breaker` is passed, but the options are not given?

lib/semian.rb
```diff
@@ -166,6 +167,32 @@ def destroy(name)
...
+    )
+  end
+
+  def create_bulkhead(name, **options)
+    bulkhead = options.fetch(:bulkhead, true)
+    return unless bulkhead
+
+    permissions = options[:permissions] || 0660
+    timeout = options[:timeout] || 0
+    Resource.new(name, tickets: options[:tickets], permissions: permissions, timeout: timeout)
```
sirupsen: Same thing goes here, we need to raise if `tickets` are not defined.

lib/semian/protected_resource.rb
sirupsen: I agree, but this will break backwards compatibility—I'm afraid we'll have to alias and deprecate this.

lib/semian/protected_resource.rb
```diff
@@ -2,35 +2,59 @@ module Semian
...
+        acquire_circuit_breaker(scope, adapter) do
+          acquire_bulkhead(timeout, scope, adapter) do
             yield self
           end
-        rescue ::Semian::TimeoutError
-          Semian.notify(:busy, self, scope, adapter)
-          raise
         end
       end
+    end
```
sirupsen: If you return `yield` in `acquire_*` if it's not enabled, you might be able to do something like this:

```ruby
def acquire(timeout: nil, scope: nil, adapter: nil, &block)
  [method(:acquire_bulkhead), method(:acquire_circuit_breaker)].inject(block) { |nested, method| 
    ->(scope, adapter) { method.call(scope, adapter, &nested) } 
  }.call(scope, adapter)
end 
```

The idea is to build a proc that properly nests the calls, with bulkhead first (note it'll build it reverse from what you might expect).

lib/semian/unprotected_resource.rb
sirupsen: Yeah, don't think this is necessary either.

test/protected_resource_test.rb
```diff
@@ -0,0 +1,90 @@
...
+  def test_acquire_without_circuit_breaker
+    Semian.register(
+      :testing,
+      tickets: 2,
+      circuit_breaker: false,
+    )
+    acquired = false
+
+    @resource = Semian[:testing]
+    @resource.acquire do
```
sirupsen: This doesn't actually test that the circuit breaker isn't used from an integration point of view. Perhaps you should try to do 10 failures and make sure you can still acquire it, or something like this?

test/protected_resource_test.rb
sirupsen: Yeah, this test doesn't seem to do this. I'd suggest trying to make this be acquired from, say, 10 processes in parallel to make sure bulkheads are _not_ enabled.

test/protected_resource_test.rb
```diff
@@ -0,0 +1,90 @@
...
+      acquired = true
+      assert_equal 1, @resource.count
+      assert_equal 2, @resource.tickets
+    end
+
+    assert acquired
+    assert_nil @resource.circuit_breaker
+  end
+
+  def test_acquire
```
sirupsen: What does this test?

lib/semian/unprotected_resource.rb
BoGs: Yea this was left from a previous change. Removed this.

lib/semian/protected_resource.rb
BoGs: @sirupsen the public interface has not changed as much as I am aware - the delegator methods are still the same and `@resource` was never accessible. I can alias resource to bulkhead just not sure its needed.

lib/semian/protected_resource.rb
BoGs: not sure ttyt if we wanted to keep the logic together it would make sense but pollute the `acquire_*` methods with multiple yields - ex. below 

```
    def acquire(timeout: nil, scope: nil, adapter: nil)
      acquire_circuit_breaker(scope, adapter) do
        acquire_bulkhead(timeout, scope, adapter) do
          yield self
        end
      end
    end

    private

    def acquire_circuit_breaker(scope, adapter)
      yield self if @circuit_breaker.nil?

      @circuit_breaker.acquire do
        yield self
      end
    rescue ::Semian::OpenCircuitError
      Semian.notify(:circuit_open, self, scope, adapter)
      raise
    end

    def acquire_bulkhead(timeout, scope, adapter)
      yield self if @bulkhead.nil?

      @bulkhead.acquire(timeout: timeout) do
        Semian.notify(:success, self, scope, adapter)
        yield self
      end
    rescue ::Semian::TimeoutError
      Semian.notify(:busy, self, scope, adapter)
      raise
    end
```

thoughts?

BoGs: The more I look at it the more I like it 😄

BoGs: ^^ the code above would need an if @bulkhead.nil? yield self and return - not sure if cleaner :)

BoGs:
```
    def acquire_circuit_breaker(scope, adapter)
      if @circuit_breaker.nil?
        yield self
        return
      end

      @circuit_breaker.acquire do
        yield self
      end
    rescue ::Semian::OpenCircuitError
      Semian.notify(:circuit_open, self, scope, adapter)
      raise
    end
```

lib/semian/protected_resource.rb
BoGs: I moved the check for null in the `acquire_*` methods the problem is that if you define `&block` it expects the previous method parameters from my testing. I changed the method to be the same print as before.

BoGs:
@sirupsen @jpittis can you guys take 👀 another - I cleaned up some tests, created some helpers, renamed some tests to explain things better.

test/helpers/circuit_breaker_helper.rb
```diff
@@ -0,0 +1,36 @@
+module CircuitBreakerHelper
+  SomeError = Class.new(StandardError)
+
+  private
+
+  def open_circuit!(resource = @resource)
+    2.times { trigger_error!(resource) }
```
sirupsen: You should consider using the config to check what the open threshold is.

test/helpers/circuit_breaker_helper.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def half_open_cicuit!(resource = @resource)
+    Timecop.travel(Time.now - 10) do
```
sirupsen: Again, you'll need to use the config.

test/protected_resource_test.rb
```diff
@@ -0,0 +1,101 @@
...
+
+    10.times do
+      Thread.new do
+        block_called = false
+        @resource = Semian[:testing]
+        @resource.acquire { block_called = true }
+        assert_equal true, block_called
+        assert_instance_of Semian::CircuitBreaker, @resource.circuit_breaker
+        assert_nil @resource.bulkhead
+      end.join
```
sirupsen: This won't do them in parallel, so it won't check the circuit. You'll need to spawn them in parallel, add them to an array, do `threads.each(&:join)`.

test/protected_resource_test.rb
```diff
@@ -0,0 +1,101 @@
...
+  end
+
+  def test_register_with_bulkhead_missing_options
+    assert_raises ArgumentError do
+      Semian.register(
+        :testing,
+        circuit_breaker: false,
+      )
+    end
+  end
```
sirupsen: These argument tests seem more appropriate in `semian_test.rb`, as they aren't specific to the protected resource.

test/protected_resource_test.rb
```diff
@@ -0,0 +1,101 @@
...
+
+    acquired = false
+
+    @resource = Semian[:testing]
+    @resource.acquire do
+      acquired = true
+      assert_equal 1, @resource.count
+      assert_equal 2, @resource.tickets
+      half_open_cicuit!(@resource)
+      assert_circuit_closed(@resource)
```
sirupsen: What's the intention of these two lines?

test/protected_resource_test.rb
```diff
@@ -0,0 +1,101 @@
...
+  def test_register_with_bulkhead_missing_options
+    assert_raises ArgumentError do
+      Semian.register(
+        :testing,
+        circuit_breaker: false,
+      )
+    end
+  end
+
+  def test_acquire_bulkhead_without_circuit_breaker
```
sirupsen: This doesn't test that the circuit breaker is not present other than checking the value. I don't really have a strong opinion on these tests, but they seem to have a bit of an identity crisis between checking what's enabled and integration testing. I think I suggested the latter, but that would assume you could use some helpers or something from the implementation tests.

PR #124, opened 2017-03-14
Add support for a minimum number of tickets when using the quota strategy

dalehamel:
# What

Allow for a minimum number of tickets to still be allocated while using quotas to provide tickets.

# Why

Since the quota strategy makes the number of tickets available proportional to the number of workers,
it's possible that the number of tickets may be very constrained during some normal operations,
such as rolling restarts during, for instance, a deploy.

To prevent these situations from resulting in a large number of tripped circuit breakers,
a minimum number of tickets can be specified.

The actual ticket count used will be the max of the tickets calculated by the quota, and the
minimum value.

# How

We simply pass along a min_ticket value, and when we calculate the quota tickets if the quota value would be less than min_tickets, we use that instead.

### DIFF
--- a/ext/semian/resource.c
+++ b/ext/semian/resource.c
@@ -12,6 +12,9 @@ check_quota_arg(VALUE quota);
 static int
 check_tickets_arg(VALUE tickets);
 
+static int
+check_min_tickets_arg(VALUE tickets);
+
 static long
 check_permissions_arg(VALUE permissions);
 
@@ -129,18 +132,20 @@ semian_resource_id(VALUE self)
 }
 
 VALUE
-semian_resource_initialize(VALUE self, VALUE id, VALUE tickets, VALUE quota, VALUE permissions, VALUE default_timeout)
+semian_resource_initialize(VALUE self, VALUE id, VALUE tickets, VALUE quota, VALUE min_tickets, VALUE permissions, VALUE default_timeout)
 {
   long c_permissions;
   double c_timeout;
   double c_quota;
   int c_tickets;
+  int c_min_tickets;
   semian_resource_t *res = NULL;
   const char *c_id_str = NULL;
 
   // Check and cast arguments
   check_tickets_xor_quota_arg(tickets, quota);
   c_quota = check_quota_arg(quota);
+  c_min_tickets = check_min_tickets_arg(min_tickets);
   c_tickets = check_tickets_arg(tickets);
   c_permissions = check_permissions_arg(permissions);
   c_id_str = check_id_arg(id);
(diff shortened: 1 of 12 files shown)


ext/semian/resource.c
```diff
@@ -234,6 +238,25 @@ check_tickets_arg(VALUE tickets)
   return c_tickets;
 }
 
+static int
+check_min_tickets_arg(VALUE min_tickets)
```
dalehamel: Warning: copy pasta

ext/semian/tickets.h
```diff
@@ -6,8 +6,13 @@ For logic specific to manipulating semian ticket counts
 
 #include "sysv_semaphores.h"
 
+#ifndef max
+// max is not defined in any standard, portable C header...
+#define max(a,b) (((a)>(b))?(a):(b))
```
dalehamel: believe it or not, in C this isn't any any standard library....

lib/semian.rb
```diff
@@ -127,6 +127,12 @@ def to_s
   # is calculated as (workers * quota).ceil
   # Mutually exclusive with the 'ticket' argument.
   #
+  # +min_tickets+: When using the quota strategy, never allocated fewer tickets than this value.
+  # This is particularly useful during deploys, where the number of tickets will be constrained
+  # due to the number of workers going offline. To prevent spurious timeouts, you may set a minimum
+  # ticket count. The resulting ticket count will be max(min_ticket_count, calculated_quota_tickets).
+  # This option is ignored unless you are using the quota strategy.
```
dalehamel: I think this is the best behavior, let me know if you disagree.

ext/semian/resource.c
sirupsen: Should we have a default minimum? I'm not sure if it would ever make sense to have less than, say, 3 tickets? What are you planning on configuring this to for our use-case?

lib/semian.rb
sirupsen: Should we call this `quota_min_tickets` to make extra certain it's only relevant for that strategy? It you think it's better, go ahead. Do you want to raise if `min_tickets` is set to `> 0` and `tickets` is set directly as well?

dalehamel: quota_min_tickets is a good idea.

I think it's also a good idea to raise as you said, will do.

ext/semian/resource.c
dalehamel: I don't think we should have a default minimum, I think that should be manually specified by the user.

I think that for our cases, a minimum of something like 5-10 is probably good, but it needs to be examined on a per-resource basis.

dalehamel:
I'm going to investigate an alternate approach that will attempt to account for boot cases without forcing this additional complexity.

PR #125, opened 2017-03-14
Allow a boot grace period using quota strategy

dalehamel:
# What

Allow for quota workers to pad the acquire timeout while workers are booting.

# Why

There is the possibility of timeouts when using the quota strategy while
workers are booting, since there are a small number of workers.

We can prevent spurious timeouts caused by a small number of tickets
during this period, by allowing for the acquire timeout to be padded.

# How

We store the desired padding value as part of the resource structure.

Whenever the ticket count is updated (workers are registered), a SETVAL occurs that will write the ctime of the semaphore.

We can compare this value with the current time to see if we are within the QUOTA_BOOT_GRACE_PERIOD. If we are, we know to replace any configured timeout value with our padded value.

### DIFF
--- a/ext/semian/resource.c
+++ b/ext/semian/resource.c
@@ -25,6 +25,9 @@ check_default_timeout_arg(VALUE default_timeout);
 static double
 check_quota_grace_timeout_arg(VALUE quota_grace_timeout);
 
+static double
+check_quota_grace_period_arg(VALUE quota_grace_timeout);
+
 static void
 ms_to_timespec(long ms, struct timespec *ts);
 
@@ -69,9 +72,8 @@ semian_resource_acquire(int argc, VALUE *argv, VALUE self)
   // we may have spurious timeouts due to a small number of registered workers.
   if(res.quota > 0) {
     semaphore_stat(res.sem_id, &sem_ds);
-    if (difftime(time(0), sem_ds.sem_ctime) < QUOTA_BOOT_GRACE_PERIOD) {
-      res.timeout.tv_sec = res.quota_grace_timeout;
-      res.timeout.tv_nsec = 0;
+    if (difftime(time(0), sem_ds.sem_ctime) < res.quota_grace_period) {
+      ms_to_timespec(res.quota_grace_timeout * 1000, &res.timeout);
     }
   }
 
@@ -155,11 +157,12 @@ semian_resource_id(VALUE self)
 }
 
 VALUE
-semian_resource_initialize(VALUE self, VALUE id, VALUE tickets, VALUE quota, VALUE permissions, VALUE default_timeout, VALUE quota_grace_timeout)
+semian_resource_initialize(VALUE self, VALUE id, VALUE tickets, VALUE quota, VALUE permissions, VALUE default_timeout, VALUE quota_grace_timeout, VALUE quota_grace_period)
 {
   long c_permissions;
   double c_timeout;
   double c_quota_grace_timeout;
+  double c_quota_grace_period;
   double c_quota;
   int c_tickets;
   semian_resource_t *res = NULL;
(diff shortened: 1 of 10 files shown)


ext/semian/resource.c
```diff
@@ -53,6 +60,21 @@ semian_resource_acquire(int argc, VALUE *argv, VALUE self)
     rb_raise(rb_eArgError, "invalid arguments");
   }
 
+  // Backup the timeout values in case we need to revert them
+  tv_sec = res.timeout.tv_sec;
```
dalehamel: We need to do this or else we could be permanently overriding the timeout value

ext/semian/resource.c
```diff
@@ -63,6 +85,10 @@ semian_resource_acquire(int argc, VALUE *argv, VALUE self)
     }
   }
 
+  // Restore the timeout values, in case they were mutated
+  res.timeout.tv_sec = tv_sec;
```
dalehamel: In most cases this will be a noop, but we need this to potentially restore the backed-up value above.

dalehamel:
@sirupsen i suspect the naming here needs some re-jiggering to make it clearer, open to ideas.

ext/semian/resource.h
```diff
@@ -9,6 +9,9 @@ Functions here are associated with rubyland operations.
 #include "types.h"
 #include "sysv_semaphores.h"
 
+// Time to wait for workers to boot using the quota strategy
+#define QUOTA_BOOT_GRACE_PERIOD 60 // seconds
```
dalehamel: Is there anything so slow that 60 seconds won't be enough? :/

ext/semian/sysv_semaphores.c
```diff
@@ -187,15 +187,25 @@ initialize_new_semaphore_values(int sem_id, long permissions)
 #endif
 }
 
+void
+semaphore_stat(int sem_id, struct semid_ds *sem_ds)
```
dalehamel: refactored this since we're doing it in a couple places now.

ext/semian/tickets.c
```diff
@@ -66,6 +66,7 @@ update_ticket_count(update_ticket_count_t *tc)
     }
   }
 
+  // This will update the ctime
```
dalehamel: this is very important to understand, as it's the magic that makes this all work.

ext/semian/resource.c
```diff
@@ -53,6 +60,21 @@ semian_resource_acquire(int argc, VALUE *argv, VALUE self)
...
+  tv_nsec = res.timeout.tv_nsec;
+
+  // If we recently booted, and are using the quota strategy
+  // we need to give the workers some time to register or else
+  // we may have spurious timeouts due to a small number of registered workers.
+  if(res.quota > 0) {
+    semaphore_stat(res.sem_id, &sem_ds);
+    if (difftime(time(0), sem_ds.sem_ctime) < QUOTA_BOOT_GRACE_PERIOD) {
+      res.timeout.tv_sec = res.quota_grace_timeout;
+      res.timeout.tv_nsec = 0;
```
sirupsen: Maybe we should use the `ms_to_timespec` helper to support doubles? I think right now that wouldn't work, e.g. if you passed `55.5`, it'd probably just do `56` since `tv_sec` probably has to be an `int`?

ext/semian/resource.h
sirupsen: This is basically around the boot time of the workers, which could, with Ruby on Rails, not unreasonably be 30-90s. I don't know how easy it would be to make this configurable (probably the best thing to do), but I'd say we might need a 90-120s default timeout here for the parameter.

test/resource_test.rb
```diff
@@ -142,6 +142,26 @@ def test_acquire_with_fork
     end
   end
 
+  def test_quota_grace_period
```
sirupsen: Nice test. Should we also test something where they sleep for longer than the padding time?

dalehamel: added a test for that to compliment this one.

lib/semian.rb
```diff
@@ -140,7 +140,18 @@ def to_s
   # +exceptions+: An array of exception classes that should be accounted as resource errors.
   #
   # Returns the registered resource.
-  def register(name, tickets: nil, quota: nil, permissions: 0660, timeout: 0, error_threshold:, error_timeout:, success_threshold:, exceptions: [])
+  def register(name,
+    tickets: nil,
+    quota: nil,
+    permissions: 0660,
+    timeout: 0,
+    quota_grace_period: 120,
```
dalehamel: @sirupsen 120 slow enough for rubby on rails?

sirupsen: yeah lol

PR #127, opened 2017-03-16
Register all workers - not just quota workers

dalehamel:
# What

Register all workers, regardless of if they are quota-based or static-ticket based workers.

# Why

Currently it's possible to have spurious timeouts when transitioning
between static and quota-based ticket allocation.

This is primarily a problem when switching from static to quota-based,
as there will be a large number of unregistered static-ticket based
workers, who will have a very small number of tickets as the
quota-ticket based workers boot and register.

By registering all workers, we can account for this issue.

# How

During initialization, all workers will register themselves, rather than only quota-based workers.

### DIFF
--- a/test/resource_test.rb
+++ b/test/resource_test.rb
@@ -249,6 +249,32 @@ class TestResource < Minitest::Test
     assert_equal 1, resource.tickets
   end
 
+  def test_switch_static_tickets_to_quota
+    workers = 50
+    quota = 0.5
+
+    # Fork a large number of workers using static ticket strategy
+    fork_workers(count: workers - 1, tickets: 5, timeout: 0.1, wait_for_timeout: true) do
+      sleep 1
+    end
+
+    # Signal static workers to shut down
+    signal_workers('TERM')
+
+    # Create a quota based worker, and ensure it accounts for the static
+    # workers that haven't shut down yet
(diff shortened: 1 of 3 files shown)


test/resource_test.rb
```diff
@@ -249,6 +249,34 @@ def test_quota_minimum_one_ticket
...
+      sleep 1
+    end
+
+    # Signal static workers to shut down
+    signal_workers('TERM')
+
+    # Create a quota based worker, and ensure it accounts for the static
+    # workers that haven't shut down yet
+    resource = create_resource :testing, quota: quota, timeout: 0.1
+    assert_equal((quota * workers).ceil, resource.tickets)
```
dalehamel: Note that this points out an edge case that already exists - when the number of workers decreases, we will always have more tickets than the quota would specify *until* new workers boot.

This should be acceptable in practice however, as the number of tickets will be corrected later during the same deploy as more quota workers boot, or at the very latest at the start of the next deploy.

sirupsen: 👍  I think we accepted that a bit ago.

dalehamel:
:man_facepalming: accidentally merged into wrong branch, actually being merged in https://github.com/Shopify/semian/pull/128

PR #131, opened 2017-03-17
Add closed connection to mysql errors

bibstha:
During upgrade to Rails5, blocking mysql with toxiproxy returned `closed MySQL connection` which wasn't trapped here.
This PR should fix this.


sirupsen:
Good catch!

PR #134, opened 2017-03-22
Add ability to unregister all resources, and clear all semian_resources

dalehamel:
# What

Functionality to track references to resource consumers - instances of
semian adaptors - has been added. Any time an instance of a semian adaptor
retrieves a semian resource, it will register itself as a consumer of the resource.

# Why

Some situations will necessitate clearing all semian resource references without actually destroying the underlying resource.

We need to clear the reference both globally (from the list of semian resources) but also any references that semian consumers may have.

An example case would be if a semian resource were initialized by a parent process before forking. In order for each child to receive their own unique resource reference (so that worker registration is accurate), clearing all semian resource references before forking is necessary.

# How

It is assumed that the resources are registered by an instance of a Semian::Adaptor. If that is the case, it will pass along a consumer reference (itself). If a resource is retrieved in another way, it is not supported and will not have its references cleared.

To avoid this preventing GC on the semian adaptor object, this reference is intentionally weak, and the weak reference is handled safely as it is indeed volatile.

On calling `Semian.unregister_all_resources`, every resource name will be unregistered and all consumers will have their references to the semian resource cleared.

### DIFF
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -1,5 +1,6 @@
 require 'forwardable'
 require 'logger'
+require 'weakref'
 
 require 'semian/version'
 require 'semian/instrumentable'
@@ -156,6 +157,14 @@ module Semian
   end
 
   def retrieve_or_register(name, **args)
+    # If consumer who retrieved / registered by a Semian::Adapter, keep track
+    # of who the consumer was so that we can clear the resource reference if needed.
+    if consumer = args.delete(:consumer)
+      if consumer.class.include?(Semian::Adapter)
+        consumers[name] ||= []
+        consumers[name] << WeakRef.new(consumer)
+      end
+    end
     self[name] || register(name, **args)
   end
 
@@ -175,16 +184,45 @@ module Semian
   # the number of registered workers.
   # Semian.destroy removes the underlying resource, but
   # Semian.unregister will remove all references, while preserving
-  # the underlying semian resource (and sysV semaphore)
+  # the underlying semian resource (and sysV semaphore).
+  # Also clears any semian_resources
+  # in use by any semian adapters if the weak reference is still alive.
   def unregister(name)
     if resource = resources.delete(name)
-      resource.bulkhead.unregister_worker
+      resource.bulkhead.unregister_worker if resource.bulkhead
+      consumers_for_resource = consumers.delete(name) || []
+      consumers_for_resource.each do |consumer|
+        begin
+          if consumer.weakref_alive?
+            consumer.clear_semian_resource
+          end
+        rescue WeakRef::RefError
+          next
+        end
+      end
+    end
+  end
+
+  # Unregisters all resources
+  def unregister_all_resources
+    resources.keys.each do |resource|
+      unregister(resource)
     end
   end
 
   # Retrieves a hash of all registered resources.
   def resources
     @resources ||= {}
+  end
+
+  # Retrieves a hash of all registered resource consumers.
+  def consumers
+    @consumers ||= {}
+  end
+
+  def reset!
+    @consumers = {}
+    @resources = {}
   end
 
   private
--- a/test/adapter_test.rb
+++ b/test/adapter_test.rb
@@ -1,0 +1,91 @@
+require 'test_helper'
+
+class TestSemianAdapter < Minitest::Test
+  def setup
+    Semian.unregister_all_resources
+    # Consumers registered in other test files must be cleared
+    Semian.reset!
+  end
+
+  def teardown
+    Semian.unregister_all_resources
+  end
+
+  def test_adapter_registers_consumer
+    assert_empty(Semian.resources)
+    assert_empty(Semian.consumers)
+    client = Semian::AdapterTestClient.new(quota: 0.5)
+    resource = client.semian_resource
+    assert_equal(resource, Semian.resources[client.semian_identifier])
+    assert_equal(client, Semian.consumers[client.semian_identifier].first)
+  end
+
+  def test_unregister
+    client = Semian::AdapterTestClient.new(quota: 0.5)
+    assert_nil(Semian.resources[:testing])
+    resource = Semian.register(:testing, tickets: 2, error_threshold: 0, error_timeout: 0, success_threshold: 0)
+    assert_equal(Semian.resources[:testing], resource)
+
+    assert_equal 1, resource.registered_workers
+
+    without_gc do
+      Semian.unregister(:testing)
+      assert_equal 0, resource.registered_workers
+
+      assert_empty(Semian.resources)
+      assert_empty(Semian.consumers)
+
+      # The first call to client.semian_resource after unregistering all resources,
+      # should return a *different* (new) resource.
+      refute_equal(resource, client.semian_resource)
+    end
+    assert_nil(Semian.resources[:testing])
+  end
+
+  def test_unregister_all_resources
+    client = Semian::AdapterTestClient.new(quota: 0.5)
+    resource = client.semian_resource
+    assert_equal(resource, Semian.resources[client.semian_identifier])
+    assert_equal(client, Semian.consumers[client.semian_identifier].first)
+
+    # need to disable GC to ensure client weak reference is alive for assertion below
+    without_gc do
+      assert_equal(resource, client.semian_resource)
+      Semian.unregister_all_resources
+
+      assert_empty(Semian.resources)
+      assert_empty(Semian.consumers)
+
+      # The first call to client.semian_resource after unregistering all resources,
+      # should return a *different* (new) resource.
+      refute_equal(resource, client.semian_resource)
+    end
+  end
+
+  def test_consumer_registration_does_not_prevent_gc
+    client = Semian::AdapterTestClient.new(quota: 0.5)
+    client.semian_resource
+    identifier = client.semian_identifier
+
+    # Release the only strong reference to the client object
+    # so that it will be cleared on the forced GC run below
+    client = nil
+    weak_ref = Semian.consumers[identifier].first
+    assert_equal(true, weak_ref.weakref_alive?)
+
+    GC.start
+
+    assert_nil weak_ref.weakref_alive?
+
+    assert_raises WeakRef::RefError do
+      weak_ref.any_method_call
+    end
+  end
+
+  def without_gc
+    GC.disable
+    yield
+  ensure
+    GC.enable
+  end
+end
--- a/test/helpers/adapter_helper.rb
+++ b/test/helpers/adapter_helper.rb
@@ -1,0 +1,27 @@
+module Semian
+  module AdapterTest
+    include Semian::Adapter
+
+    def semian_identifier
+      :semian_adapter_test
+    end
+
+    def raw_semian_options
+      @client_options
+    end
+  end
+
+  class AdapterTestClient
+    include AdapterTest
+
+    def initialize(**args)
+      @client_options = args.merge(success_threshold: 1,
+                                   error_threshold: 1,
+                                   error_timeout: 1)
+    end
+
+    def ==(other)
+      inspect == other.inspect
+    end
+  end
+end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -111,3 +111,7 @@ Style/IfUnlessModifier:
 
 Style/CaseIndentation:
   IndentWhenRelativeTo: end
+
+Lint/UselessAssignment:
+  Exclude:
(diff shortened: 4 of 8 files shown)


test/adapter_test.rb
```diff
@@ -0,0 +1,58 @@
...
+    client.semian_resource
+    identifier = client.semian_identifier
+
+    client = nil
+    weak_ref = Semian.consumers[identifier].first
+    assert_equal(true, weak_ref.weakref_alive?)
+
+    GC.start
+
+    assert_nil client
```
dalehamel: just so rubocop didn't yell at me for setting it to nil and not using it :| bad form i guess.

lib/semian.rb
```diff
@@ -156,6 +157,14 @@ def register(name, **options)
   end
 
   def retrieve_or_register(name, **args)
+    # If consumer who retrieved / registered by a Semian::Adapter, keep track
+    # of who the consumer was so that we can clear the resource reference if needed.
+    if consumer = args.delete(:consumer)
+      if consumer.class.included_modules.include?(Semian::Adapter)
```
fw42: what's the difference between `included_modules.include?` and `include?`

test/adapter_test.rb
```diff
@@ -0,0 +1,58 @@
+require 'test_helper'
+
+class TestSemianAdapter < Minitest::Test
+  def setup
+    Semian.unregister_all_resources
+    # Consumers registered in other test files must be cleared
+    Semian.instance_variable_set(:@consumers, nil)
```
fw42: can you do `Semian.consumers.clear`? Seems a little less hacky

test/adapter_test.rb
```diff
@@ -0,0 +1,58 @@
...
+    Semian.unregister_all_resources
+  end
+
+  def test_adapter_registers_consumer
+    assert_equal(true, Semian.resources.empty?)
+    assert_equal(true, Semian.consumers.empty?)
+    client = Semian::AdapterTestClient.new(quota: 0.5)
+    resource = client.semian_resource
+    assert_equal(resource, Semian.resources[client.semian_identifier])
+    assert_equal(client.inspect, Semian.consumers[client.semian_identifier].first.inspect)
```
fw42: why `inspect`?

test/adapter_test.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    GC.disable # need to disable GC to ensure client weak reference is alive for assertion below
+    Semian.unregister_all_resources
+
+    assert_equal(true, Semian.resources.empty?)
+    assert_equal(true, Semian.consumers.empty?)
+    assert_nil client.instance_variable_get(:@semian_resource)
+    GC.enable
```
fw42: not a huge deal but the `GC.disable` will leak if the assertions fail and the `GC.enable` never runs. Maybe consider adding a helper method for this:
```ruby
def without_gc
  GC.disable
  yield
ensure
  GC.enable
end
```

lib/semian.rb
byroot: None, except it's slower because it has to allocate an Array.

The proper way to do this is `consumer.is_a?(Semian::Adapter)`

dalehamel: Thanks, wasn't aware of that :+1:

dalehamel: Actually I don't think that's correct @byroot 

```
irb(main):001:0> module Bar; end
=> nil
irb(main):002:0> class Foo
irb(main):003:1> include Bar
irb(main):004:1> end
=> Foo
irb(main):005:0> Foo.is_a?(Bar)
false
irb(main):006:0> Foo.include?(Bar)
=> true
```

So I think include? is the way to go (didn't know this shorthand)

test/adapter_test.rb
dalehamel: this was me being lazy and not writing a comparator. minitest was complaining that the inspect was identical but that it didn't have a comparator for Semian::AdapterTestClient. I can write one.

lib/semian.rb
byroot: @dalehamel in your example you invoke it on the class, whereas in the code we're commenting you call `.class` on the instance.

`instance.class.include?(Module)` vs `instance.is_a?(Module)`

test/adapter_test.rb
dalehamel: agreed

test/adapter_test.rb
```diff
@@ -0,0 +1,58 @@
...
+    resource = client.semian_resource
+    assert_equal(resource, Semian.resources[client.semian_identifier])
+    assert_equal(client.inspect, Semian.consumers[client.semian_identifier].first.inspect)
+
+    GC.disable # need to disable GC to ensure client weak reference is alive for assertion below
+    Semian.unregister_all_resources
+
+    assert_equal(true, Semian.resources.empty?)
+    assert_equal(true, Semian.consumers.empty?)
+    assert_nil client.instance_variable_get(:@semian_resource)
```
fw42: test the interface, not the implementation :-(

test/adapter_test.rb
dalehamel: Added one, but it just compares via inspect.

test/adapter_test.rb
dalehamel: I'm going to just make Semian.reset!

test/adapter_test.rb
```diff
@@ -0,0 +1,58 @@
...
+    client = Semian::AdapterTestClient.new(quota: 0.5)
+    resource = client.semian_resource
+    assert_equal(resource, Semian.resources[client.semian_identifier])
+    assert_equal(client.inspect, Semian.consumers[client.semian_identifier].first.inspect)
+
+    GC.disable # need to disable GC to ensure client weak reference is alive for assertion below
+    Semian.unregister_all_resources
+
+    assert_equal(true, Semian.resources.empty?)
+    assert_equal(true, Semian.consumers.empty?)
```
fw42: nitpick I think there is `assert_empty` which you could use here

test/adapter_test.rb
dalehamel: @fw42 I agree, but I'm not sure the best way to work around this.

Calling semian_resource will create a new one, which affects what I'm trying to test here.

Any ideas?

dalehamel:
@fw42 any ideas for how to test that the semian_resource is nil in a less leaky abstraction way?

fw42:
> Calling semian_resource will create a new one, which affects what I'm trying to test here.

Maybe asserting that you are getting a new one (rather than the same one as before) is equivalent to asserting that the old one isn't there anymore?

dalehamel:
> Maybe asserting that you are getting a new one (rather than the same one as before) is equivalent to asserting that the old one isn't there anymore?

That's a good point, I appreciate the fresh perspective. Yeah, I think that will suffice.

dalehamel:
@fw42 think I addressed everything, another :eyes: would be appreciated before merging.

I'll hold off on the merge until I'm sure this actually fixes the problem I'm trying to fix though.

lib/semian.rb
```diff
@@ -178,7 +187,25 @@ def destroy(name)
   # the underlying semian resource (and sysV semaphore)
   def unregister(name)
     if resource = resources.delete(name)
-      resource.bulkhead.unregister_worker
+      resource.bulkhead.unregister_worker if resource.bulkhead
```
fw42: why is this new `if` needed now?

dalehamel: @fw42 I noticed a failed test where it is possible for the bulkhead to be nil, I hadn't considered this before so I added this guard to handle that case.

test/adapter_test.rb
```diff
@@ -0,0 +1,68 @@
...
+      refute_equal(resource, client.semian_resource)
+    end
+  end
+
+  def test_consumer_registration_does_not_prevent_gc
+    client = Semian::AdapterTestClient.new(quota: 0.5)
+    client.semian_resource
+    identifier = client.semian_identifier
+
+    client = nil
```
fw42: is this here on purpose? if yes then why do you need the `assert_nil client` below?

test/helpers/adapter_helper.rb
```diff
@@ -0,0 +1,27 @@
...
+    include AdapterTest
+
+    def initialize(**args)
+      @client_options = args.merge(success_threshold: 1,
+                                   error_threshold: 1,
+                                   error_timeout: 1)
+    end
+
+    def ==(other)
+      inspect == other.inspect
```
fw42: This still seems a little bit hacky and I don't really understand why we need this. Why inspect? Maybe `object_id == other.object_id` makes more sense?

fw42: or `client_options == other.client_options`?

test/adapter_test.rb
dalehamel: @fw42 this is to force the garbage collection (by dropping the only strong reference). 

The assert_nil is a hack to prevent rubocop from complaining about the "useless variable assignment".

Alternatively, I suppose I could just whitelist this with rubocop.

test/helpers/adapter_helper.rb
dalehamel: @fw42 that's because when we do comparisons, it's between a strong and a weak reference to the same object.

comparing them via inspect to assert that it's the same object, even though the ID doesn't match.

test/adapter_test.rb
fw42: Ah, I see. Maybe move it into a method to force the GC? Like
```ruby
def identifier_of_new_client
  client = Semian::AdapterTestClient.new(quota: 0.5)
  client.semian_resource
  client.semian_identifier
end

identifier = identifier_of_new_client
```
or maybe
```ruby
identifier = Semian::AdapterTestClient
  .new(quota: 0.5)
  .tap(&:semian_resource)
  .semian_identifier
```
If you like yours better, then I think a comment might make sense, like
```ruby
client = nil # force client to get garbage collected
assert_nil client # make rubocop shut up about unused variable
```

dalehamel:
I realized that it would be possible to unregister more resources that are actually registered in some cases - if you try to unregister a resource registered in the parent from the child after forking, for instance.

If that happened, the default behavior would be to block until more registered workers showed up, which is definitely not what we want.

From `man semop`:

```
 If sem_op is less than zero, the process must have alter permission on the sem‐
       aphore set.  If semval is greater than  or  equal  to  the  absolute  value  of
       sem_op,  the operation can proceed immediately: the absolute value of sem_op is
       subtracted from semval, and, if SEM_UNDO is specified for this  operation,  the
       system  adds  the absolute value of sem_op to the semaphore adjustment (semadj)
       value for this semaphore.  If the absolute value of sem_op is greater than sem‐
       val,  and  IPC_NOWAIT is specified in sem_flg, semop() fails, with errno set to
       EAGAIN (and none of the operations in sops is performed).
```

So we can specify the IPC_NOWAIT flag, which will throw EAGAIN if it's not able to proceed (the count is 0). We can simply ignore that, as it's the behavior that we want - if we have hit 0 workers, no more workers can possibly be unregistered.

fw42:
good catch

dalehamel:
> good catch

:cold_sweat: thank - it was your comment that made me realize this edge case

lib/semian.rb
```diff
@@ -175,10 +184,29 @@ def destroy(name)
...
-  # the underlying semian resource (and sysV semaphore)
+  # the underlying semian resource (and sysV semaphore).
+  # Also clears any semian_resources
+  # in use by any semian adapters if the weak reference is still alive.
   def unregister(name)
     if resource = resources.delete(name)
-      resource.bulkhead.unregister_worker
+      resource.bulkhead.unregister_worker if resource.bulkhead
+      consumers_for_resource = consumers.delete(name) || []
+      consumers_for_resource.each do |consumer|
```
sirupsen: In what scenarios can we have multiple adapter consumers for a resource?

dalehamel: @sirupsen I believe that if the consumer goes away (client is cleared) the old semian resource can stay around while a new client is created.

sirupsen: What would clear the client?

dalehamel: I don't know, it's impossible to know what the behavior of the user of the adapter will be.

Are you arguing that this should be a hash?

sirupsen: Not arguing for anything necessarily, just trying to understand why it was written to support more than one consumer :)

dalehamel: I can't recall exactly, but I don't know for certain that there can be only one consumer for a given resource.

PR #136, opened 2017-03-23
Expose sysv IPC key to semian resource in ruby.

dalehamel:
# What

Exposes the string representation of the hexadecimal sysvIPC key.

The Sysv IPC resource key is a portable hexadecimal string that uniquely identifies a particular resource.

# Why

The sysv IPC key is extremely useful for manually diagnosing and debugging via `ipcs` as it is portable across systems. I had previously been manually adding an `sprintf` on a regular basis in order to debug, so may as well make this a first class citizen

# How

Some minor code reorganization, adding new members to the resource struct and passing it in and referencing it to the sysv_semaphore initializer.

### DIFF
--- a/ext/semian/resource.c
+++ b/ext/semian/resource.c
@@ -158,6 +158,14 @@ semian_resource_id(VALUE self)
 }
 
 VALUE
+semian_resource_key(VALUE self)
+{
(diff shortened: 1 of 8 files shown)


sirupsen:
😍

PR #137, opened 2017-03-23
Expose registered worker count to resource in rubyland

dalehamel:
# What

Expose the registered worker count to the ruby

# Why

The number of registered workers is extremely important to keep track of for the quota strategy. Exposing it to the resource allows for it to be emitted as a statsd metric very easily.

# How

Essentially, this is just a standard accessor pattern.

### DIFF
--- a/ext/semian/resource.c
+++ b/ext/semian/resource.c
@@ -135,6 +135,21 @@ semian_resource_tickets(VALUE self)
 }
 
 VALUE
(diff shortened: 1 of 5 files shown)


dalehamel:
Sorry for all the reviewers requested, the usual suspects for reviews are away.

Adding @sirupsen just so he can see it when he gets back.

PR #139, opened 2017-03-29
Specify a minimum number of tickets of 2 using quota strategy

dalehamel:
# What

Provide a value for the minimum number of tickets using the ticket quota strategy.

We default this to 2, to prevent issues where there is a small number of registered workers. It may be overridden.

This may be overridden by the user, but 2 is likely to be a reasonable default to prevent an overly-sensitive bulkhead.

# Why

If we have a very small number of registered workers, the bulkhead may be overly sensitive. For instance, if we have 2 workers with a quota of 0.5:

* 2 * 0.5 = 1 ticket
* Worker 1 acquires the ticket, but needs X time before it can release it
* During this, worker 2 tries to acquire the ticket, but has a timeout of Y

If X > Y in this situation, worker 2 will time out. 

This could be the case for resources that are accessed very infrequently or have a very small number of workers that might access them.

# How

If the number of workers calculated by the quota is less than the minimum, use the minimum instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,8 +185,8 @@ In this case, we'd allow 50% of the workers on a particular host to connect to t
 
 - You must pass **exactly** one of ticket or quota.
 - Tickets available will be the ceiling of the quota ratio to the number of workers
- - So, with one worker, there will always be a minimum of 1 ticket
 - Workers in different processes will automatically unregister when the process exits.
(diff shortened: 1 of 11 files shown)


dalehamel:
A better solution than this is actually to just append a small decimal to the quota.

EG, a quota of 0.5 with 2 workers yields one ticket, but 0.51 yields 2 which is sufficient to stop this.

PR #142, opened 2017-03-31
Add helper to forcefully reset worker counts

dalehamel:
# What

Add a helper to forcefully reset registered worker counts to zero.

# Why

In case if the worker count gets corrupted, this is safe to do for resources using static tickets.

# How

This method would need to be manually invoked, and then the next time the workers restart they should correctly register themselves. semctl with SETVAL will clear the SEM_UNDO table, which should cause the next registration to proceed correctly.

This is only safe on resources that are using static tickets or for debugging / diagnostic purposes, and should not be called regularly.

### DIFF
--- a/ext/semian/resource.c
+++ b/ext/semian/resource.c
@@ -93,17 +93,19 @@ semian_resource_destroy(VALUE self)
 VALUE
 semian_resource_reset_workers(VALUE self)
 {
-
+  int ret;
   semian_resource_t *res = NULL;
 
   TypedData_Get_Struct(self, semian_resource_t, &semian_resource_type, res);
 
   sem_meta_lock(res->sem_id);
   // This SETVAL will purge the SEM_UNDO table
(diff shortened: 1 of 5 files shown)


ext/semian/resource.c
```diff
@@ -91,6 +91,24 @@ semian_resource_destroy(VALUE self)
...
+{
+
+  semian_resource_t *res = NULL;
+
+  TypedData_Get_Struct(self, semian_resource_t, &semian_resource_type, res);
+
+  sem_meta_lock(res->sem_id);
+  // This SETVAL will purge the SEM_UNDO table
+  if (semctl(res->sem_id, SI_SEM_REGISTERED_WORKERS, SETVAL, 0) == -1) {
+    raise_semian_syscall_error("semctl()", errno);
```
camilo: Is it possible that you are not releasing the lock here?

PR #143, opened 2017-04-03
Must be able to access name without a bulkhead

dalehamel:
# What

Prevent a nil pointer dereference if attempting to read the name of a protected resource without a bulkhead.

# Why

Since bulkheads are now optional, we shouldn't delegate name explicitly to them, as this can result in a nil pointer dereference.

# How

The name is attempted to be read from the bulkhead first, falling back to the circuit breaker. If neither works, it will be nil, but it won't cause an exception to be raised from a nil pointer dereference.

### DIFF
--- a/lib/semian/protected_resource.rb
+++ b/lib/semian/protected_resource.rb
@@ -2,15 +2,16 @@ module Semian
   class ProtectedResource
     extend Forwardable
 
-    def_delegators :@bulkhead, :destroy, :count, :semid, :tickets, :registered_workers, :name
+    def_delegators :@bulkhead, :destroy, :count, :semid, :tickets, :registered_workers
     def_delegators :@circuit_breaker, :reset, :mark_failed, :mark_success, :request_allowed?,
                    :open?, :closed?, :half_open?
 
-    attr_reader :bulkhead, :circuit_breaker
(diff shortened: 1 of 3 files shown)


dalehamel:
We might just want "allow_nil" here, not sure what the best behavior is.

lib/semian/protected_resource.rb
```diff
@@ -2,15 +2,15 @@ module Semian
   class ProtectedResource
     extend Forwardable
 
-    def_delegators :@bulkhead, :destroy, :count, :semid, :tickets, :registered_workers, :name
     def_delegators :@circuit_breaker, :reset, :mark_failed, :mark_success, :request_allowed?,
```
sirupsen: Shouldn't we do the same if the circuit breaker is disabled?

dalehamel: Yeah i think both should just do allow_nil.

dalehamel:
> Yeah, I think we should consider just doing allow_nil. This is definitely an error, I don't think we should mock every call.

@sirupsen looks like `allow_nil` is a rails thing, if we are just using `forwardable` we might need to manually delegate things to make them safe when the delegate is nil.

PR #144, opened 2017-04-13
Add new mysql connection error

tamcgoey:
During our gem bump of `mysql2` (to 0.4.5), we noticed that we weren't throwing any `Mysql2::CircuitOpenError` exceptions, since a new mysql exception message was being used: `MySQL client is not connected`.

cc @georgetzavelas @bibstha

### DIFF
--- a/lib/semian/mysql2.rb
+++ b/lib/semian/mysql2.rb
@@ -25,6 +25,7 @@ module Semian
       /MySQL server has gone away/i,
       /Too many connections/i,
       /closed MySQL connection/i,
+      /MySQL client is not connected/i,
     )
 
     ResourceBusyError = ::Mysql2::ResourceBusyError


sirupsen:
Is this happening in jobs or in web requests? Typically this error happens on code-paths such as:

```ruby
# web request 1 with mysql down

begin
  Shop.first # fails
rescue
  # request continues, but client has been set to nil since it failed
end

Shop.first # no client connected error

# web request 2

# client has been reconnected when checked out in the connection pool middleware
Shop.first # succeeds
```

I think this error typically indicates rescuing `ActiveRecord` and `MySQL` connectivity exceptions where you should just fail and allow the middleware to re-connect, or re-connect inline.

georgetzavelas:
Yeah it seems the error message changes 3 times before settling on `MySQL client is not connected` when hitting our endpoint:
```
(byebug) get '/services/migrations'
*** ActiveRecord::StatementInvalid Exception: Mysql2::Error: [mysql_master] Lost connection to MySQL server during query: SELECT * FROM schema_migrations

nil
(byebug) get '/services/migrations'
*** ActiveRecord::StatementInvalid Exception: Mysql2::Error: [mysql_master] MySQL server has gone away: SELECT * FROM schema_migrations

nil
(byebug) get '/services/migrations'
*** ActiveRecord::StatementInvalid Exception: Mysql2::Error: MySQL client is not connected: SELECT * FROM schema_migrations

nil
(byebug) get '/services/migrations'
*** ActiveRecord::StatementInvalid Exception: Mysql2::Error: MySQL client is not connected: SELECT * FROM schema_migrations

nil
(byebug) get '/services/migrations'
*** ActiveRecord::StatementInvalid Exception: Mysql2::Error: MySQL client is not connected: SELECT * FROM schema_migrations
```

sirupsen:
Are those failing fast? Has `0.4.5` introduced its own circuit breaker perhaps?

tamcgoey:
It looks like this is new behaviour (a bug fix) for version 0.4.5.

[It was introduced in this commit](https://github.com/brianmario/mysql2/commit/f14023fcfee9e85e6fc1b0e568048811518f8c23), in order to [Move macro REQUIRE_CONNECTED to client.c](https://github.com/brianmario/mysql2/issues/763).

I didn't seen any mention of rolling their own circuit breaker.

tamcgoey:
Wanted to check back and see if we could support this, since it's blocking our upgrade to the latest mysql2 version ([which introduced the new error message](https://github.com/Shopify/semian/pull/144#issuecomment-293921192)).

cc @sirupsen

sirupsen:
Sorry, for the delay here. I was away yesterday. 😬 

Are you seeing these problems in production? In the test above @georgetzavelas are you taking down MySQL? I would agree that the circuit breaker error is the appropriate one here.

I'm concerned whether you can get in a state where you're not reconnecting, unless that console is in a functional testing environment where reconnection would _not_ occur (since it happens in a middleware in Rails).

If you can ack that, I'll merge—otherwise, let's get to the root of a nasty issue in your app. The reason I didn't merge this right away is that concern, I want to make sure you're not covering up something super scary 😱

nicolaslupien:
Btw, in '/services/migrations' we don't use AR connection directly without any check/rescue 
```ruby
render json: ApplicationRecord.connection.select_values("SELECT * FROM schema_migrations").to_json
```
https://github.com/Shopify/shipify/blob/master/app/controllers/services/ping_controller.rb#L15

sirupsen:
When will it reconnect after that failure if the database recovers?

sirupsen:
What I would expect to happen is that it'd call `verify!`, which will do a ping to MySQL. If that ping fails, it'll reconnect. In the console session above, is MySQL continuously down, or has it recovered?

georgetzavelas:
@sirupsen sorry for any confusion about my console test, that was me executing the API call in the middle of a debugger session just to illustrate the error messages.

PR #147, opened 2017-06-07
adapters: expose resource exceptions

sirupsen:
@hkdsun


sirupsen:
We don't need this. The adapter error is included into the base error.

PR #149, opened 2017-06-09
Open circuit on server errors (conditionally)

andrewlouis93:
Context: Exceptions are not raised for 5xxs. If the endpoint was always responding, but returning 5xxs, Semian wouldn't open the circuit. 

This PR allows exceptions to be raised on such requests conditionally, allowing Semian to open/close circuit this case. 

cc/ @albiere

### DIFF
--- a/lib/semian/circuit_breaker.rb
+++ b/lib/semian/circuit_breaker.rb
@@ -4,12 +4,13 @@ module Semian
 
     def_delegators :@state, :closed?, :open?, :half_open?
 
-    def initialize(name, exceptions:, success_threshold:, error_threshold:, error_timeout:, implementation:)
+    def initialize(name, open_circuit_server_errors: false, exceptions:, success_threshold:, error_threshold:, error_timeout:, implementation:)
       @name = name.to_sym
       @success_count_threshold = success_threshold
       @error_count_threshold = error_threshold
       @error_timeout = error_timeout
       @exceptions = exceptions
+      @open_circuit_server_errors = open_circuit_server_errors
 
       @errors = implementation::SlidingWindow.new(max_size: @error_count_threshold)
       @successes = implementation::Integer.new
@@ -24,6 +25,7 @@ module Semian
       result = nil
       begin
         result = yield
+        mark_failed(result) if result.is_a?(::Net::HTTPServerError) && @open_circuit_server_errors
       rescue *@exceptions => error
         mark_failed(error)
         raise error
--- a/lib/semian/circuit_breaker.rb
+++ b/lib/semian/circuit_breaker.rb
@@ -4,13 +4,12 @@ module Semian
 
     def_delegators :@state, :closed?, :open?, :half_open?
 
-    def initialize(name, open_circuit_server_errors: false, exceptions:, success_threshold:, error_threshold:, error_timeout:, implementation:)
+    def initialize(name, exceptions:, success_threshold:, error_threshold:, error_timeout:, implementation:)
       @name = name.to_sym
       @success_count_threshold = success_threshold
       @error_count_threshold = error_threshold
       @error_timeout = error_timeout
       @exceptions = exceptions
-      @open_circuit_server_errors = open_circuit_server_errors
 
       @errors = implementation::SlidingWindow.new(max_size: @error_count_threshold)
       @successes = implementation::Integer.new
@@ -25,7 +24,6 @@ module Semian
       result = nil
       begin
         result = yield
-        mark_failed(result) if result.is_a?(::Net::HTTPServerError) && @open_circuit_server_errors
       rescue *@exceptions => error
         mark_failed(error)
         raise error
--- a/lib/semian/net_http.rb
+++ b/lib/semian/net_http.rb
@@ -36,7 +36,6 @@ module Semian
       ::Timeout::Error, # includes ::Net::ReadTimeout and ::Net::OpenTimeout
       ::SocketError,
       ::Net::HTTPBadResponse,
-      ::Net::HTTPFatalError,
       ::Net::HTTPHeaderSyntaxError,
       ::Net::ProtocolError,
       ::EOFError,
@@ -88,10 +87,6 @@ module Semian
       raw_semian_options.nil?
     end
 
-    def fatal_server_errors?
-      raw_semian_options.fetch(:fatal_server_errors)
-    end
-
     def connect
       return raw_connect if disabled?
       acquire_semian_resource(adapter: :http, scope: :connection) { raw_connect }
@@ -99,13 +94,7 @@ module Semian
 
     def request(req, body = nil, &block)
       return raw_request(req, body, &block) if disabled?
-      acquire_semian_resource(adapter: :http, scope: :query) do
-        result = raw_request(req, body, &block)
-
-        result.error! if result.is_a?(::Net::HTTPServerError) && fatal_server_errors?
-
-        result
-      end
+      acquire_semian_resource(adapter: :http, scope: :query) { raw_request(req, body, &block) }
     end
   end
 end
--- a/lib/semian/net_http.rb
+++ b/lib/semian/net_http.rb
@@ -104,7 +104,7 @@ module Semian
     private
 
     def handle_error_responses(res)
-      return unless raw_semian_options.fetch(:open_circuit_server_errors)
+      return unless raw_semian_options.fetch(:open_circuit_server_errors, false)
       semian_resource.mark_failed(res) if res.is_a?(::Net::HTTPServerError)
     end
   end
--- a/lib/semian/net_http.rb
+++ b/lib/semian/net_http.rb
@@ -36,6 +36,7 @@ module Semian
       ::Timeout::Error, # includes ::Net::ReadTimeout and ::Net::OpenTimeout
       ::SocketError,
       ::Net::HTTPBadResponse,
+      ::Net::HTTPFatalError,
       ::Net::HTTPHeaderSyntaxError,
       ::Net::ProtocolError,
       ::EOFError,
@@ -87,6 +88,10 @@ module Semian
       raw_semian_options.nil?
     end
 
+    def fatal_server_errors?
+      raw_semian_options.fetch(:fatal_server_errors)
+    end
+
     def connect
       return raw_connect if disabled?
       acquire_semian_resource(adapter: :http, scope: :connection) { raw_connect }
@@ -94,7 +99,13 @@ module Semian
 
     def request(req, body = nil, &block)
       return raw_request(req, body, &block) if disabled?
-      acquire_semian_resource(adapter: :http, scope: :query) { raw_request(req, body, &block) }
+      acquire_semian_resource(adapter: :http, scope: :query) do
+        result = raw_request(req, body, &block)
+
+        result.error! if result.is_a?(::Net::HTTPServerError) && fatal_server_errors?
+
+        result
+      end
     end
   end
 end
--- a/lib/semian/net_http.rb
+++ b/lib/semian/net_http.rb
@@ -45,8 +45,8 @@ module Semian
 
     # The naked methods are exposed as `raw_query` and `raw_connect` for instrumentation purpose
     def self.included(base)
-      base.send(:alias_method, :raw_request, :request)
-      base.send(:remove_method, :request)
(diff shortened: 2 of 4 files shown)


jpittis:
My understanding is that Semian has two jobs.
1. To protect the client from slow requests eating up capacity.
2. To protect a service that is at capacity.

Treating 500s as exceptions is only helping out (2). Assuming the returns of these 500s are fast, the client shouldn't really care. (If not, then they'd cause a timeout exception instead?)

I'm trying to think of why we would want opt-in 500s opening the circuit rather than on by default...

For example, what if a service is returning 500s to 50% of requests. (Maybe because there's a 50% chance of being load balanced to a bad node.) Semian could open the circuit which would make 100% of requests fail...

I like the sound of this but we should make sure we want this on by default.

lib/semian/net_http.rb
```diff
@@ -87,14 +88,24 @@ def disabled?
       raw_semian_options.nil?
     end
 
+    def fatal_server_errors?
```
sirupsen: This should not be a public method. Since you only use it in one place, I'd inline it.

lib/semian/net_http.rb
```diff
@@ -36,6 +36,7 @@ def semian_identifier
       ::Timeout::Error, # includes ::Net::ReadTimeout and ::Net::OpenTimeout
       ::SocketError,
       ::Net::HTTPBadResponse,
+      ::Net::HTTPFatalError,
```
sirupsen: If you include it here by default, why does your option matter? I think your option should ensure that `::Net::HTTPFatalError` is in resource exceptions. My impression of reading this code would be that it _will_ open the circuit, but it won't raise..?

test/net_http_test.rb
```diff
@@ -312,6 +313,46 @@ def test_adding_custom_errors_do_trip_circuit
...
+      end
+    end
+  end
+
+  def test_5xxs_dont_raise_exceptions_unless_fatal_server_flag_enabled
+    with_semian_configuration do
+      with_server do
+        http = Net::HTTP.new(HOSTNAME, TOXIC_PORT)
+        http.raw_semian_options[:error_threshold].times do
+          http.get("/500")
```
sirupsen: Why don't you need a `::Net::HTTPFatalError` here?

lib/semian/net_http.rb
```diff
@@ -87,14 +88,24 @@ def disabled?
       raw_semian_options.nil?
     end
 
+    def fatal_server_errors?
+      raw_semian_options.fetch(:fatal_server_errors)
```
sirupsen: Where can you document this option? I'd suggest adding a section for adapter-specific configuration in the `Adapters` section.

lib/semian/net_http.rb
```diff
@@ -87,14 +88,24 @@ def disabled?
...
       acquire_semian_resource(adapter: :http, scope: :connection) { raw_connect }
     end
 
     def request(req, body = nil, &block)
       return raw_request(req, body, &block) if disabled?
-      acquire_semian_resource(adapter: :http, scope: :query) { raw_request(req, body, &block) }
+      acquire_semian_resource(adapter: :http, scope: :query) do
+        result = raw_request(req, body, &block)
+
+        result.error! if result.is_a?(::Net::HTTPServerError) && fatal_server_errors?
```
sirupsen: Why do you want to raise an exception and change the behavior of `Net::HTTP` from Semian entirely? That seems counter-intuitive.

lib/semian/net_http.rb
andrewlouis93: Your impression in correct, it's the code within the Net HTTP request method that raises the exception.

test/net_http_test.rb
albiere: Because the configuration `fatal_server_errors` needs to be set to `true` and this doesn't happen by default.

lib/semian/net_http.rb
albiere: We thought that raising an exception will be the right way to do for Semian to open the circuit. Do you think that we could simply open the circuit and return the 500 response normally?

test/net_http_test.rb
andrewlouis93: ^ +1 - we don't want to change the existing behaviour, unless the option is set to true.

andrewlouis93: In the default semian config there - the option is set to false. https://github.com/Shopify/semian/pull/149/files/0e51ba07fa90c9f8b4ccff0e396226f43127ba36#diff-416f865ec7d8b97ae535d742b47da46eR22

That test makes sure the existing behaviour is preserved in this case.

lib/semian/net_http.rb
albiere: > I think your option should ensure that ::Net::HTTPFatalError is in resource exceptions

Just to clarify, you think we should only have the `::Net::HTTPFatalError` in the list of exceptions if the option `fatal_server_errors` is enabled?

andrewlouis93: And 👍 to including it conditionally.

lib/semian/net_http.rb
sirupsen: I think you should return the 500 response normally, otherwise you break all the clients and leak this abstraction.

andrewlouis93: @sirupsen are you suggesting that we raise the exception somewhere else? Maybe here: https://github.com/Shopify/semian/blob/4e61453710bf0dc8dc5de35f83bfee31f3d92742/lib/semian/adapter.rb#L32? 

The rationale of putting it there is that we wanted to limit this behaviour to Net::HTTP requests. 

Are there any other alternatives I'm missing?

sirupsen: My point is that if you start raising, you've changed the API for `Net::HTTP`. That is something we've been tempted to do many times in the course of developing Semian (changing client drivers behavior based on failure), but it leaks the abstraction to clients.

I don't think you should raise, I think you should just return the `5xx` quicker. Whether you get a `5xx` directly from the server, or you're hitting a circuit and getting it quicker, the client should deal with it the same way.

andrewlouis93: @sirupsen What do you think about the approach in the last commit? We would be now able to trip the circuit on 5xxs by simply marking them as failures given the user sets that option to `true`, and that logic is now decoupled from the adapters.

albiere: @andrewlouis93 You say it's decoupled from the adapters, but that's not really a good thing. You're coupling code that should only belong to the net/http adapter into the main implementation of the circuit breaker.

lib/semian/circuit_breaker.rb
```diff
@@ -24,6 +25,7 @@ def acquire
       result = nil
       begin
         result = yield
+        mark_failed(result) if result.is_a?(::Net::HTTPServerError) && @open_circuit_server_errors
```
sirupsen: This can't leak into the circuit breaker. You must contain this in the `::Net::HTTP` adapter. You can look at MySQL which needs to do something similar to this, I think.

test/net_http_test.rb
```diff
@@ -312,6 +313,44 @@ def test_adding_custom_errors_do_trip_circuit
...
+      tickets: 3,
+      success_threshold: 1,
+      error_threshold: 3,
+      error_timeout: 10,
+      open_circuit_server_errors: true,
+    }
+    modified_semian_config = proc do |host, port|
+      next nil if host == "127.0.0.1" && port == 8474 # disable if toxiproxy
+      semian_config.merge(name: "#{host}_#{port}")
+    end
```
sirupsen: Could you make this part of the `with_semian_configuration` helper as a default, so we don't duplicate the Toxiproxy stuff? Feels really dirty 😬

lib/semian/net_http.rb
```diff
@@ -87,6 +87,16 @@ def disabled?
       raw_semian_options.nil?
     end
 
+    def acquire_semian_resource(*)
+      result = super
+
+      if result.is_a?(::Net::HTTPServerError) && raw_semian_options.fetch(:open_circuit_server_errors)
```
sirupsen: Wait... `Net::HTTP` returns a constant as the result on a request on `5xx`? That seems highly strange :s

andrewlouis93: That was just a stopping point after a while of looking at this, that's not what I think this should look like. 

I started out with something simple like this: 

```
    def acquire_semian_resource(*)
      result = super

      if result.is_a?(::Net::HTTPServerError) && raw_semian_options.fetch(:open_circuit_server_errors)
        semian_resource.mark_failed(result)
      end

      result
    end
```

But this started failing the test I wrote (the one that tests that 500s open the circuit). 

It appeared that each [`http.get("/500")`](https://github.com/Shopify/semian/pull/149/files#diff-416f865ec7d8b97ae535d742b47da46eR333) would reach `semian_resource.mark_failed(result)` in the snippet above _twice_. Causing the circuit to be opened prematurely _before_ the `error_threshold` is reached. 

This happens because `raw_request` calls our `request` (and `connect` as well!) method again if there isn't a connection already. See: https://github.com/ruby/ruby/blob/trunk/lib/net/http.rb#L1425

Not sure what the path forwards looks like, we're still thinking about it. One option is: 

```
    def request(req, body = nil, &block)
      return raw_request(req, body, &block) if disabled?
      start if !started?
      acquire_semian_resource(adapter: :http, scope: :query) { raw_request(req, body, &block) }
    end
```
Which would avoid that second `request` invocation from `raw_request`. 

We're still thinking about it, let me know if you have any thoughts in the meantime.

sirupsen:
How's it going? Looks like you found another place to monkey-patch?

albiere:
@sirupsen Yeah! It's a trade-off. If we override the `request` method, we need to check whether the connection has started. For example:

```ruby
def request(req, body = nil, &block)
  return raw_request(req, body, &block) if disabled? || !started?
  acquire_semian_resource(adapter: :http, scope: :query) do 
    response = raw_request(req, body, &block) 
    # logic to mark resource as failed.
    response
  end
end
```

This is necessary because the implementation of net/http will make a recursive call if the connection has not started. By overriding the `transport_request`, the monkey-patched method will be called just once for 500 responses (given that 5xxs do not have the headers that the `sspi_auth?(res)` checks).

Wdyt?

For reference: https://github.com/ruby/ruby/blob/trunk/lib/net/http.rb#L1425

test/net_http_test.rb
```diff
@@ -312,6 +312,43 @@ def test_adding_custom_errors_do_trip_circuit
...
+      }
+    end
+
+    with_semian_configuration(options) do
+      with_server do
+        http = Net::HTTP.new(HOSTNAME, PORT)
+        http.raw_semian_options[:error_threshold].times do
+          http.get("/500")
+        end
+        assert_raises Net::CircuitOpenError do
```
sirupsen: As discussed earlier, what might be counter-intuitive (to some) about this is that normally on `5xx` you just receive a status code (not an exception). With this patch, you'll get exceptions on `5xx`. Most code (that handles connection failures) should handle this well—but some might not.

I don't think that's a huge deal, given you have to handle those conditions—but it's worth keeping in mind. I don't think people will find one ultimately more intuitive than the other, just want to make clear that this is a choice.

test/net_http_test.rb
```diff
@@ -312,6 +312,43 @@ def test_adding_custom_errors_do_trip_circuit
     end
   end
 
+  def test_5xxs_trip_circuit_when_fatal_server_flag_enabled
+    options = proc do |host, port|
+      {
+        tickets: 2,
```
sirupsen: I'd consider disabling tickets for this test, since you're not testing them. You can do this by setting `tickets: false`.

sirupsen:
@jpittis @methodmissing can you sign off on this as well? 👀

test/net_http_test.rb
albiere: True. I like the idea of having semian failing those 5xxs quickly and returning the 5xx rather than raising an exception. It feels like that's exactly what developers expect to get when using net/http.

sirupsen: That's the model I went for in the MySQL adapter with `#ping`, where it'll return `false` if the circuit is open as the fallback—not raise an exception. I don't have a strong opinion on this.

sirupsen:
Bump, @jpittis / @methodmissing would one of you be able to review this today/tomorrow for Edgescale?

PR #150, opened 2017-06-15
circuit_breaker: thread-safe by default

sirupsen:
This is a stab at making Semian thread-safe for circuit breakers. It's not absolutely perfect, I'm sure—but I'm reasonably confident this is _good enough_ (and it's hard to do more than that with the limited concurrency tooling for Ruby).

This is coming out of a need for internal applications to adopt Semian for Sidekiq and Puma.

**How did you determine it's thread-safe?** I went through the logic from boot (creation of the resource) to checking the circuit breaker. The major culprit for races is around adding entries to the sliding window and trimming it. The rest of the races fixed are fairly harmless.

**Why didn't you make bulkheads thread-safe as well?** As outlined in the README, this is not trivial. However, I believe that for most applications—and especially threaded ones—circuit breakers will be enough if you have a reasonably low timeout (see the README for more details about why that is). Thus the recommendation is to disable bulkheads for threaded apps.

**Why did you enable thread-safety by default?** It's not a breaking change, but it may have performance implications. Semian is a major hotspot, so we want a way to disable it, which I'll use in e.g. Shopify Core. People can disable it as desired.

**Why are there no new tests?** Testing for race conditions in Ruby is very tricky, and generally not worth the effort—it brings as much confidence in your ability to write sketchy signaling tests as it does actually test for races. The existing test suite is quite good, and all the mutex code is hit.

**What's with all this `::Simple` and `::ThreadSafe` stuff?** In preparation in the past for different adapters (e.g. a SysV-based circuit breaker), this was all abstracted out. This actually serves me nicely now. I don't want to create a ton of files, so I just kept it like this—we don't autoload in this library, so I think it's OK.

\cc @gigr @shelsen @meagar @jacobbednarz

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,9 +127,16 @@ client = Redis.new(semian: {
 })
 ```
 
-#### Warning: Thread unsafe
+#### Thread Safety
 
-Semian is currently known to be thread-unsafe, and is intended to be used by *separate processes*.
+Semian's circuit breaker implementation is thread-safe by default as of
+`v0.7.0`. If you'd like to disable it for performance reasons, pass
+`thread_safety_disabled: true` to the resource options.
+
+Bulkheads should be disabled (pass `tickets: false`) in a threaded environment
+(e.g. Puma or Sidekiq), but can safely be enabled in non-threaded environments
+(e.g. Resque and Unicorn). As described in this document, circuit breakers alone
+should be adequate in most environments with reasonably low timeouts.
 
 Internally, semian uses `SEM_UNDO` for several sysv semaphore operations:
 
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -1,6 +1,7 @@
 require 'forwardable'
 require 'logger'
 require 'weakref'
+require 'thread'
 
 require 'semian/version'
 require 'semian/instrumentable'
@@ -78,7 +79,6 @@ require 'semian/simple_state'
 #    end
 #
 # This is the same as the previous example, but overrides the timeout from the default value of 500 milliseconds to 1 second.
-#
 module Semian
   extend self
   extend Instrumentable
@@ -235,6 +235,7 @@ module Semian
     circuit_breaker = options.fetch(:circuit_breaker, true)
     return unless circuit_breaker
     raise ArgumentError unless required_keys?([:success_threshold, :error_threshold, :error_timeout], options)
+    implementation = options[:thread_safety_disabled] ? ::Semian::Simple : ::Semian::ThreadSafe
 
     exceptions = options[:exceptions] || []
     CircuitBreaker.new(
@@ -243,7 +244,7 @@ module Semian
       error_threshold: options[:error_threshold],
       error_timeout: options[:error_timeout],
       exceptions: Array(exceptions) + [::Semian::BaseError],
-      implementation: ::Semian::Simple,
+      implementation: implementation,
     )
   end
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -115,3 +115,6 @@ Style/CaseIndentation:
 Lint/UselessAssignment:
   Exclude:
     - 'test/adapter_test.rb'
+
+Style/ModuleLength:
+  Enabled: false
--- a/lib/semian/circuit_breaker.rb
+++ b/lib/semian/circuit_breaker.rb
@@ -43,7 +43,7 @@ module Semian
     end
 
     def mark_failed(_error)
-      push_time(@errors, duration: @error_timeout)
+      push_time(@errors)
       if closed?
         open if error_threshold_reached?
       elsif half_open?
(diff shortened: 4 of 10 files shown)


lib/semian.rb
```diff
@@ -150,10 +154,12 @@ def to_s
   #
   # Returns the registered resource.
   def register(name, **options)
-    circuit_breaker = create_circuit_breaker(name, **options)
-    bulkhead = create_bulkhead(name, **options)
-
-    resources[name] = ProtectedResource.new(bulkhead, circuit_breaker)
+    @lock.synchronize do
```
sirupsen: Since we register resources so infrequently (just boot), I think it's OK to have the lock even if we disable thread safety and have contention then.

README.md
```diff
@@ -127,9 +127,16 @@ client = Redis.new(semian: {
...
+#### Thread Safety
 
-Semian is currently known to be thread-unsafe, and is intended to be used by *separate processes*.
+Semian's circuit breaker implementation is thread-safe by default as of
+`v0.7.0`. If you'd like to disable it for performance reasons, pass
+`thread_safety_disabled: true` to the resource options.
+
+Bulkheads should be disabled (pass `tickets: false`) in a threaded environment
+(e.g. Puma or Sidekiq), but can safely be enabled in non-threaded environments
+(e.g. Resque and Unicorn). As described in this document, circuit breakers along
```
shelsen: typo: `alone`

lib/semian.rb
```diff
@@ -79,6 +80,7 @@
 #
 # This is the same as the previous example, but overrides the timeout from the default value of 500 milliseconds to 1 second.
 #
+# rubocop:disable ModuleLength
```
casperisfine: I'd disable that cop entirely in the YAML

sirupsen: Done

lib/semian/test_helpers.rb
```diff
@@ -0,0 +1,36 @@
+module Semian
+  module TestHelpers
+    Error = Class.new(StandardError)
+
+    def open_circuit!(resource)
+      error_count = resource.circuit_breaker.error_count_threshold
+      error_count.times { trigger_error!(resource) }
+    end
+
+    def half_open_cicuit!(resource)
```
gigr: Should this be `half_open_circuit`?

sirupsen: Oops, included this by accident. Will remove.

sirupsen:
@kirs if you have time to give this some 👀 !

jacobbednarz:
I'm far from a thread safety expert and I only needed to tango with this beast once in the past. To test whether some of our code paths were thread safe, I took the approach from http://rockyj.in/2014/08/16/test_thread_safety_with_concurrent_ruby.html and wrote some tests against it for our particular use case - not super thorough but it helped answer some questions. Would love to hear about some of the testing/tooling you have used here or in the future for it.

In saying that, I can't see anything that jumps out as a thread safety concern here. My usual 🚨  goes off when I see any operator in conjunction with the equals operator (i.e `||=` or `+=`) since they boil down to two operations which could end up being race conditions. `Array#<<` is another one which you've addressed but isn't usually a big deal since [GIL assists in making it atomic...to an extent](https://www.jstorimer.com/blogs/workingwithcode/8100871-nobody-understands-the-gil-part-2-implementation).

We're running unicorn in production and only just dipping our toes into the sidekiq world so this would be a nice addition for us too 😃

lib/semian.rb
```diff
@@ -150,10 +152,12 @@ def to_s
   #
   # Returns the registered resource.
   def register(name, **options)
-    circuit_breaker = create_circuit_breaker(name, **options)
-    bulkhead = create_bulkhead(name, **options)
-
-    resources[name] = ProtectedResource.new(bulkhead, circuit_breaker)
+    @lock.synchronize do
+      return resources[name] if resources[name]
```
sirupsen: If you acquire the lock the first time, you'd set it—any other thread that'd race in here (because of the `||` clause in `retrieve_or_register`) would then receive the first resource created when it got the lock (but we want to wait for the first one that races here to return it).

Maybe I'm not understanding you correctly, but I am fairly sure this doesn't break any functionality?

sirupsen:
@jacobbednarz yeah, this is my impression as well. Although, to be honest, I'm not completely sure how reliably you'll be able to determine whether something is "one instruction", e.g. if you disassemble `omg << "what"`, you'll see a `PUT` and it calling the `<<` method in two separate instructions 🤷‍♂️ I don't think it really matters, since you're not comparing a value (with e.g. `+= 1` you'd need a CAS), but it's somewhat difficult to reason about—so even in that case I've opted for a lock here (basically anything that mutates).

lib/semian.rb
jpittis: The code looks good to me from a safety point of view. 👍 

The issue is that this changes the behaviour of the public API `register`.
Before, future calls to register would overwrite, now they don't.

sirupsen: OH, I see what you mean. Good point. Nice catch. I'll fix that in `retrieve_or_register` tomorrow.

sirupsen:
Released in 0.7.0

@gigr @shelsen should be ready for you 👍

PR #151, opened 2017-06-15
change name delegator from bulkhead to circuit_breaker

jpittis:
Before now, `:name` was delegated to `Bulkhead`. Because both bulkheads and circuit breakers are now optional, it makes sense that `:name` lives in `ProtectedResource`.

Before this change, any attempt to grab a resources name if bulkheads were turned off would result in an exception.

I also added a test to make sure no one initializes a resource without a circuit breaker nor a bulkhead.

r: @sirupsen

### DIFF
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -153,7 +153,11 @@ module Semian
     circuit_breaker = create_circuit_breaker(name, **options)
     bulkhead = create_bulkhead(name, **options)
 
-    resources[name] = ProtectedResource.new(bulkhead, circuit_breaker)
+    if circuit_breaker.nil? && bulkhead.nil?
+      raise ArgumentError, 'Both bulkhead and circuitbreaker cannot be disabled.'
+    end
+
+    resources[name] = ProtectedResource.new(name, bulkhead, circuit_breaker)
   end
 
   def retrieve_or_register(name, **args)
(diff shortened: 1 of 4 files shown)


lib/semian.rb
```diff
@@ -153,7 +153,11 @@ def register(name, **options)
     circuit_breaker = create_circuit_breaker(name, **options)
     bulkhead = create_bulkhead(name, **options)
 
-    resources[name] = ProtectedResource.new(bulkhead, circuit_breaker)
+    if circuit_breaker.nil? && bulkhead.nil?
+      raise ArgumentError, 'Both bulkhead and circuitbreaker configured off.'
```
sirupsen: Disabled might be more natural here in the error "configured off" sounds a little strange.

lib/semian/protected_resource.rb
```diff
@@ -2,13 +2,14 @@ module Semian
   class ProtectedResource
     extend Forwardable
 
-    def_delegators :@bulkhead, :destroy, :count, :semid, :tickets, :registered_workers, :name
```
sirupsen: If it's not too much trouble to create a failing test case for this, that'd be great

sirupsen:
Let's release https://github.com/Shopify/semian/pull/150 and this one together in 0.7.0

jpittis:
LGTU @sirupsen? I'd be pleased to 🚢 this after we've fuzzy tested the thread safe stuff.

PR #152, opened 2017-06-20
Make configuration of dependencies (mysql, redis) flexible

albiere:
Semian currently expects the dependencies to run all locally. This PR makes the configuration of the dependencies more flexible. 

The hosts and ports for the dependencies can all be configured within `/test/config/hosts.yml`. This PR makes it easier for developers to use tools like docker and docker-compose locally in development.

In order to illustrate this scenario, have a look at the following `docker-compose.yml' and `dockerfile` below:

```yaml
version: '3'
services:
  mysql:
    image: mysql:8.0.1
    ports:
      - 3306
    environment:
      - MYSQL_ALLOW_EMPTY_PASSWORD=yes
  redis:
    image: redis:3.0.7
    ports:
      - 6379
  toxiproxy:
    image: shopify/toxiproxy
    ports:
      - 8474
      - 13307
    depends_on:
      - mysql
      - redis
  library:
    build: .
    command: tail -f /dev/null
    volumes:
      - .:/home/semiandev
    ports:
      - 31050
      - 31150
    depends_on:
      - toxiproxy
    extra_hosts:
      - 'library:127.0.0.1'
```

```dockerfile
FROM ruby:2.4.1

RUN mkdir -p /home/semiandev

ADD . .
RUN gem install bundler && bundle

WORKDIR /home/semiandev

CMD ["/bin/bash"]
```

By executing `docker-compose up`, we can have all dependencies running in a separate container. The only requirement here is to change the `hosts.yml` file accordantly.

```yaml
toxiproxy:
  host: toxiproxy
  port: 8474

mysql:
  host: mysql
  port: 3306
  toxic_port: 13306

redis:
  host: redis
  port: 6379
  toxic_port: 16379

server:
  host: library
  port: 31050
  toxic_port: 31051
```

Now you could simply run `docker exec -it semian_library_1 bundle exec rake test` and everything would still be running just fine.


test/config/hosts.yml
```diff
@@ -0,0 +1,18 @@
...
+  host: 127.0.0.1
+  port: 3306
+  toxic_port: 13306
+
+redis:
+  host: 127.0.0.1
+  port: 6379
+  toxic_port: 16379
+
+server:
```
sbfaulkner: nitpicky naming thing... should this be `app` or `thin` or something?

sbfaulkner: (ie. instead of `server`)

test/config/semian_test_config.rb
```diff
@@ -0,0 +1,38 @@
...
+      def define_helper_method(service, attribute)
+        __send__(:define_method, "#{service}_#{attribute}") do
+          SemianTestConfig[service].fetch(attribute)
+        end
+      end
+    end
+
+    SemianTestConfig.all.keys.each { |service| define_helper_methods(service) }
+  end
+end
```
sbfaulkner: this seems like overkill, BUT... results in nice clean  helpers and tests :-)

test/config/hosts.yml
```diff
@@ -0,0 +1,18 @@
...
+
+mysql:
+  host: 127.0.0.1
+  port: 3306
+  toxic_port: 13306
+
+redis:
+  host: 127.0.0.1
+  port: 6379
+  toxic_port: 16379
```
sirupsen: I'd use `toxiproxy_port`, as its more technically correct.

test/config/hosts.yml
```diff
@@ -0,0 +1,18 @@
+toxiproxy:
+  host: 127.0.0.1
```
sirupsen: To be consistent with the Toxiproxy vocabulary, I'd use `upstream_host` and `upstream_port`.

test/config/semian_test_config.rb
sirupsen: I think this is overkill. IMO let's remove this complexity. Let's require this class in `test_helper` and use `SemianTestConfig[:mysql_toxiproxy_host]` and so on.

test/test_helper.rb
```diff
@@ -12,24 +12,36 @@
...
-    listen: 'localhost:16379',
-  },
-  {
-    name: 'semian_test_net_http',
-    upstream: 'localhost:31050',
-    listen: 'localhost:31051',
-  },
-])
+
+class ToxiproxyConfig
```
sirupsen: Running all this code in the body of the class is pretty sketchy. If we remove the nice helpers, we can move this to the top-level namespace as well.

albiere:
@sirupsen Could you please have another look at this? :)

test/config/semian_config.rb
```diff
@@ -0,0 +1,29 @@
...
+    def flatten_keys(entries, object = {}, parent_key = '')
+      entries.each_with_object(object) do |(key, value), hash|
+        next flatten_keys(value, hash, key) if value.is_a?(Hash)
+
+        hash["#{parent_key}_#{key}"] = value
+        hash
+      end
+    end
+  end
+end
```
sbfaulkner: @albiere I think it's still a bit over-complicated...

why not just a simple Hash... 

```
CONFIG_FILE = File.expand_path('../hosts.yml', __FILE__)
SemianConfig = YAML.load_file(CONFIG_FILE)
```

with the flattened config...

```
toxiproxy_upstream_host: 127.0.0.1
toxiproxy_upstream_port: 8474

mysql_host: 127.0.0.1
mysql_port: 3306
mysql_toxiproxy_port: 13306

redis_host: 127.0.0.1
redis_port: 6379
redis_toxiproxy_port: 16379

http_host: 127.0.0.1
http_port: 31050
http_toxiproxy_port: 31051
```

albiere: hahaha... I'm definitely over-complicating things. Thanks!

PR #153, opened 2017-06-22
Minor version bump 0.7.1

andrewlouis93:
- Includes the behaviour to [enable open circuiting on 5xxs conditionally](https://github.com/Shopify/semian/pull/149)
- Also includes the [configurable dependency hosts changes](https://github.com/Shopify/semian/pull/152)

r/ @albiere @sirupsen @jpittis

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+# v0.7.1
+
+*  Feature: Add the behaviour to enable open circuiting on 5xxs conditionally  (#149)
+*  Refactor: Configurable hosts for Semian's development dependencies (#152)
+
+# v0.7.0
+
+* Thread-safety for circuit breakers by default (#150)
+* Fix bug where calling name on a protected resource without a semaphore would fail (#151)
+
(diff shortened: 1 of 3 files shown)


jpittis:
Looks like the changelog hasn't been updated in a while.
I feel like the 5xx feature should be recorded somewhere... maybe under http adapter section of the readme?

andrewlouis93:
@jpittis Just updated the Changelog and README, what do you think?

sirupsen:
Do you mind adding 0.7.0 notes as well?

You can add:

```
0.7.0

* Thread-safety for circuit breakers by default
* Fix bug where calling name on a protected resource without a semaphore would fail
```

andrewlouis93:
@sirupsen Added this in there - those are the right PR IDs yeah? 
```
# v0.7.0
 
 * Thread-safety for circuit breakers by default (#150)
 * Fix bug where calling name on a protected resource without a semaphore would fail (#151)
```

sirupsen:
Remember to create a tag in master.

PR #155, opened 2017-07-04
Reflect the correct syscall in semctl error

dalehamel:
@ibawt PTAL

tl;dr i must have copy-pasta'd this, because this is clearly the wrong syscall that we are reporting here.

cc @sirupsen @csfrancis


dalehamel:
for helping to track down https://github.com/Shopify/cloudplatform/issues/484#issuecomment-312890985

PR #161, opened 2017-07-12
More reliable way to measure wait time.

simon-shopify:
This moves the benchmarking code for ticket acquisition as close as possible to the system call to eliminate anything that could artificially inflate the numbers.

### DIFF
--- a/ext/semian/resource.c
+++ b/ext/semian/resource.c
@@ -63,7 +63,12 @@ semian_resource_acquire(int argc, VALUE *argv, VALUE self)
     }
   }
 
-  return rb_ensure(rb_yield, self, cleanup_semian_resource_acquire, self);
+  VALUE wait_time = Qnil;
+  if (res.wait_time >= 0) {
+    wait_time = LONG2NUM(res.wait_time);
+  }
+
+  return rb_ensure(rb_yield, wait_time, cleanup_semian_resource_acquire, self);
 }
 
 VALUE
@@ -219,6 +224,7 @@ semian_resource_initialize(VALUE self, VALUE id, VALUE tickets, VALUE quota, VAL
   ms_to_timespec(c_timeout * 1000, &res->timeout);
   res->name = strdup(c_id_str);
   res->quota = c_quota;
+  res->wait_time = -1;
 
   // Initialize the semaphore set
   initialize_semaphore_set(res, c_id_str, c_permissions, c_tickets, c_quota);
--- a/ext/semian/sysv_semaphores.c
+++ b/ext/semian/sysv_semaphores.c
@@ -1,4 +1,5 @@
 #include "sysv_semaphores.h"
+#include <time.h>
 
 static key_t
 generate_key(const char *name);
@@ -11,6 +12,9 @@ wait_for_new_semaphore_set(key_t key, long permissions);
 
 static void
 initialize_new_semaphore_values(int sem_id, long permissions);
+
+static long
+diff_timespec_ms(struct timespec *end, struct timespec *begin);
 
 // Generate string rep for sem indices for debugging puproses
 static const char *SEMINDEX_STRING[] = {
@@ -142,11 +146,20 @@ acquire_semaphore(void *p)
 {
   semian_resource_t *res = (semian_resource_t *) p;
   res->error = 0;
+  res->wait_time = -1;
 #ifdef DEBUG
   print_sem_vals(res->sem_id);
 #endif
+
+  struct timespec begin, end;
+  int benchmark_result = clock_gettime(CLOCK_MONOTONIC, &begin);
   if (perform_semop(res->sem_id, SI_SEM_TICKETS, -1, SEM_UNDO, &res->timeout) == -1) {
     res->error = errno;
+  }
+  if (benchmark_result == 0) {
+    if (clock_gettime(CLOCK_MONOTONIC, &end) == 0) {
+      res->wait_time = diff_timespec_ms(&end, &begin);
+    }
   }
   return NULL;
 }
@@ -230,4 +243,12 @@ wait_for_new_semaphore_set(key_t key, long permissions)
   }
 
   return sem_id;
+}
+
+static long
+diff_timespec_ms(struct timespec *end, struct timespec *begin)
+{
+  long end_ms = (end->tv_sec * 1e3) + (end->tv_nsec / 1e6);
+  long begin_ms = (begin->tv_sec * 1e3) + (begin->tv_nsec / 1e6);
+  return end_ms - begin_ms;
 }
--- a/ext/semian/resource.h
+++ b/ext/semian/resource.h
@@ -10,6 +10,7 @@ Functions here are associated with rubyland operations.
 #include "sysv_semaphores.h"
 
 // Ruby variables
+ID id_wait_time;
 ID id_timeout;
 int system_max_semaphore_count;
 
--- a/ext/semian/semian.c
+++ b/ext/semian/semian.c
@@ -53,6 +53,7 @@ void Init_semian()
   rb_define_method(cResource, "reset_registered_workers!", semian_resource_reset_workers, 0);
   rb_define_method(cResource, "unregister_worker", semian_resource_unregister_worker, 0);
 
+  id_wait_time = rb_intern("wait_time");
   id_timeout = rb_intern("timeout");
 
   if (semctl(0, 0, SEM_INFO, &info_buf) == -1) {
--- a/ext/semian/types.h
+++ b/ext/semian/types.h
@@ -35,6 +35,7 @@ typedef struct {
   uint64_t key;
   char *strkey;
   char *name;
+  long wait_time;
 } semian_resource_t;
 
 #endif // SEMIAN_TYPES_H
--- a/lib/semian/protected_resource.rb
+++ b/lib/semian/protected_resource.rb
@@ -46,20 +46,14 @@ module Semian
       if @bulkhead.nil?
         yield self
       else
-        acquisition_start = current_milliseconds
-        @bulkhead.acquire(timeout: timeout) do
-          wait_time = current_milliseconds - acquisition_start
-          Semian.notify(:success, self, scope, adapter, wait_time: wait_time)
+        @bulkhead.acquire(timeout: timeout) do |wait_time|
+          Semian.notify(:success, self, scope, adapter, wait_time)
           yield self
         end
       end
     rescue ::Semian::TimeoutError
       Semian.notify(:busy, self, scope, adapter)
       raise
-    end
-
-    def current_milliseconds
-      Process.clock_gettime(Process::CLOCK_MONOTONIC, :millisecond)
     end
   end
 end
--- a/test/instrumentation_test.rb
+++ b/test/instrumentation_test.rb
@@ -40,7 +40,7 @@ class TestInstrumentation < Minitest::Test
 
   def test_success_instrumentation_wait_time
     hit = false
-    subscription = Semian.subscribe do |*_, wait_time:|
+    subscription = Semian.subscribe do |*_, wait_time|
       hit = true
       assert(wait_time.is_a?(Integer))
     end


ext/semian/sysv_semaphores.h
```diff
@@ -84,7 +84,13 @@ create_semaphore(int key, long permissions, int *created);
...
 // The call may be timed or untimed
 int
-perform_semop(int sem_id, short index, short op, short flags, struct timespec *ts);
+perform_semop(
+  int sem_id,
+  short index,
+  short op,
+  short flags,
+  struct timespec *ts,
+  int *elapsed_ms);
```
csfrancis: Should be `wait_time_ms`, I think.

ext/semian/sysv_semaphores.c
```diff
@@ -80,19 +81,27 @@ set_semaphore_permissions(int sem_id, long permissions)
...
+    clock_gettime(CLOCK_MONOTONIC, &begin);
+  }
   if (ts) {
     return semtimedop(sem_id, &buf, 1, ts);
   } else {
     return semop(sem_id, &buf, 1);
   }
+  if (wait_time_ms) {
+    clock_gettime(CLOCK_MONOTONIC, &end);
+    *wait_time_ms = begin.tv_sec * 1e3 + begin.tv_nsec / 1e6 - end.tv_sec * 1e3 - end.tv_nsec / 1e6;
```
csfrancis: I find this harder to understand without brackets.

ext/semian/resource.c
```diff
@@ -63,7 +64,10 @@ semian_resource_acquire(int argc, VALUE *argv, VALUE self)
     }
   }
 
-  return rb_ensure(rb_yield, self, cleanup_semian_resource_acquire, self);
+  VALUE options = rb_hash_new();
+  rb_hash_aset(options, ID2SYM(rb_intern("wait_time")), INT2NUM(wait_time_ms));
```
csfrancis: It would be more #webscale to cache the `rb_intern` in a global.

ext/semian/sysv_semaphores.c
dalehamel: This line isn't very descriptive, I'd recommend refactoring to a static function to better describe the nature of this (relatively complicated) timing operation.

Alternatively, if you're concerned about speed, an inline could make sense - but i'd been avoiding them where possible.

ext/semian/sysv_semaphores.c
```diff
@@ -80,19 +81,27 @@ set_semaphore_permissions(int sem_id, long permissions)
...
+perform_semop(int sem_id, short index, short op, short flags, struct timespec *ts, int *wait_time_ms)
 {
   struct sembuf buf = { 0 };
 
   buf.sem_num = index;
   buf.sem_op  = op;
   buf.sem_flg = flags;
 
+  struct timespec begin, end;
+  if (wait_time_ms) {
```
dalehamel: I'd suggest moving both this and the timing finish to a helper function, as this complicates the control flow for a bit

ext/semian/resource.c
```diff
@@ -54,7 +54,8 @@ semian_resource_acquire(int argc, VALUE *argv, VALUE self)
   }
 
   /* release the GVL to acquire the semaphore */
-  acquire_semaphore_without_gvl(&res);
+  int wait_time_ms = 0;
```
dalehamel: please move declaration to the top

dalehamel: actually, why don't we just add this as a member to the struct?

ext/semian/sysv_semaphores.c
```diff
@@ -80,19 +81,27 @@ set_semaphore_permissions(int sem_id, long permissions)
 }
 
 int
-perform_semop(int sem_id, short index, short op, short flags, struct timespec *ts)
+perform_semop(int sem_id, short index, short op, short flags, struct timespec *ts, int *wait_time_ms)
```
dalehamel: I think it could make sense to just define a type which includes both timespec and an additional int for wait_time_ms, though either way the call signature would have to change.

I think having a single argument for timing related info is cleaner, but I also thing the separate variable might be more apparent. 

something to watch though, as this function alright takes a lot of parameters.

ext/semian/sysv_semaphores.c
```diff
@@ -130,22 +139,29 @@ get_semaphore(int key)
...
 }
 
+struct acquire_semaphore_args {
+  semian_resource_t *res;
+  int *wait_time_ms;
+};
+
 void *
-acquire_semaphore_without_gvl(void *p)
+acquire_semaphore_without_gvl(semian_resource_t *res, int *wait_time_ms)
```
dalehamel: Is the main purpose of this to figure out how long we are stuck waiting on a syscall without the GVL?

ext/semian/sysv_semaphores.c
```diff
@@ -130,22 +139,29 @@ get_semaphore(int key)
...
 {
-  WITHOUT_GVL(acquire_semaphore, p, RUBY_UBF_IO, NULL);
+  struct acquire_semaphore_args args = { .res = res, .wait_time_ms = wait_time_ms };
```
dalehamel: might be cleaner to just include this as a member of semian_resource_t - and we actually own that type, so we don't have to do any sort of struct-wrapping.

ext/semian/sysv_semaphores.c
```diff
@@ -130,22 +139,29 @@ get_semaphore(int key)
...
+  WITHOUT_GVL(acquire_semaphore, &args, RUBY_UBF_IO, NULL);
   return NULL;
 }
 
 static void *
 acquire_semaphore(void *p)
 {
-  semian_resource_t *res = (semian_resource_t *) p;
+  struct acquire_semaphore_args *args = p;
+  semian_resource_t *res = args->res;
```
dalehamel: why is this declaration needed?

ext/semian/sysv_semaphores.h
```diff
@@ -84,7 +84,13 @@ create_semaphore(int key, long permissions, int *created);
 // Wrapper to performs a semop call
 // The call may be timed or untimed
 int
-perform_semop(int sem_id, short index, short op, short flags, struct timespec *ts);
+perform_semop(
```
dalehamel: Yeah, another sign the call sig is getting complicated

ext/semian/sysv_semaphores.h
```diff
@@ -104,7 +110,7 @@ get_semaphore(int key);
 
 // Decrements the ticket semaphore within the semaphore set
 void *
-acquire_semaphore_without_gvl(void *p);
+acquire_semaphore_without_gvl(semian_resource_t *p, int *wait_time_ms);
```
dalehamel: Are we sure it will always be a semian_resource_t? (haven't traced the code).

simon-shopify:
> Try and come up with a test case where you could prove the timing was inaccurate

I'll try to explain a bit more what I am doing. So I suspect the timing is accurate. However, the numbers are far off. Given that the original PR was a quick one and even I thought it was a bit sloppy, I am taking the time to do it right on the off chance that something like

* the GC,
* a rogue statsd event fired,
* some other syscall like [the cleanup semop](https://github.com/Shopify/semian/blob/4606b13ff08bd542e0fef6e19ca61d48f44e5c90/ext/semian/resource.c#L242)

is causing this delay and I can save myself a few days of work. It is really about improving the code. If you feel that this PR (once comments are addressed) is making Semian worse, I don't have a proof that it makes the measurements more accurate, so I'll just close it.

alexsnaps:
Random thought, could the GVL be contended when we reacquire it (e.g. GC)?

csfrancis:
> Random thought, could the GVL be contended when we reacquire it (e.g. GC)?

That seems unlikely for Unicorn workers since they're effectively single threaded (there are some background threads that do things like New Relic). But perhaps GC.

dalehamel:
> I'll try to explain a bit more what I am doing. So I suspect the timing is accurate. However, the numbers are far off.

Can you clarify what you mean? For me (without context) it sounds like you're contradicting yourself.

> Given that the original PR was a quick one and even I thought it was a bit sloppy

I think it's great to get scrappy WIP PRs up, I understood what you are trying to do and just thought I would offer tips on how to integrate this cleanly - I hope I didn't deter you.

simon-shopify:
> Can you clarify what you mean? For me (without context) it sounds like you're contradicting yourself.

What I meant is I think the numbers reflect reality, but these numbers are far off what I expected. I need to investigate

* if this discrepancy is real;
* if so, what is the cause.

I need to eliminate the uncertainty of the former before doing the latter, otherwise I might waste my time profiling something that doesn't exist.

> I think it's great to get scrappy WIP PRs up, I understood what you are trying to do and just thought I would offer tips on how to integrate this cleanly - I hope I didn't deter you.

I appreciate the tips, I just want to clear the issue with proving the timing are off before spending more time on this.

simon-shopify:
Ok, I updated the PR. I extracted the concept of **operation** from the concept of **resource**. It contains information about a single Semian operation like whether it errored and performance measurements.

Can I get another round of reviews? I'd appreciate if you kept the feedback as a list of action items, and make clear what is a blocker or not to merge this PR.

ext/semian/resource.c
```diff
@@ -239,9 +256,7 @@ cleanup_semian_resource_acquire(VALUE self)
 {
   semian_resource_t *res = NULL;
   TypedData_Get_Struct(self, semian_resource_t, &semian_resource_type, res);
-  if (perform_semop(res->sem_id, SI_SEM_TICKETS, 1, SEM_UNDO, NULL) == -1) {
-    res->error = errno;
```
csfrancis: Why did you get rid of this?

simon-shopify:
> I feel that this implementation is adding too much code for what it needs to do. Do we really need to support multiple benchmark topics when we're only currently just using one?
> 
> Why not just add a wait_time member to semian_resource_t?

Sure, I can change it back. What do you think, @dalehamel ?

ext/semian/sysv_semaphores.c
```diff
@@ -130,24 +131,28 @@ get_semaphore(int key)
...
+void
+acquire_semaphore_without_gvl(semian_operation_t *operation)
 {
-  WITHOUT_GVL(acquire_semaphore, p, RUBY_UBF_IO, NULL);
-  return NULL;
+  WITHOUT_GVL(acquire_semaphore, operation, RUBY_UBF_IO, NULL);
 }
 
 static void *
 acquire_semaphore(void *p)
```
dalehamel: I assume this was originally a semian_resource_t? If so, then this should be a fair bit simpler if you just add the timing as a member to that struct.

dalehamel:
> Why not just add a wait_time member to semian_resource_t?

Yeah I think I suggested something along these lines originally. That's what I was trying to get at with:

> Can some of the arguments that have been added instead just become struct members, where a struct is already being passed especially if we own the struct type? For cases where we don't (timespec) should we wrap it with a custom type?

The code here is well written and well organized, but I think Scott's critique is on-point that just adding another member to this struct is probably a simpler way.

I think you can refactor your timing helper functions to just be static members of whatever file you are doing the timing in.

My apologies for not being able to give a very detailed review here. Please just take my comments as suggestions and not requirements.

simon-shopify:
Updated.

simon-shopify:
> I would prefer if we stuck to just using wait_time (and pass that to the block) instead of creating an unnecessary layer of indirection with benchmarks.

oops that was a remnant from the second version… next time when I'm not sure I'll keep old versions around.

ext/semian/resource.c
```diff
@@ -63,7 +63,12 @@ semian_resource_acquire(int argc, VALUE *argv, VALUE self)
     }
   }
 
-  return rb_ensure(rb_yield, self, cleanup_semian_resource_acquire, self);
+  VALUE options = rb_hash_new();
+  if (res.wait_time >= 0) {
+    rb_hash_aset(options, ID2SYM(id_wait_time), INT2NUM(res.wait_time));
+  }
+
+  return rb_ensure(rb_yield, options, cleanup_semian_resource_acquire, self);
```
csfrancis: Why pass a hash here when you could just pass `wait_time`?

simon-shopify:
Updated: turned `wait_time` from an option to a positonal parameter.

simon-shopify:
> My apologies for not being able to give a very detailed review here. Please just take my comments as suggestions and not requirements.

I guess that means :shipit: ?

PR #164, opened 2017-08-17
Make sure internal semaphore release is protected.

simon-shopify:
A code change moved `sem_meta_unlock` outside of the protected scope. This PR brings it back inside.

### DIFF
--- a/ext/semian/sysv_semaphores.c
+++ b/ext/semian/sysv_semaphores.c
@@ -63,9 +63,23 @@ initialize_semaphore_set(semian_resource_t* res, const char* id_str, long permis
     rb_raise(eInternal, "error incrementing registered workers, errno: %d (%s)", errno, strerror(errno));
   }
 
+  int state = 0;
   sem_meta_lock(res->sem_id); // Sets otime for the first time by acquiring the sem lock
-  configure_tickets(res->sem_id, tickets,  quota);
+
+  configure_tickets_args_t configure_tickets_args = (configure_tickets_args_t){
+    .sem_id = res->sem_id,
+    .tickets = tickets,
+    .quota = quota,
+  };
+  rb_protect(
+    configure_tickets,
+    (VALUE)&configure_tickets_args,
+    &state);
+
   sem_meta_unlock(res->sem_id);
+  if (state) {
+    rb_jump_tag(state);
+  }
 }
 
 void
--- a/ext/semian/tickets.c
+++ b/ext/semian/tickets.c
@@ -2,20 +2,19 @@
 
 // Update the ticket count for static ticket tracking
 static VALUE
-update_ticket_count(update_ticket_count_t *tc);
+update_ticket_count(int sem_id, int count);
 
 static int
 calculate_quota_tickets(int sem_id, double quota);
 
 // Must be called with the semaphore meta lock already acquired
-void
-configure_tickets(int sem_id, int tickets, double quota)
+VALUE
+configure_tickets(VALUE value)
 {
-  int state = 0;
-  update_ticket_count_t tc;
+  configure_tickets_args_t *args = (configure_tickets_args_t *)value;
 
-  if (quota > 0) {
-    tickets = calculate_quota_tickets(sem_id, quota);
+  if (args->quota > 0) {
+    args->tickets = calculate_quota_tickets(args->sem_id, args->quota);
   }
 
   /*
@@ -23,10 +22,10 @@ configure_tickets(int sem_id, int tickets, double quota)
     We need to throw an error if we set it to 0 during initialization.
     Otherwise, we back out of here completely.
   */
-  if (get_sem_val(sem_id, SI_SEM_CONFIGURED_TICKETS) == 0 && tickets == 0) {
+  if (get_sem_val(args->sem_id, SI_SEM_CONFIGURED_TICKETS) == 0 && args->tickets == 0) {
     rb_raise(eSyscall, "More than 0 tickets must be specified when initializing semaphore");
-  } else if (tickets == 0) {
-    return;
+  } else if (args->tickets == 0) {
+    return Qnil;
   }
 
   /*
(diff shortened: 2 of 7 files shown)


dalehamel:
> A code change moved sem_meta_unlock outside of the protected scope

@simon-shopify what was the change?

simon-shopify:
> @simon-shopify what was the change?

I think that was #120.

dalehamel:
@simon-shopify i'm going on vacation in an hour, don't have time to give you a proper review. 

Seems sane to me.

simon-shopify:
> @simon-shopify i'm going on vacation in an hour, don't have time to give you a proper review.

No worries, I'll ask someone else. Enjoy!

csfrancis:
@simon-shopify could you add a test for this?

simon-shopify:
> @simon-shopify could you add a test for this?

I'd like to, but I'm not sure how to do it. Is there a way to mock syscalls from Ruby? The issue is that it configures the ticket right after getting it so I don't have a chance to mess with it directly to trigger an error condition.

ext/semian/tickets.c
```diff
@@ -2,71 +2,65 @@
...
+update_ticket_count(int sem_id, int count);
 
 static int
 calculate_quota_tickets(int sem_id, double quota);
 
 // Must be called with the semaphore meta lock already acquired
-void
-configure_tickets(int sem_id, int tickets, double quota)
+VALUE
+configure_tickets(VALUE value)
```
sirupsen: Why did you need to change the signature?

ext/semian/sysv_semaphores.c
```diff
@@ -63,9 +63,23 @@ initialize_semaphore_set(semian_resource_t* res, const char* id_str, long permis
...
+  int state = 0;
   sem_meta_lock(res->sem_id); // Sets otime for the first time by acquiring the sem lock
-  configure_tickets(res->sem_id, tickets,  quota);
+
+  configure_tickets_args_t configure_tickets_args = (configure_tickets_args_t){
+    .sem_id = res->sem_id,
+    .tickets = tickets,
+    .quota = quota,
+  };
+  rb_protect(
```
sirupsen: [As far as I can gather](https://silverhammermba.github.io/emberb/c/#rb_protect) this is required because `configure_tickets` might raise, which has to be done within a `rb_protect` call to not segfault the VM?

simon-shopify:
> I don't see anything wrong, but can you provide some more context? How did this come up for us? How did you pin it down to this problem?

All of the resource semaphores are used (this is common in our job servers). We still use a number of tickets proportional to the number of workers with rescue. This causes a lot of thrashing on the semaphore limits on every deploy. In particular, in this case, a worker wanted to reduce the number of tickets because it was getting shut down.
* It grabbed the internal semaphore.
* Timed out as it tried to reduce the count.
* Raised.

Since the internal semaphore was not protected, it was never released, so the count could not be further adjusted.

ext/semian/tickets.c
simon-shopify: `rb_protect` can only call functions with the exact signature.

ext/semian/sysv_semaphores.c
simon-shopify: It is not to avoid a segfault. See the explanation below.

sirupsen:
What happens to that process that's trying to grab the internal lock and decrease the ticket count? Does it raise and abort abnormally now?

simon-shopify:
> What happens to that process that's trying to grab the internal lock and decrease the ticket count? Does it raise and abort abnormally with an exception?

Yes, in addition to leaving the semaphore locked.

simon-shopify:
I added a test.

sirupsen:
> Yes, in addition to leaving the semaphore locked.

Isn't it dangerous to keep the semaphore locked? What will unlock it? Why are we OK with raising and exiting when we failed to decrement, which is likely not super uncommon?

simon-shopify:
> Isn't it dangerous to keep the semaphore locked? What will unlock it? 

Sorry I was unclear. The bug is that the semaphore remains locked. This PR fixes it so it unlocks if there is a timeout.

> Why are we OK with raising and exiting when we failed to decrement, which is likely not super uncommon?

I don't think we are OK with it, but what is the alternative? I see it as graceful degradation.

sirupsen:
> I don't think we are OK with it, but what is the alternative? I see it as graceful degradation.

The alternative would seem to be to log, instead of raise. This doesn't really seem exceptional

simon-shopify:
> The alternative would seem to be to log, instead of raise. This doesn't really seem exceptional

Yes that could work as well. What do you think @csfrancis?

Note that this would not remove the need to `rb_protect` the semaphore since there are other reason it can raise.

sirupsen:
> Note that this would not remove the need to rb_protect the semaphore since there are other reason it can raise.

👍 Mostly concerned with common, non-exceptional exceptions being raised and aborting normal shutdown, potentially leading to other issues down the line.

csfrancis:
> Yes that could work as well. What do you think @csfrancis?

I'm a bit torn. On one hand I like raising because if this happens, the Semian ticket count is **not** configured the way the client expects. But on the other hand, subsequent calls to acquire the resource _should_ eventually end up reconfiguring the ticket count.

I'd prefer to leave it as-is, but if you want to change it to log that's fine with me as well.

simon-shopify:
Here is what I suggest:
* Merge this PR like this to fix the regression, but keep the existing behaviour.
* Open a second PR to turn the exception into a log when there is contention on the internal lock.
* Make a release.

test/protected_resource_test.rb
```diff
@@ -112,4 +112,43 @@ def test_responds_to_name_when_bulkhead_or_circuit_breaker_disabled
     assert_equal :no_bulkhead, Semian[:no_bulkhead].name
     assert_equal :no_circuit_breaker, Semian[:no_circuit_breaker].name
   end
+
+  def test_gracefully_fails_when_unable_to_decrement_ticket_count
```
csfrancis: Very nice test!

simon-shopify: I'm pretty satisfied with it, the only downside is that the timeout for the internal semaphore is 5s and not configurable, so it does take 5s.

csfrancis:
> Merge this PR like this to fix the regression, but keep the existing behaviour.

I would do this and cut a release now. Changing the behaviour to log is not as important as fixing this bug.

PR #171, opened 2017-11-20
Including Timeout error for mysql2 adapter

sribalakumar:
A MySQL Semian resource can be a bad resource if the queries are taking a longer time to respond and are exceeding the timeout options configured in database.yml of the application. Under such scenarios instead of waiting for a longer time and increasing the request queue it is good to have a circuit breaker in place and fail faster.

In the current implementation a Semian resource is not marked as failed when there is a MySQL timeout for the query.

Sample error trace for a long running query looks like -
  Mysql2::Error: [mysql_192.16.10.158:3306] Timeout waiting for a response from the last query. (waited 18 seconds): select sleep(30)
   ActiveRecord::StatementInvalid (Mysql2::Error: [mysql_192.16.10.158:3306] Lost connection to MySQL server during query: select sleep(30)):

### DIFF
--- a/test/mysql2_test.rb
+++ b/test/mysql2_test.rb
@@ -56,6 +56,31 @@ class TestMysql2 < Minitest::Test
     end
   end
 
+  def test_read_timeout_error_open_the_circuit
+    client = connect_to_mysql!
+
+    (ERROR_THRESHOLD).times do
+      assert_raises Mysql2::Error do
+        # Should raise an exception like -
+        # Mysql2::Error Exception: [mysql_testing] Timeout waiting for a response from the last query. (waited 2 seconds)
+        client.query('SELECT sleep(5)')
+      end
+    end
+
+    assert_raises Mysql2::CircuitOpenError do
+      client.query('SELECT sleep(5)')
+    end
+
+    # After Mysql2::CircuitOpenError check regular queries are working fine.
+    query_string = "1 + 1"
+    result = nil
+    Timecop.travel(ERROR_TIMEOUT + 1) do
+      result = client.query("SELECT #{query_string};")
+    end
+
+    assert_equal 2, result.first[query_string]
+  end
+
   def test_connect_instrumentation
     notified = false
     subscriber = Semian.subscribe do |event, resource, scope, adapter|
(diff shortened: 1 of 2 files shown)


sirupsen:
I'm extremely surprised this wasn't there before, perhaps it's because we often see our queries killed by `pt-kill` (in which case the connection is closed). Where do you configure your timeout for the client for this to take effect?

sribalakumar:
For the above error trace where a query is timed-out after 18s, timeout is configured in database.yml file of the rails app.

``` yml
adapter:  mysql2
....
....
pool: 5
reconnect: true
read_timeout: 18
```

sribalakumar:
@sirupsen Any updates, the following the strace when is strace value we can see it is setting a read timeout on the socket  -

sirupsen:
Sorry, been busy for Black Friday and Cyber Monday.

Can you add a Toxiproxy test that reproduces this error with the MySQL client?

sribalakumar:
Sure, will get it included over the weekend.

sribalakumar:
@sirupsen Have included Toxiproxy test and verified it to run successfully on ruby 2.3.1 with `gem mysql2 --version 0.4.10`.

Sorry for the delay in getting back with the test case :)

Travis check for the PR is failing in `net_http_test` today, is the build broken ?

sirupsen:
@kirs @csfrancis since you've been looking at MySQL error conditions internally quite a bit lately, can you review?

test/mysql2_test.rb
```diff
@@ -56,6 +56,31 @@ def test_query_errors_does_not_open_the_circuit
     end
   end
 
+  def test_read_timeout_error_open_the_circuit
+    client = connect_to_mysql!
+
+    (ERROR_THRESHOLD).times do
+      assert_raises Mysql2::Error do
+        # Should raise an exception like -
```
kirs: Why not encode this assumption into code, like:

```ruby
error = assert_raises Mysql2::Error do
  ...
end
assert_matches "Timeout waiting for a response from the last query", error.to_s
```

kirs:
@sribalakumar I'll be happy to merge this as soon as the CI passes :) Could you please sign the CLA agreement from the CI status?

sribalakumar:
@kirs CLA is completed. The Travis CI errors that I see are elsewhere in other places like net_http_test and grpc_test, is it a known problem ?

PR #175, opened 2018-02-11
[CI] Test against Ruby 2.x versions

nicolasleger:
Build already failed and broken on master https://travis-ci.org/Shopify/semian
Got the same errors


kirs:
Thanks for the PR! We've updated ruby versions in https://github.com/Shopify/semian/pull/181. I'll close this PR.

PR #176, opened 2018-03-05
 Introduce a Vagrantfile with Ubuntu Xenial

devlinzed:
This allows people to work on Semian from non-Linux machines.  The base image is Ubuntu Xenial, and Ruby 2.4.3 is installed from source using ruby-build.

Inside the image, Toxiproxy is not started because the deb package doesn't provide a systemd service.  This means toxiproxy-server has to be started before the tests can be run inside the virtual machine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ vendor/
 *.swp
 *.swo
 nohup.out
+.vagrant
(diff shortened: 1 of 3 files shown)


devlinzed:
@sirupsen You bet I can!  However, the tests seem to fail non-deterministically.  I'm working on sorting that out.  I don't think it's related to my changes.

sirupsen:
@devlinzed I'm really impressed with your work here. Thank you!

Yeah, I don't think Travis is your fault. Merge!

PR #182, opened 2018-03-14
Provide more details when raising ArgumentError over missing options

devlinzed:
Unfortunately, I'm not sure how to test this.  :{

### DIFF
--- a/test/semian_test.rb
+++ b/test/semian_test.rb
@@ -15,7 +15,7 @@ class TestSemian < Minitest::Test
   end
 
   def test_register_with_circuit_breaker_missing_options
-    assert_raises ArgumentError do
+    exception = assert_raises ArgumentError do
       Semian.register(
         :testing,
         error_threshold: 2,
@@ -23,15 +23,18 @@ class TestSemian < Minitest::Test
         bulkhead: false,
       )
     end
+    assert_equal exception.message,
+      "Missing required arguments for Semian: [:success_threshold]"
   end
 
   def test_register_with_bulkhead_missing_options
-    assert_raises ArgumentError do
+    exception = assert_raises ArgumentError do
       Semian.register(
         :testing,
         circuit_breaker: false,
       )
     end
+    assert_equal exception.message, "Must pass exactly one of ticket or quota"
   end
 
   def test_unsuported_constants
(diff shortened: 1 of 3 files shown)


test/semian_test.rb
```diff
@@ -15,7 +15,7 @@ def test_unsupported_acquire_yields
   end
 
   def test_register_with_circuit_breaker_missing_options
-    assert_raises ArgumentError, "CI stub" do
+    assert_raises ArgumentError do
```
jacobbednarz: Testing the exception message isn't straight forward and the only way I've had it work in the past is with two assertions. One for the `assert_raises` and the part `assert_equals`/`assert_match`. In your case:

```rb
def test_register_with_circuit_breaker_missing_options
  exception = assert_raises ArgumentError do
    Semian.register(
      :testing,
      error_threshold: 2,
      error_timeout: 5,
      bulkhead: false,
    )
  end
  assert_equal("CI stub", exception.message)
end
```

(I haven't used this in quite a while so YMMV)

I did manage to find others hitting this scenario in the minitest GitHub issue queue and one of the solutions was to add an assertion like `assert_raises_with_message` (see seattlerb/minitest#627) but it's been stagnating without any progress for a while so I wouldn't be holding my breathe on it.

devlinzed: Do you mind if I make a different exception type like `MissingOptionsError`?  I would need to decide whether it should inherit from `Semian::BaseError` or `ArgumentError` for compatibility.

devlinzed: @jacobbednarz I realized changing the exception type may require a major version bump, which I'd like to avoid.  For that reason I did go with your approach.  Let me know if I can :shipit:.

jacobbednarz: `ArgumentError` makes sense to me here. You're expecting a particular set of arguments and if that isn't met, you are raising an exception.

jacobbednarz:
Ping @sirupsen for a re-review here as the functionality has changed since his last look (especially regarding his feelings on the multiple assertions in one test).

devlinzed:
Would the alternative be having one test that `assert_raises`, and another that does `begin ... rescue ... assert end`?

jacobbednarz:
Generally, yes. I prefer to include as many assertions in one test that I need to feel confident the thing does what I expect it to but the maintainers of this might have other thoughts 😄

sirupsen:
OK with me 👍

PR #184, opened 2018-04-03
Handle DNS resolution failures on connect

csfrancis:
This allows Semian to track DNS resolution failures during `connect`. With the `redis` gem, this manifests as a `SocketError`.

Unfortunately, `hiredis-rb` masks the error by raising a generic `RuntimeError` on any connection failure (including DNS resolution): https://github.com/redis/hiredis-rb/blob/2b46b19dda98ba487e61884087b60d692bd0753e/ext/hiredis_ext/connection.c#L193. Also, the vendor'd `hiredis` gem also obscures the resolution error: https://github.com/redis/hiredis/blob/3d8709d19d7fa67d203a33c969e69f0f1a4eab02/net.c#L330-L335.

I don't really like this solution, but it's the best I can come up with, without having to make a bunch of upstream PRs. I wasn't able to come up with a test for the `hiredis` path, because I couldn't determine if there's a way to use both `redis` and `hiredis` gems separately in test.

### DIFF
--- a/lib/semian/redis.rb
+++ b/lib/semian/redis.rb
@@ -14,6 +14,7 @@ class Redis
 
   ResourceBusyError = Class.new(SemianError)
   CircuitOpenError = Class.new(SemianError)
+  ResolveError = Class.new(SemianError)
 
   alias_method :_original_initialize, :initialize
 
@@ -48,6 +49,7 @@ module Semian
 
     ResourceBusyError = ::Redis::ResourceBusyError
     CircuitOpenError = ::Redis::CircuitOpenError
+    ResolveError = ::Redis::ResolveError
 
     # The naked methods are exposed as `raw_query` and `raw_connect` for instrumentation purpose
     def self.included(base)
@@ -71,7 +73,14 @@ module Semian
     end
 
     def connect
-      acquire_semian_resource(adapter: :redis, scope: :connection) { raw_connect }
+      acquire_semian_resource(adapter: :redis, scope: :connection) do
+        begin
+          raw_connect
+        rescue SocketError, RuntimeError => e
+          raise ResolveError.new(semian_identifier) if e.message =~ /(can't resolve)|(name or service not known)/i
+          raise
+        end
+      end
     end
 
     private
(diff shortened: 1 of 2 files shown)


fw42:
Looks related to https://github.com/redis/hiredis-rb/issues/40 :-(

fw42:
Also https://github.com/Shopify/semian/pull/72 and https://github.com/Shopify/semian/pull/73

csfrancis:
I've updated the test to make it closer match the behaviour of #73.

sirupsen:
What about MySQL?

csfrancis:
From what I can tell, [MySQL DNS failures will raise a `Mysql2Error`](https://github.com/brianmario/mysql2/blob/bf227ace695b5e1b836d5b788a87dd98f7acb546/ext/mysql2/client.c#L438-L462). The client itself does not actually do any connecting -- it's done by the lower level libmysqlclient driver. In the case of an error, the `mysql_real_connect` call will return `NULL` and `mysql2` will raise.

It would be great if we could fix this at a lower level so we don't have to patch every adapter. One option would be to [use `resolve-replace.rb`](http://www.mccartie.com/2016/03/14/modify-your-ruby-app's-dns-lookup.html), but the problem with that is it only replaces the resolver for Ruby sockets. For libraries like hiredis and libmemcached that are written as C extensions, it would be ineffective because they're using `getaddrinfo` directly, and we're dependent on them actually raising the right exception to be handled by Semian.

sirupsen:
@csfrancis when we deploy service to service proxies, do you think it'd make sense to build DNS into that server with resiliency? It may be OK to wait with investing until that point.

csfrancis:
Do you mean build an actual DNS server into the service proxy? I don't think that would be worth it.

sirupsen:
Yeah, I meant a "proxying DNS server", I suppose. That said, this seems more like a client bug (not using the right exception) than something that a proxy would fix.

pushrax:
Depending on configuration, the proxy may end up being the resolver without actually being a DNS server.

PR #188, opened 2018-07-30
circuit: introduce half_open_resource_timeout

sirupsen:
When a circuit is busy (for those at Shopify, see https://github.com/Shopify/shopify/pull/164168) you may reduce a substantial amount of capacity if you have large timeouts (see https://github.com/Shopify/semian/pull/145). One way to circumvent this is to have very large `error_threshold` timeouts, as a multiple of your timeout. E.g. if your resource timeout is 30s, then an error_timeout of 30s will make sure you have 50% of capacity available, and 60s will make sure you have 25% available (assuming that the circuits trigger equally over the timeout period, which is not an unreasonable assumption in production). However, this has obvious drawbacks, making recovering from incidents take longer for the affected resource.

Instead, we can reduce the resource timeout in the `half_open?` state to achieve something similar. Normally, timeouts are to allow outlier queries to go through—in the face of an incident, we can eat that (we'd want to kill those anyway) and reduce the timeout dramatically while `half_open?` on the circuit.

The ugly thing about this, is that it couples the circuit with the resource. However, I think it makes more sense for the circuit to know about the resource, than for the resource to know about the circuit—that's why I chose to implement it this way.

This will need to be implemented per adapter, by implementing `with_resource_timeout`.

cc @Shopify/servcomm

### DIFF
--- a/lib/semian/circuit_breaker.rb
+++ b/lib/semian/circuit_breaker.rb
@@ -4,21 +4,22 @@ module Semian
 
     def_delegators :@state, :closed?, :open?, :half_open?
 
-    attr_reader :name
+    attr_reader :name, :half_open_resource_timeout
 
-    def initialize(name, exceptions:, success_threshold:, error_threshold:, error_timeout:, implementation:)
+    def initialize(name, exceptions:, success_threshold:, error_threshold:, error_timeout:, implementation:, half_open_resource_timeout: nil)
       @name = name.to_sym
       @success_count_threshold = success_threshold
       @error_count_threshold = error_threshold
       @error_timeout = error_timeout
       @exceptions = exceptions
+      @half_open_resource_timeout = half_open_resource_timeout
 
       @errors = implementation::SlidingWindow.new(max_size: @error_count_threshold)
       @successes = implementation::Integer.new
       @state = implementation::State.new
     end
 
-    def acquire
+    def acquire(resource = nil, &block)
       return yield if disabled?
 
       half_open if open? && error_timeout_expired?
@@ -27,7 +28,7 @@ module Semian
 
       result = nil
       begin
-        result = yield
+        result = maybe_with_half_open_resource_timeout(resource, &block)
       rescue *@exceptions => error
         mark_failed(error)
         raise error
@@ -121,6 +122,20 @@ module Semian
 
     def disabled?
       ENV['SEMIAN_CIRCUIT_BREAKER_DISABLED'] || ENV['SEMIAN_DISABLED']
+    end
+
+    def maybe_with_half_open_resource_timeout(resource, &block)
+      result = nil
+
+      if half_open? && @half_open_resource_timeout && resource.respond_to?(:with_resource_timeout)
+        resource.with_resource_timeout(@half_open_resource_timeout) do
+          result = block.call
+        end
+      else
+        result = block.call
+      end
+
+      result
     end
   end
 end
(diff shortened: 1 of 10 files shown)


lib/semian/circuit_breaker.rb
```diff
@@ -122,5 +123,19 @@ def log_state_transition(new_state)
     def disabled?
       ENV['SEMIAN_CIRCUIT_BREAKER_DISABLED'] || ENV['SEMIAN_DISABLED']
     end
+
+    def maybe_with_half_open_resource_timeout(resource)
```
sirupsen: This is the coupling I think is ugly; passing in the resource and figuring out whether it quacks correctly here. When I do a 2nd iteration here with comments, I'll likely refactor to pass the raw resource into the initializer of the bulkhead and circuit—which seems a bit more clean.

sirupsen:
Ugh... for some reason this didn't surface in local tests, and only fails in one place on CI, but sometimes it won't allow reconfiguring the timeouts for an established client 😖

lib/semian/mysql2.rb
```diff
@@ -83,6 +83,23 @@ def query(*args)
       end
     end
 
+    def with_resource_timeout(temp_timeout)
```
casperisfine: Should we support different timeouts for read / write, or is it assumed this timeout is simply for the recovery phase so there is no need for it to be fine grained?

lib/semian/circuit_breaker.rb
```diff
@@ -27,7 +28,7 @@ def acquire
 
       result = nil
       begin
-        result = yield
+        result = maybe_with_half_open_resource_timeout(resource) { yield }
```
casperisfine: /nitpick, pass `&block` to safe one stack frame and it's nicer.

sirupsen:
I read the MySQL2 source, and we can get the `read_timeout` without https://github.com/brianmario/mysql2/pull/955—so let's do that for now.

This should be good to go.

test/mysql2_test.rb
```diff
@@ -34,7 +34,7 @@ def test_semian_can_be_disabled
   end
 
   def test_connection_errors_open_the_circuit
-    @proxy.downstream(:latency, latency: 1200).apply do
+    @proxy.downstream(:latency, latency: 2200).apply do
```
csfrancis: Curious why you had to change this?

sirupsen: I upped the default timeout to 2s, to show that it changes to 1s during the half open state.

lib/semian/mysql2.rb
```diff
@@ -83,6 +83,20 @@ def query(*args)
...
+      prev_read_timeout = @read_timeout
+      # C-ext reads this directly, writer method will configure
+      # properly on the client but based on my read--this is good enough
+      # until we get https://github.com/brianmario/mysql2/pull/955 in
+      @read_timeout = temp_timeout
+
+      yield
+    ensure
+      @read_timeout = prev_read_timeout
+    end
```
pushrax: In general this pattern makes me a bit uncomfortable when expressed like this. If someone were to add a line of code to the top of this method that might raise, the `ensure` could set a nil value. An alternative that more clearly maps to the intention is:

```
def
  prev = current
  begin
    current = temp
    # things
  ensure
    current = prev
  end
end
```

kirs:
@sirupsen have you looked into how hard it would be to make other adapters (e.g. Redis) to support the same `with_resource_timeout` trick?

PR #189, opened 2018-08-08
circuit: half_open? returns true when about to transition

sirupsen:
The circuit maintains its state on `@state`. Before this change, the definition of `half_open?` was that `@state == :half_open`. This changes the definition of `half_open?` to that _either_ `@state == :half_open` or the error timeout window has passed—so we're effectively `half_open?`.

**Why not just get rid of `@state == :half_open`?** If the circuit stays in the `half_open` state a while, `error_timeout_expired?` is no longer possible to use, i.e. you can't remove that `open?` conditional easily, I don't think.

**Why do you need this?** I'd like to be able to add in Marginalia the state of the circuit further up the stack (since Marginalia works at the AR-level). However, with Semian patching the MySQL client (and not AR)—this information is not present that far up the stack. It makes sense to me to be able to call `Semian[resource_name].circuit_breaker.half_open?` if it's going to transition on the `#acquire` of the circuit.

### DIFF
--- a/lib/semian/simple_state.rb
+++ b/lib/semian/simple_state.rb
@@ -27,7 +27,7 @@ module Semian
         @value = :closed
       end
 
-      def half_open
+      def half_open!
         @value = :half_open
       end
 
--- a/test/net_http_test.rb
+++ b/test/net_http_test.rb
@@ -446,7 +446,7 @@ class TestNetHTTP < Minitest::Test
       uri = URI("http://#{hostname}:#{toxic_port}/200")
       http.raw_semian_options[:error_threshold].times do
         # Cause error error_threshold times so circuit opens
-        Toxiproxy[toxic_name].downstream(:latency, latency: 150).apply do
+        Toxiproxy[toxic_name].downstream(:latency, latency: 500).apply do
           request = Net::HTTP::Get.new(uri)
           assert_raises Net::ReadTimeout do
             http.request(request)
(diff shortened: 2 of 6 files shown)


test/net_http_test.rb
```diff
@@ -446,7 +446,7 @@ def open_circuit!(hostname: nil, toxic_port: nil, toxic_name: "semian_test_net_h
       uri = URI("http://#{hostname}:#{toxic_port}/200")
       http.raw_semian_options[:error_threshold].times do
         # Cause error error_threshold times so circuit opens
-        Toxiproxy[toxic_name].downstream(:latency, latency: 150).apply do
+        Toxiproxy[toxic_name].downstream(:latency, latency: 500).apply do
```
sirupsen: This was flaky. The timeout is 100ms, but I expect it's not exactly 100ms 🙃 This failed quite a lot locally.

test/protected_resource_test.rb
```diff
@@ -1,4 +1,5 @@
 require 'test_helper'
+require 'securerandom'
```
sirupsen: This works fine when run as `bundle exec rake test`, but did not work in isolation and uses `SecureRandom`.

lib/semian/simple_state.rb
```diff
@@ -27,7 +27,7 @@ def close
         @value = :closed
       end
 
-      def half_open
+      def half_open!
```
sirupsen: When there's two methods of the same name (even if they have different `?`/`!`) suffixes—I like to name the more dangerous one `!`.

PR #190, opened 2018-08-14
Make OOM an adapter error

fw42:
Follow-up to https://github.com/Shopify/semian/pull/172.

Fixes an oversight from the previous PR. This essentially makes it so that `OutOfMemoryError#semian_identifier` is a thing, which is necessary to properly integrate this type of error into Shopify core's fallback evaluation logic.

### DIFF
--- a/lib/semian/redis.rb
+++ b/lib/semian/redis.rb
@@ -13,6 +13,7 @@ class Redis
   end
 
   class OutOfMemoryError < Redis::CommandError
+    include ::Semian::AdapterError
   end
 
   ResourceBusyError = Class.new(SemianError)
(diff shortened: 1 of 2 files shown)


lib/semian/redis.rb
```diff
@@ -13,6 +13,7 @@ def initialize(semian_identifier, *args)
   end
 
   class OutOfMemoryError < Redis::CommandError
+    include ::Semian::AdapterError
```
casperisfine: I think it would be better to do:

`OutOfMemoryError.include(::Semian::AdapterError)`

This way we're sure the class is already loaded, and we don't mess things up by defining it ourselves.

casperisfine: Scratch that. I forgot this isn't an upstream exception class.

PR #192, opened 2018-09-13
reset successes when half opening

jpittis:
Seems like in https://github.com/Shopify/semian/pull/188 the calls to `log_state_transition(:half_open)` and `@successes.reset` were removed.

@sirupsen @pushrax

### DIFF
--- a/lib/semian/circuit_breaker.rb
+++ b/lib/semian/circuit_breaker.rb
@@ -2,9 +2,9 @@ module Semian
   class CircuitBreaker #:nodoc:
     extend Forwardable
 
-    def_delegators :@state, :closed?, :open?, :half_open!
+    def_delegators :@state, :closed?, :open?, :half_open?
 
-    attr_reader :name, :half_open_resource_timeout
+    attr_reader :name, :half_open_resource_timeout, :error_timeout, :state
 
     def initialize(name, exceptions:, success_threshold:, error_threshold:, error_timeout:, implementation:, half_open_resource_timeout: nil)
       @name = name.to_sym
@@ -21,7 +21,7 @@ module Semian
 
     def acquire(resource = nil, &block)
       return yield if disabled?
-      half_open! if half_open?
+      transition_to_half_open if transition_to_half_open?
 
       raise OpenCircuitError unless request_allowed?
 
@@ -37,33 +37,33 @@ module Semian
       result
     end
 
-    def half_open?
-      (open? && error_timeout_expired?) || @state.half_open?
+    def transition_to_half_open?
+      open? && error_timeout_expired? && !half_open?
     end
 
     def request_allowed?
-      closed? || half_open?
+      closed? || half_open? || transition_to_half_open?
     end
 
     def mark_failed(_error)
       push_time(@errors)
       if closed?
-        open if error_threshold_reached?
+        transition_to_open if error_threshold_reached?
       elsif half_open?
-        open
+        transition_to_open
       end
     end
 
     def mark_success
       return unless half_open?
       @successes.increment
-      close if success_threshold_reached?
+      transition_to_close if success_threshold_reached?
     end
 
     def reset
       @errors.clear
       @successes.reset
-      close
+      transition_to_close
     end
 
     def destroy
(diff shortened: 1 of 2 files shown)


sirupsen:
These lines didn't change in that PR as far as I can see? What tests are failing? I am not sure what this is doing.

jpittis:
[If you look at the current Semian master, there is code path that resets `@successes` to zero.](https://github.com/Shopify/semian/blob/master/lib/semian/circuit_breaker.rb#L91)

lib/semian/circuit_breaker.rb
```diff
@@ -87,7 +87,7 @@ def open
 
     def half_open
       log_state_transition(:half_open)
-      @state.half_open
+      half_open!
```
pushrax: IMO this delegator should just be removed, and `State#half_open!` can then be renamed back to `State#half_open`.

sirupsen:
I just spent some more time with it this morning. Wonderful find. How did you get here? 👍 

At first, I was terrified of the consequences this might have for production. However, on further thought and testing, it seems that the bug here boils down to `success_threshold` essentially _always_ be the equivalent of `1`. I was afraid this would be disastrous, rendering circuits useless, however, then the tests wouldn't have passed. In other words, since we don't have simulation for this yet, it's hard to say whether this would make an incident worse or better 🤷‍♂️ 

With that in mind, it makes sense to me no existing test failed. I wholeheartedly agree with Justin's observation that we need to clean up the delegate. I've done in my commit by distinguishing between (1) Am I in the `half_open?` state, and (2) Am I about to transition to the `half_open?` state. The former being a state, the latter being a decision the circuit makes. I think this source of confusion is what caused me to introduce the bug in the previous PR.

@jpittis hope you don't mind, but since this is a fallout for my PR I considered your PR a glorified issue of "hey I found this bug and I think this is a solution" and fixed it here and added tests. Especially because few have a working Semian testing environment (but I do, with the `Vagrantfile`).

Please review @jpittis @pushrax, then let's get this out this week!

jpittis:
We found this bug because we were re-implementing a Semian circuit breaker in Go for the semian sinulator.

As you say, I meant this PR as an issue + diff.

I didn’t raise a fuss because I came to the same conclusion as you that the bug is relatively harmless.

lib/semian/circuit_breaker.rb
```diff
@@ -37,33 +37,33 @@ def acquire(resource = nil, &block)
       result
     end
 
-    def half_open?
```
pushrax: This change reverts https://github.com/Shopify/semian/pull/189, is that what you wanted?

max611:
@sirupsen Yep will look into it! 👍

lib/semian/circuit_breaker.rb
```diff
@@ -74,20 +74,20 @@ def destroy
...
+    def transition_to_open
       log_state_transition(:open)
       @state.open
     end
 
-    def half_open
+    def transition_to_half_open
       log_state_transition(:half_open)
-      @state.half_open
+      @state.half_open!
```
pushrax: There's a slight inconsistency here now — `half_open!` gets a bang while `open` and `close` do not.

test/circuit_breaker_test.rb
```diff
@@ -98,6 +98,36 @@ def test_request_allowed_query_doesnt_trigger_transitions
     assert_predicate @resource, :open?
   end
 
+  def test_open_close_open_cycle
+    resource = Semian.register(:open_close, tickets: 1, exceptions: [SomeError], error_threshold: 2, error_timeout: 5, success_threshold: 2)
+
+    half_open_future_travel = -> { Time.now + resource.circuit_breaker.error_timeout + 1 }
```
pushrax: `Timecop.travel` already computes the absolute time from the delta in this [fancy method](https://github.com/travisjeffery/timecop/blob/71cbbd36ab28d95662ab980643bdeac96039b8b6/lib/timecop/time_stack_item.rb#L100-L126).

lib/semian/circuit_breaker.rb
```diff
@@ -74,20 +74,20 @@ def destroy
...
 
-    def open
+    def transition_to_open
       log_state_transition(:open)
       @state.open
     end
 
-    def half_open
+    def transition_to_half_open
       log_state_transition(:half_open)
```
pushrax: I don't think this line is tested.

pushrax:
There were some comments left unaddressed.

max611:
@pushrax my bad, missed those 🤦‍♂️ I'll create a new PR to address the comments

PR #193, opened 2018-09-17
Use LRU Cache for Semian resources

max611:
This PR aims to refactor the Semian resources to a type of LRU cache. This will allow us to remove unused objects from the resources array and hopefully gain some performance out of it. (Will benchmark to make sure we see an improvement)

As of now, we never clean up the resources and it can end up with thousands of resources and therefore leak some memory because of it. Instead of using a simple array, this PR aims to add a new object called `LRUCache` to dynamically remove empty resources. This will allow us to use less resources and less memory.

An example would be using many webhooks and creating a resource for everyone of them, you could easily grow the amount of resources to a large number and keep them in the resources object and never clearing them which would cause memory usage for no reason. 

One of the goals is to make sure this is backward compatible and also make sure this is an improvement. I will post benchmarks once I run them.


lib/semian.rb
```diff
@@ -157,7 +158,8 @@ def register(name, **options)
       raise ArgumentError, 'Both bulkhead and circuitbreaker cannot be disabled.'
     end
 
-    resources[name] = ProtectedResource.new(name, bulkhead, circuit_breaker)
+    resources.set(name, ProtectedResource.new(name, bulkhead, circuit_breaker))
+    resources.get(name).value
```
jpittis: Doing a set and then a get sounds like you're doing extra unnecessary book-keeping work.

lib/semian.rb
```diff
@@ -216,7 +218,7 @@ def unregister_all_resources
 
   # Retrieves a hash of all registered resources.
   def resources
-    @resources ||= {}
+    @resources ||= LRUCache.new(5)
```
jpittis: Setting a max size is good to start but we should also investigate an option that lets this grow unbounded and only pulls resources off the LRU if their error windows are empty.

lib/semian/lru_cache.rb
```diff
@@ -0,0 +1,79 @@
+class LRUCache
```
jpittis: Overall I'm finding this linked list stuff hard to follow. It doesn't matter in this prototype but I'd love it if we could find a way to make this easier to understand in the "good copy".

Whether that be through a comment, finding the right abstraction or factoring out of key methods.

test/adapter_test.rb
```diff
@@ -12,7 +12,7 @@ def teardown
   end
 
   def test_adapter_registers_consumer
-    assert_empty(Semian.resources)
+    assert_empty(Semian.resources.table)
```
jpittis: I feel like this either needs a separate test or a test re-name.

test/lru_cache_test.rb
```diff
@@ -0,0 +1,63 @@
+require 'test_helper'
+
+class TestLRUCache < Minitest::Test
+  def setup
+    @lru_cache = LRUCache.new(3)
```
jpittis: I'm pretty sure an LRU of size 3 isn't big enough to test all the branches. Maybe size 5?

test/lru_cache_test.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def test_set_item
+    @lru_cache.set('key', 'value')
+    assert @lru_cache.table['key'].value == 'value'
```
jpittis: I tend to dislike tests that assert on implementation details like this.

Don't we really just want to test that a key value that was set is returned by get?

test/semian_test.rb
```diff
@@ -61,4 +61,14 @@ def test_disabled_via_semian_wide_env_var
...
   end
+
+  def test_lru_resources
+    Semian.reset!
+    max_resources = 5
+    10.times do |i|
+      Semian.resources.set("key_#{i}", i)
+    end
+
+    assert Semian.resources.table.count <= max_resources
```
jpittis: It shouldn't be `<= max_resouces` it should be `== max_resources` no?

lib/semian/lru_cache.rb
max611: I've changed the implementation to use a simple hash instead, I used the Linked List stuff from online finding related to LRU implementation with Ruby. The hash gives us the same results and makes it way easier to read.

jacobbednarz:
@max611 👋 Would you be able to expand a little on what issue (if any?) or benefit implementing a LRU cache aims to address within Semian? We run it internally for quite a few of our services and I'm interested to hear the motivation here and whether it's something we need to be on the look out for.

Thanks!

lib/semian/lru_cache.rb
```diff
@@ -0,0 +1,50 @@
...
+  def set(key, value)
+    @table.delete(key)
+    @table[key] = value
+    @table.shift if @table.length > @max_items
+    value
+  end
+
+  def get(key)
+    found = true
+    value = @table.delete(key){ found = false }
```
jpittis: I don't see the point of this fancy `found` stuff.

`{}.delete("key") == nil`

So you can effectively do `found = @table.delete(key)` no?

test/lru_cache_test.rb
```diff
@@ -0,0 +1,63 @@
+require 'test_helper'
+
+class TestLRUCache < Minitest::Test
+  def setup
+    @lru_cache = LRUCache.new(5)
+  end
+
+  def test_set_item
```
jpittis: You now have two tests that do the same thing so maybe one test called `test_set_get_item`?

test/lru_cache_test.rb
```diff
@@ -0,0 +1,63 @@
...
+    @lru_cache.set('key', 'value')
+    assert_equal @lru_cache.get('key'), 'value'
+  end
+
+  def test_remove_item
+    @lru_cache.set('a', 1)
+    @lru_cache.set('b', 2)
+    @lru_cache.set('c', 3)
+
+    assert_equal @lru_cache.table.count, 3
```
jpittis: You can use the public `@lru_cache.count` here.

test/lru_cache_test.rb
```diff
@@ -0,0 +1,63 @@
...
+    assert_equal @lru_cache.table.values.last, @lru_cache.get('c')
+    assert_equal @lru_cache.table.values.first, @lru_cache.get('a')
+  end
+
+  def test_limit_to_max_items
+    15.times do |i|
+      @lru_cache.set("key_#{i}", i)
+    end
+
+    assert @lru_cache.table.count <= 5
```
jpittis: Isn't this a bug if `@lru_cache.count == 4` here?

lib/semian/lru_cache.rb
```diff
@@ -0,0 +1,50 @@
...
+    if found
+      @table[key] = value
```
sirupsen: Clever! Note that this won't be thread-safe unless you do a synchronize around this when `thread_safety_disabled` is not set to `true`.

lib/semian.rb
```diff
@@ -216,7 +217,7 @@ def unregister_all_resources
 
   # Retrieves a hash of all registered resources.
   def resources
-    @resources ||= {}
+    @resources ||= LRUCache.new(100)
```
sirupsen: The hardcoded limit of 100 seems problematic, but I'm sure you just put it here as a placeholder. 👍  It means that this is _not_ backward compatible. In Shopify, we have way more than 100 resources already—so this would be fairly destructive. I think we can resolve this in two ways without requiring a major version bump for other consumers of Semian outside of Shopify:

* Only destroy resources that have no used bulkheads and no errors in the window
* Provide a configurable limit

(2) is going to mean we'll need to manually count this in every app, which will age poorly. (1) seems hard, but I think might be worth it 🤔 WDYT?

lib/semian/lru_cache.rb
```diff
@@ -0,0 +1,50 @@
...
+    @table.delete(key)
+  end
+
+  def clear
+    @table.clear
+  end
+
+  def count
+    @table.size
+  end
```
sirupsen: Using `Forwardable` for these methods may be simpler.

max611:
Hey @jacobbednarz, I've updated the PR description to give more details! The goal is mainly to avoid creating tons of resources without ever destroying them which could cause memory problems. We'll try to make sure this is not a breaking change.

lib/semian.rb
max611: I've updated the PR and I think for now we could simply delete resources that are not using bulkheads and the errors are empty. Once we see if it works, we could add a new config variable that could limit the number of resources possible at any given time and if we reach that limit, we delete the oldest one. Adding a limit might not be that useful, I'll do some benchmarking.

jacobbednarz:
🙇 much appreciated @max611! looking forward to seeing it land 😃

lib/semian/lru_cache.rb
```diff
@@ -0,0 +1,51 @@
...
+  def delete(key)
+    @table.delete(key)
+  end
+
+  private
+
+  def clear_unused_resources
+    @table.each do |key, value|
+      next unless value.instance_of?(Semian::ProtectedResource)
+      @table.delete(key) unless value&.circuit_breaker.in_use? && !!value&.bulkhead
```
sirupsen: This in `in_use?` pattern is perfect. Makes a ton of sense to me. What may be a tad bit surprising to some users is that circuit breaker resources are cleared, but those with bulkheads are _never_ cleared. I think it's reasonable to do, however, clearing a bulkhead should not be a big deal. That said, it's somewhat harder to reason about a newly created bulkhead that's "warming up" as the workers register, than an established one—so I think it makes sense to leave these out for now.

I think you should delegate that decision down to the `bulkhead` though. You can define a method in `ext/semian/semian.c` like so:

```
VALUE
semian_resource_in_use(VALUE self)
{
  return Qtrue;
}
```

And then attach it to the object at the bottom:

```
rb_define_method(cResource, "in_use?", semian_resource_in_use, 0);
```

However, it's not unreasonable in that function to call `semian_resource_tickets` and if any tickets are claimed, returning `false`. As mentioned earlier, the amount of churn would thus be concerning. `in_use?` could be `false` if _it's not currently claimed_ and _it hasn't been claimed in the past 60 seconds_. The latter information can be found through the [`semctl(2)`](http://man7.org/linux/man-pages/man2/semctl.2.html) system call, that's already used in the C extension.

Additionally, it would make sense to start LRUing the bulkheads if you reach the ceiling on the number of semaphores allowed on the system, a number you can obtain with the `IPC_INFO` call to `semctl(2)`, see again the [man page](http://man7.org/linux/man-pages/man2/semctl.2.html). Being close to this ceiling at claim time, could invoke this method.

To boil this down: You can do much more here, but, you're hitting diminishing very quickly. I'd suggest just putting `in_use?` to return `Qtrue` in the C extension for the time being.

sirupsen: I think it would be good custom here to call `destroy` on all the resource (the circuit + bulkhead, even if the latter doesn't exist) before deleting it from the hash. I don't expect that to make any difference now, but maybe it would in the future if the Circuit starts doing things like modifying timeouts permanently and needs to be cleaned up 🤷‍♂️

lib/semian/lru_cache.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def delete(key)
+    @table.delete(key)
+  end
+
+  private
+
+  def clear_unused_resources
+    @table.each do |key, value|
+      next unless value.instance_of?(Semian::ProtectedResource)
```
sirupsen: When would it ever _not_ be a protected resource?

test/lru_cache_test.rb
```diff
@@ -0,0 +1,89 @@
+require 'test_helper'
+
+class TestLRUCache < Minitest::Test
```
sirupsen: Several of the tests aren't passing for me?

lib/semian/lru_cache.rb
```diff
@@ -0,0 +1,51 @@
+class LRUCache
+  extend Forwardable
+  def_delegators :@table, :size, :count
+  def_delegators :@table, :empty?
+  def_delegators :@table, :keys
+  def_delegators :@table, :clear
+
+  attr_accessor :max_items, :table
+  def initialize(max_items)
+    @max_items = max_items
```
sirupsen: Great, let's 🔥 `max_items` 👏

lib/semian/protected_resource.rb
```diff
@@ -27,6 +27,10 @@ def acquire(timeout: nil, scope: nil, adapter: nil, resource: nil)
       end
     end
 
+    def in_use?
+      true
+    end
```
sirupsen: Why do you need this?

test/lru_cache_test.rb
```diff
@@ -0,0 +1,89 @@
...
+  def setup
+    @lru_cache = LRUCache.new(5)
+  end
```
sirupsen: You need a test with a bulkhead too, to ensure those don't get deleted.

lib/semian/lru_cache.rb
```diff
@@ -0,0 +1,51 @@
...
+    @table = {}
+  end
+
+  def set(key, value)
+    @table.delete(key)
+    @table[key] = value
+    clear_unused_resources
```
sirupsen: I think this is a good place to do it. The `set` is not in a hot path, and this is simple.

lib/semian/lru_cache.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def delete(key)
+    @table.delete(key)
+  end
+
+  private
+
+  def clear_unused_resources
+    @table.each do |key, value|
```
sirupsen: I am a little concerned about the performance of this. I don't know how big this set gets, but likely this is 100+ elements in production?

lib/semian/lru_cache.rb
max611: I think I added this line when I was running some tests and forgot to remove it  🤦‍♂️

lib/semian/lru_cache.rb
max611: I synced with @jpittis last week and ran some benchmarks on Friday. Going over the complete hash is indeed very slow and could cause problems if we have many many resources.  We were thinking about maybe simply removing 2 resources when running this method and simply start removing them if we are above our max items number (I don't think removing 2 elements make things very slow so we could also remove the `max_items` completely. Something like:
```
def clear_unused_resources
    return if @table.length < @max_items
    cleaned_elements = 0
    @table.each do |key, value|
      break if cleaned_elements >= 2
      unless value&.circuit_breaker.in_use? && !!value&.bulkhead
        cleaned_elements += 1
        delete(key)
      end
    end
  end
```

Let me know what you think

lib/semian/lru_cache.rb
max611: Alright so the problem is with this test: https://github.com/Shopify/semian/blob/master/test/resource_test.rb#L80

I was under the impression it would be possible to use a `Resource` Do you think I could get a better fix than verifying the instance?

test/lru_cache_test.rb
max611: Done!

test/lru_cache_test.rb
max611: Should be fixed now! 🎉

lib/semian.rb
```diff
@@ -13,6 +13,7 @@
 require 'semian/simple_sliding_window'
 require 'semian/simple_integer'
 require 'semian/simple_state'
+require 'semian/lru_cache'
```
sirupsen: What does the "cache" refer to here? Isn't it just an LRU? Or a LRU hash?

lib/semian/lru_cache.rb
```diff
@@ -0,0 +1,54 @@
...
+  private
+
+  def clear_unused_resources
+    cleaned_elements = 0
+    @table.each do |key, value|
+      break if cleaned_elements >= 2
+      next if value.instance_of?(Semian::Resource)
+      next if value.circuit_breaker&.in_use?
+      next if value.created_at + minimum_time_in_queue > Time.now
+      next if value.bulkhead
```
sirupsen: You added `in_use?` on the Bulkhead, I'd expect you to be using it here :)

lib/semian/lru_cache.rb
```diff
@@ -0,0 +1,54 @@
...
+    @table.delete(key)
+  end
+
+  private
+
+  def clear_unused_resources
+    cleaned_elements = 0
+    @table.each do |key, value|
+      break if cleaned_elements >= 2
+      next if value.instance_of?(Semian::Resource)
```
sirupsen: For some reason, Github is not letting me comment where you addressed this. I don't see why this would ever be true. If a test is violating this, let's make sure that test is using a protected resource IMO.

lib/semian/lru_cache.rb
```diff
@@ -0,0 +1,54 @@
+class LRUCache
+  extend Forwardable
+  def_delegators :@table, :size, :count
+  def_delegators :@table, :empty?
+  def_delegators :@table, :keys
+  def_delegators :@table, :clear
+
+  attr_accessor :table, :minimum_time_in_queue
+  def initialize(minimum_time_in_queue = 5)
+    @minimum_time_in_queue = minimum_time_in_queue * 60
```
sirupsen: It's not really a queue, `minimum_time_in_lru` perhaps sound more accurate?

lib/semian/lru_cache.rb
```diff
@@ -0,0 +1,54 @@
...
+  def delete(key)
+    @table.delete(key)
+  end
+
+  private
+
+  def clear_unused_resources
+    cleaned_elements = 0
+    @table.each do |key, value|
+      break if cleaned_elements >= 2
```
sirupsen: Not completely sure how it'd fare in the real world, your worst case here is still to traverse _a lot_ of resources 🤷‍♂️  What kinds of benchmarks have you run? Again, because this is only run on `set`, it doesn't seem too bad, I'm just not sure I understand what logic lead you to this decision.

lib/semian/lru_cache.rb
```diff
@@ -0,0 +1,54 @@
...
+      next if value.instance_of?(Semian::Resource)
+      next if value.circuit_breaker&.in_use?
+      next if value.created_at + minimum_time_in_queue > Time.now
```
sirupsen: I'm not sure I understand what this heuristic does. For the first 5 minutes, it won't remove it even if it's empty? Why does that make a difference for an empty circuit?

lib/semian/lru_cache.rb
sirupsen: Also, no test fails in the test suite for this LRU if I remove this line.

lib/semian/lru_cache.rb
sirupsen: No test fails if I remove it or change the number though.

lib/semian/lru_cache.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def []=(key, value)
+    set(key, value)
+  end
+
+  def [](key)
+    get(key)
+  end
+
+  def delete(key)
```
sirupsen: Any reason this is not a delegate?

lib/semian/lru_cache.rb
```diff
@@ -0,0 +1,54 @@
+class LRUCache
+  extend Forwardable
+  def_delegators :@table, :size, :count
+  def_delegators :@table, :empty?
+  def_delegators :@table, :keys
+  def_delegators :@table, :clear
```
sirupsen: You can do all of this in one line, like in `ProtectedResource`—I think that's a bit cleaner.

lib/semian/lru_cache.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  attr_accessor :table, :minimum_time_in_queue
```
sirupsen: Do you really want people to change these two at runtime? They seem like they should be `attr_reader`, no?

lib/semian/lru_cache.rb
```diff
@@ -0,0 +1,54 @@
...
+  def initialize(minimum_time_in_queue = 5)
```
sirupsen: If you want this in seconds, you should pass it in seconds IMO.

lib/semian/lru_cache.rb
max611: The problem I was getting is, for example, this test: https://github.com/Shopify/semian/blob/master/test/protected_resource_test.rb#L96

We register the resource, but at this point, there is no error in the window, which will cause the `clean_unused_resource` method to remove it as soon as it gets there. I feel like I might be missing something or maybe this should not happen.

max611:
I've run some benchmarks, let me know if the code makes sense. 

I changed the `minimum_time_in_lru` to 0 in both the cases.
LRU (Clearing only 2 items at once)
```
500 resources       0.655  (± 0.0%) i/s -      4.000  in   6.523365s
1000 resources       0.389  (± 0.0%) i/s -      2.000  in   5.138045s
2500 resources       0.149  (± 0.0%) i/s -      1.000  in   6.726565s
5000 resources       0.055  (± 0.0%) i/s -      1.000  in  18.310367s
10000 resources       0.023  (± 0.0%) i/s -      1.000  in  44.148743s
```

Average time to process one request
```
With 10 000 resources: 0.004628790610585202
With 20 000 resources: 0.004784476699998049
With 50 000 resources: 0.005969256284977528
```

LRU (Clearing all unused resources)
```
500 resources       0.572  (± 0.0%) i/s -      3.000  in   5.518382s
1000 resources       0.343  (± 0.0%) i/s -      2.000  in   5.865382s
2500 resources       0.112  (± 0.0%) i/s -      1.000  in   8.945068s
5000 resources       0.048  (± 0.0%) i/s -      1.000  in  20.826114s
10000 resources       0.020  (± 0.0%) i/s -      1.000  in  49.520911s
```

Average time to process one request
```
With 10 000 resources: 0.0043376523082934
With 20 000 resources: 0.005373792125002546
With 50 000 resources: 0.005837922908061519
```


Hash
```
500 resources       0.691  (± 0.0%) i/s -      4.000  in   5.793268s
1000 resources       0.403  (± 0.0%) i/s -      3.000  in   7.446834s
2500 resources       0.113  (± 0.0%) i/s -      1.000  in   8.888770s
5000 resources       0.046  (± 0.0%) i/s -      1.000  in  21.661930s
10000 resources       0.030  (± 0.0%) i/s -      1.000  in  33.491730s
```

Average time to process one request
```
With 10 000 resources: 0.0034239859479996086
With 20 000 resources: 0.0033998030318058226
With 50 000 resources: 0.003728271214527558
```

It seems like the difference between removing all the unused resources or only 2 does not change much on performance.

lib/benchmarker.rb
```diff
@@ -0,0 +1,218 @@
...
+        end
+      end
+    end
+    # puts "Average = #{@request_time.reduce(:+) / @request_time.size.to_f}"
+  end
+
+  def create_request(host, port, i)
+    random = rand(1...100)
+    average_time = average do
+      if random >= 0 && random <= 50
```
max611: This allows us to have a `success` or `failure` request, this make sure the data is not always a `success` or a `failure`

lib/benchmarker.rb
```diff
@@ -0,0 +1,218 @@
...
+        number_of_resources.times do |i|
+          create_request(host, port, i)
+        end
+      end
+    end
+    # puts "Average = #{@request_time.reduce(:+) / @request_time.size.to_f}"
+  end
+
+  def create_request(host, port, i)
+    random = rand(1...100)
```
max611: I'm currently trying with a fixed seed to make sure we are getting the same results in all the experiments but I'm currently not sure if it works well. I will dig into it more.
```
random_with_seed = Random.new(1)
random = random_with_seed.rand(100)
```

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,53 @@
...
+  def [](key)
+    get(key)
+  end
+
+  private
+
+  def clear_unused_resources
+    cleaned_elements = 0
+    @table.each do |key, value|
+      break if cleaned_elements >= 2
```
max611: From the benchmarks, we can see that it does not bring much value to remove only 2 resources per set. I think we should be fine with removing all unused resources and not limiting them.

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,53 @@
+class LRUHash
+  extend Forwardable
+  def_delegators :@table, :size, :count, :empty?, :keys, :clear, :delete
+
+  attr_reader :table, :minimum_time_in_lru
+  def initialize(minimum_time_in_lru = 300)
+    @minimum_time_in_lru = minimum_time_in_lru
```
max611: The problem I was getting and that got fixed by adding that, is this test: https://github.com/Shopify/semian/blob/master/test/protected_resource_test.rb#L96

We register the resource, but at this point, there is no error in the window, which will cause the clean_unused_resource method to remove it as soon as it gets there. I feel like I might be missing something or maybe this should not happen.

lib/benchmarker.rb
```diff
@@ -0,0 +1,218 @@
...
+    average_time = average do
+      if random >= 0 && random <= 50
+        Net::HTTP.get(URI("http://#{SemianConfig['toxiproxy_upstream_host']}:#{3000}/"))
```
max611: To make sure i'm creating a bunch of resources, I have to monkey patch the `semian_identifier` to add a random number at the end so we always create a new resource

lib/semian/lru_hash.rb
sirupsen: Yeah, thinking more about it I think it's reasonable to have this. I think the heuristic here really is that it only really makes sense to have a circuit if you have an error.

lib/semian/lru_hash.rb
sirupsen: Note that your benchmarks are going to be fairly contrived. This may actually work well in the real world, but I have no idea, and no proof—in the absence of that, I think you should remove this complexity because it's counter-intuitive. Counterintuitive behaviour is OK in the presence of a proof, but otherwise, 🔥

lib/benchmarker.rb
```diff
@@ -0,0 +1,218 @@
+$LOAD_PATH.unshift File.expand_path('../', __FILE__)
```
sirupsen: This testing is more elaborate than I anticipated. Why are you going through the HTTP stack here with Toxiproxy? You're using a bunch of test helpers, if you want to use them, instead of duplicating the code can you create helpers that you can include here if you want to continue to use the `Net::HTTP` stack?

Also, we should move this to `/test/benchmark` directory instead—`lib/` is going to be included with the gem, which we definitely don't want :)

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,53 @@
...
+    set(key, value)
+  end
+
+  def [](key)
+    get(key)
+  end
+
+  private
+
+  def clear_unused_resources
```
sirupsen: For performance, I'd suggest you don't claim this within the synchronize, but instead claim the mutex around `#delete` yourself here.

test/lru_hash_test.rb
```diff
@@ -0,0 +1,100 @@
+require 'test_helper'
```
sirupsen: `minitest/hell` I believe it's called is a good way to test this file for thread-sfaety :)

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,53 @@
+class LRUHash
+  extend Forwardable
+  def_delegators :@table, :size, :count, :empty?, :keys, :clear, :delete
+
+  attr_reader :table, :minimum_time_in_lru
+  def initialize(minimum_time_in_lru = 300)
+    @minimum_time_in_lru = minimum_time_in_lru
+    @table = {}
+    @lock = Mutex.new
```
sirupsen: You should pass the `thread_safety_disabled` and replace `Mutex` with a mock here, e.g. `class NoopMutex; def synchronize(*); end; end` when we don't have thread-safety enabled. A mutex is a substantial performance hit, so let's not use it unless necessary.

lib/benchmarker.rb
max611: I mainly wanted to replicate a real-life environment but I might have gone too far and that's most likely not needed. I'll move everything into a helper! 👍

max611:
@sirupsen I will investigate on Core but I was seeing mainly an average of 100 resources but I will dig deeper. In my comment https://github.com/Shopify/semian/pull/193#issuecomment-424492158 The average time is for a `set` action. So before the LRU it takes an average of `0.003954889166600413` seconds to do one `set` and with the LRU it takes `0.005914116217800153` seconds. (This is over 5000 `set`) I might be missing something and not replicate a real life application. Let me know what you think.

lib/semian.rb
```diff
@@ -150,6 +151,7 @@ def to_s
   #
   # Returns the registered resource.
   def register(name, **options)
+    thread_safe = options[:thread_safety_disabled]
```
max611: I this legit? I feel like that might be a bit hacky

sirupsen: For backward compatibility, you'll need to support both, unfortunately. However, you can emit to `STDERR` that this option per resource will be removed in favour of a global one.

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,61 @@
+class LRUHash
+  extend Forwardable
+  def_delegators :@table, :size, :count, :empty?, :keys, :clear, :delete
+  attr_reader :table, :minimum_time_in_lru
+
+  class NoopMutex
+    def synchronize(*)
```
sirupsen: You're going to need to yield here, which leads me to believe you should probably make sure this code-path is hit :)

lib/semian/resource.rb
```diff
@@ -5,7 +5,7 @@ class Resource #:nodoc:
     class << Semian::Resource
       # Ensure that there can only be one resource of a given type
       def instance(*args)
-        Semian.resources[args.first] ||= new(*args)
+        Semian.resources[args.first] ||= ProtectedResource.new(args.first, new(*args), nil)
```
max611: https://github.com/Shopify/semian/blob/master/test/resource_test.rb#L80 was failing since we were using a `Resource` with `Semain.resources`. The error was raising [here](https://github.com/Shopify/semian/pull/193/files#diff-0c0b1885ac3076e71428d7414e517530R51) since `circuit_breaker` does not exist with `Resource`. Let me know if this fix make sense.

lib/semian.rb
```diff
@@ -238,14 +237,18 @@ def thread_safe?
...
   end
 
   def create_circuit_breaker(name, **options)
     circuit_breaker = options.fetch(:circuit_breaker, true)
     return unless circuit_breaker
     require_keys!([:success_threshold, :error_threshold, :error_timeout], options)
+    unless options[:thread_safety_disabled].nil?
+      logger.info("NOTE: thread_safety_disabled per resource will be replaced by a global setting")
+    end
+    update_thread_safe(options[:thread_safety_disabled].nil? ? thread_safe? : !options[:thread_safety_disabled])
```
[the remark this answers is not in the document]
sirupsen: It seems quite confusing to me that configuring a resource for thread-safety will toggle a global setting. I think this violates the "principle of least surprise." The way I imagined this, is that here, you'd have `thread_safe = Semian.thread_safe? || !thread_safety_diabled`—with no side-effect.

Instead, I think you should introduce the global setting in this release and when you enable this for our single-threaded apps (e.g. Shopify)—you enable the global setting as well when you bump the gem.

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,63 @@
...
+      yield
+    end
+  end
+
+  def initialize(minimum_time_in_lru = 300)
```
sirupsen: It's fairly counterintuitive that the `minimum_time_in_lru` is the default only option to an initializer. That's going to be painful if we plan on maintaining backward compatibility in future releases. Can you make this a named argument? When I looked at this initially, I thought you'd re-introduced size.

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,63 @@
+class LRUHash
```
sirupsen: From my naive PoV, this looks thread-safe.

lib/semian/resource.rb
```diff
@@ -5,7 +5,7 @@ class Resource #:nodoc:
...
       end
     end
```
sirupsen: This seems to mock the circuits when there's no C implementation? In this case, you'll need to mock out `in_use?` here as well.

lib/semian/resource.rb
sirupsen: Yeah, this seems correct. I think that "Resource" and "Bulkhead" are somewhat convoluted here. A useful refactor would be to change `Semian::Resource` to `Semian::Bulkhead` instead. I think it's sensible to change a resource to always be a "protected resource"; with it having a bulkhead and circuit object within.

lib/semian.rb
```diff
@@ -216,7 +217,7 @@ def unregister_all_resources
 
   # Retrieves a hash of all registered resources.
   def resources
-    @resources ||= {}
+    @resources ||= LRUHash.new(minimum_time_in_lru: 300)
```
jpittis: If we're gonna hardcode 300, I'd rather we don't hardcode it twice.
A constant `MINIMUM_TIME_IN_LRU` or a `new_lru_hash` method would be my preference.

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def [](key)
+    get(key)
+  end
+
+  private
+
+  def clear_unused_resources
+    @table.each do |key, value|
```
jpittis: We're doing an O(n) operation here which means there's absolutely no value in this being an LRU to begin with.

Maybe we should parameterize how many elements of the table are looked at.

As we discussed over lunch of a few days ago: `clear_unused_resources(2)` should do the trick.

Another optimization I'm pretty sure we can do here (though it may not be necessary) is that once we skip a resource because it's in use, as can assume the future items in the queue will also be in use. (Assuming you implement `updated_at` rather than `created_at`.)

lib/semian/resource.rb
jpittis: I want to emphasize that I find this quite confusing.
This should either be made less confusing or a comment should be added IMO.

lib/semian/lru_hash.rb
jpittis: I must say I disagree here. I don't have proof but I'd expect that iterating over `1000` items hash is slower than iterating over a `2` items. You might say that the max size of resources in core is `100` so it's really a comparison between `2` and `100` but I expect the resources count to increase when we use this for things like webhooks.

Assuming we go with the `updated_at` approach rather than `created_at`, I think this can be solved via `@table.take(2).each do |key, value|` which is hardly much more complexity.

lib/semian/resource.rb
max611: I would prefer to make it less confusing without having to add a comment. Do you have any idea how I could do that? I honestly don't really know how to make it simpler without modifying any behaviour.

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,63 @@
+class LRUHash
+  extend Forwardable
+  def_delegators :@table, :size, :count, :empty?, :keys, :clear, :delete
```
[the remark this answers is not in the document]
max611: I'll look into that 👀 Do you think it would be safe to simply lock the resource on those calls?

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def [](key)
+    get(key)
+  end
+
+  private
+
+  def clear_unused_resources
+    @table.take(2).each do |key, resource|
```
max611: After discussing with @jpittis, I've realized that if we have 100 resources and all of them are in use, we will still go through all the resources. Since the `get` puts the value at the back of the hash, it's safe to assume that the un-used values are at the front and will get cleared first. The reason I break on `resource.updated_at + minimum_time_in_lru > Time.now` is if we hit that minimum time, it means all the other resources are also going to be true, therefore, we do not need to keep going through the hash. Let me know if that makes sense, I will run some benchmark with and without `take(2)` to see the difference.

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,75 @@
...
+        Mutex.new
+      else
+        NoopMutex.new
+      end
+  end
+
+  def set(key, resource)
+    delete(key)
+    @lock.synchronize do
+      @table[key] = resource
```
max611: Now that `delete` is thread safe, I'm trying to figure out if we still need to synchronize on this.

jpittis:
I want to make my argument regarding cache eviction clear to both @sirupsen and @max611 (and to myself) so I'm going to write it out in whole bellow:

````ruby
def clear_unused_resources
  @table.each do |key, value|
    next if value.circuit_breaker&.in_use? || value.bulkhead&.in_use?
    next if value.created_at + minimum_time_in_lru > Time.now

    @lock.synchronize do
      if resource = delete(key)
        resource.destroy
      end
    end
  end
end
`````

In 989f5fa992a4252a077db39a8d817054dbb0cd4f, `clear_unused_resources` is being called on every `set` operation.

Notice that it does an O(n) iteration over the `@tables` hash. Notice that nowhere in the block does it break from the loop. This implementation means that `LRUHash` has the same behaviour as a standard Ruby `Hash` but with a performance overhead. If we are going to keep this implementation, we might as well delete all the `LRUHash` code and do an O(n) iteration over a plain Ruby `Hash`.

## Using updated_at rather than created_at.

````ruby
next if value.created_at + minimum_time_in_lru > Time.now
````

I think rather than the above line, we should replace it with:

````ruby
next if value.updated_at + minimum_time_in_lru_since_update > Time.now
````

Obviously we don't want to evict resources that have been recently created. But I would argue we also don't want to evict resources that have been recently used but were created a long time ago.

We can kill both birds with one stone by just keeping track of the `value.updated_at` and initializing `updated_at` to the creation time.

## Knowing when to stop iterating.

It would be nice to stop iterating over the LRU once we know there are no more resources to cleanup. Let's see if we can find the predicate required to make this decision.

Imagine two resources in the LRU with one to the left and one to the right. If we can prove a property for these two items, we know from induction that it holds for the rest:

- If the first resource has an error in it's window, could the resource to it's right not have an error in it's window? Yes. If the resource to it's right's window size is smaller.

- If the first resource has an error in it's window, could the resource to it's right be past it's `minimum_time_in_lru_since_update`? We know that resource to the right will always have larger `updated_at` time but we don't know how much larger.

Conclusion: We can't infer properties about the right resource based on the left resource being `in_use?`.

- If the first resource is younger than `minimum_time_in_lru_since_update`, we know that the resource to it's right is also younger than `minimum_time_in_lru_since_update`.

- If the first resource younger than `minimum_time_in_lru_since_update`, the resource to the right could could either be in use or not.

Conclusion: If the first resource is `value.updated_at + minimum_time_in_lru_since_update > Time.now`, we know the next one will be too.

## Only removing 2 resources.

Another option to stop the O(n) operation is to limit the number of resources we remove at a time.
If we remove 2 resources every time to create 1 resource, that will ensure that we never leak unused resources.

I originally thought we could do this via:

````ruby
@table.take(2).each do |key, value|
````

But I no longer think this is possible. If the first two elements are in use, it could be the third element is not in use and we want to delete that one. `take(2)` does not work like this.

Maybe something like

````ruby
@table.filter_by_predicate.take(2)
````

might work.

**Comment by jpittis** [comment]
The other thing I'm concerned with now is that `clear_unused_resources` is not actually thread safe.

I don't think it's possible to iterate over a hash without locking and then lock every time you find an item you want to delete. This means it's possible that another thread will be iterating over the hash at the same time and the delete will collide with the other thread's iteration.

**Comment by sirupsen** [comment]
@jpittis yes, you nailed something I haven't been able to articulate. I agree, we should move to `updated_at`.

**Comment by max611** [comment]
@jpittis and @sirupsen, I've updated the PR to use `updated_at` instead of `created_at` and added more locking into `lru_hash.rb`. (Let me know if I missed anything)

I've also refactored the benchmark file to use simple objects instead of creating new net calls, which simplifies everything. Here's the result:

LRU Hash

Average
```
Average for 500 resources = 9.676011999363255e-05
Average for 1000 resources = 0.00017510321300642318
Average for 2500 resources = 0.000384409976005918
Average for 5000 resources = 0.0007303653397967537
```

Total:
```
500 resources      21.565  (± 4.6%) i/s -    108.000  in   5.027448s
1000 resources       6.761  (± 0.0%) i/s -     34.000  in   5.043767s
2500 resources       1.281  (± 0.0%) i/s -      7.000  in   5.473631s
5000 resources       0.313  (± 0.0%) i/s -      2.000  in   6.394637s
```

### Old way
Average
```
Average for 500 resources = 4.700791600271259e-05
Average for 1000 resources = 4.916882000134137e-05
Average for 2500 resources = 7.797212199675413e-05
Average for 5000 resources = 0.00011448747660165281
```

Total
```
500 resources      38.831  (±12.9%) i/s -    192.000  in   5.026113s
1000 resources      14.700  (±13.6%) i/s -     73.000  in   5.052168s
2500 resources       4.642  (± 0.0%) i/s -     24.000  in   5.198669s
5000 resources       1.568  (± 0.0%) i/s -      8.000  in   5.109207s
```

We do see a small difference starting at 2500 and a bigger difference at 5000 resources, but this will allow us to reduce memory. Here's a benchmark for 1000 resources memory wise:

LRU Hash
```
19.873M memsize (   599.608k retained)
302.847k objects (     7.935k retained)
50.000  strings (    50.000  retained)
```

Old way
```
7.204M memsize (     1.067M retained)
50.112k objects (    14.521k retained)
50.000  strings (    50.000  retained)
```

In terms of retained memory, we save around 50% since my benchmark is doing 50/50 in use / unsused resources.

File: lib/semian.rb [hunk 54]
```diff
@@ -226,7 +228,16 @@ def consumers
...
-    @resources = {}
+    @resources = LRUHash.new(minimum_time_in_lru: MINIMUM_TIME_IN_LRU)
+  end
+
+  def thread_safe?
+    return @thread_safe if defined?(@thread_safe)
+    @thread_safe = true
+  end
+
+  def define_thread_safe(thread_safe)
```
max611: make this thread_safe=

lib/semian.rb
```diff
@@ -245,10 +255,19 @@ def create_circuit_breaker(name, **options)
...
       exceptions: Array(exceptions) + [::Semian::BaseError],
       half_open_resource_timeout: options[:half_open_resource_timeout],
-      implementation: implementation,
+      implementation: implementation(**options),
     )
   end
 
+  def implementation(**options)
+    unless options[:thread_safety_disabled].nil?
+      logger.info("NOTE: thread_safety_disabled per resource will be replaced by a global setting")
```
max611: longer

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,80 @@
...
+  end
+
+  def [](key)
+    get(key)
+  end
+
+  private
+
+  def clear_unused_resources
+    @lock.synchronize do
```
max611: return if locked.

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,80 @@
...
+      unused_resources.take(2).each do |resource|
+        break if resource.updated_at + minimum_time_in_lru > Time.now
```
max611: break instead of take to avoid 2 x O(n)

test/benchmark/benchmarker.rb
```diff
@@ -0,0 +1,103 @@
+$LOAD_PATH.unshift File.expand_path('../../../lib', __FILE__)
```
max611: better name

lib/semian.rb
```diff
@@ -88,6 +89,7 @@ module Semian
   TimeoutError = Class.new(BaseError)
   InternalError = Class.new(BaseError)
   OpenCircuitError = Class.new(BaseError)
+  MINIMUM_TIME_IN_LRU = 300
```
max611: concern of lruhash, not parent

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,80 @@
+class LRUHash
```
max611: comment about what you're trying to do

lib/semian/lru_hash.rb
max611: add comment here about what it's doing

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  def [](key)
+    get(key)
+  end
+
+  private
+
+  def clear_unused_resources
+    # Clears resources that have not been used in the last 5 minutes.
+    Semian.notify(:lru_hash_cleaned, self, :cleaning, :lru_hash)
```
sirupsen: This is technically not backwards compatible. The second argument is typically a resource where you can call e.g. `bulkhead` on. You're hacking around this in your tests, which is an indication this won't work for other consumers. Does not seem easy to change, but I am personally OK with allowing this specific backwards incompatibility since it should uncover in a test suite easily.

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,91 @@
...
+    return if @lock.locked?
```
jpittis: There's a race here. There's a chance it's not locked, on line 78, then it locks and then it reaches line 79.

To do this check atomically, you can use `try_lock`.

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,91 @@
...
+    @table = {}
+    @lock =
+      if Semian.thread_safe?
+        Mutex.new
+      else
+        NoopMutex.new
+      end
+  end
+
+  def set(key, resource)
```
jpittis: I'm not convinced this moving of the mutex is correct. The delete + set are no longer atomic.

Can you explain how you're convinced this is correct?

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,91 @@
...
+    delete(key)
+    @lock.synchronize do
+      @table[key] = resource
+    end
+    clear_unused_resources
+    resource.updated_at = Time.now
+  end
+
+  def get(key)
+    found = delete(key)
```
jpittis: Some as above. Can you explain your reasoning behind this lock placement being correct?

max611: I've added a lock on the `delete` method since it's a public method and I want to avoid race condition there, so I did not want to lock inside the lock but now that I think about it. I can fix that by simply calling `@table.delete(key)` here, I'll get that fixed!

sirupsen:
@max611 what's blocking completion here? What can @jpittis and I do to help?

max611:
@sirupsen I think we're good. Could you give the PR another look @jpittis ? Let me know if I missed anything.

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,94 @@
+class LRUHash
+  # This LRU (Least Recently Used) hash will allow
+  # the cleaning of resources as time goes.
+  # The  goal is to remove the least recently used resources
```
jpittis: Two nitpicks here. "time goes on." and a double space after "The".

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,94 @@
...
+    @table.each do |_, resource|
+      break if resource.updated_at + minimum_time_in_lru > Time.now
+      next if resource.in_use?
+
+      resource = @table.delete(resource.name)
+      if resource
+        resource.destroy
+      end
+    end
+    @lock.unlock
```
jpittis: I think this needs to be in an `ensure` block to avoid accidental lock poisoning.

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  def clear_unused_resources
+    # Clears resources that have not been used in the last 5 minutes.
+    Semian.notify(:lru_hash_cleaned, self, :cleaning, :lru_hash)
+    return unless @lock.try_lock
+    @table.each do |_, resource|
+      break if resource.updated_at + minimum_time_in_lru > Time.now
+      next if resource.in_use?
+
+      resource = @table.delete(resource.name)
```
jpittis: What are the semantics of deleting an item in a Hash while iterating over it?
Let's just be double sure this is safe.

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,94 @@
...
+        NoopMutex.new
+      end
+  end
+
+  def set(key, resource)
+    @lock.synchronize do
+      @table.delete(key)
+      @table[key] = resource
+    end
+    resource.updated_at = Time.now
```
jpittis: I can't think of why setting `resource.updated_at` outside the mutex is unsafe. That being said, this feels weird because we're writing to a shared resource without a lock. Feel free to mark this comment as resolved without changing anything.

lib/semian/lru_hash.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def get(key)
+    @lock.synchronize do
+      found = @table.delete(key)
+      if found
+        @table[key] = found
+      end
+    end
+    @table[key]
```
jpittis: I think you can get rid of the line

````ruby
@table[key]
````

and just return `found` from inside the `@lock.synchronize` block. No?

File: test/lru_hash_test.rb [hunk 70]
```diff
@@ -2,6 +2,7 @@
 
 class TestLRUHash < Minitest::Test
   def setup
+    Semian.thread_safe = true
```
jpittis: This state is going to leak. You could cache the value in `@old_thread_safe` and set it in a `def teardown`. This is a nitpick.

lib/semian/lru_hash.rb
jpittis: Unfortunately I think this is going to raise an exception.

> Raises ThreadError if mutex wasn't locked by the current thread.

I think an `if @lock.locked?` will do the trick.

lib/semian/lru_hash.rb
```diff
@@ -86,6 +94,8 @@ def clear_unused_resources
           resource.destroy
         end
       end
+    ensure
+      @lock.unlock if @lock.locked?
```
jpittis: Do we need this if statement if the ensure only happens after `try_lock`?

max611: Oops you're right, we won't get into the `begin` block unless the lock is locked.

lib/semian/circuit_breaker.rb
```diff
@@ -74,7 +74,8 @@ def destroy
     end
 
     def in_use?
-      @errors.size > 0
+      return false if @errors.size <= 0
```
jpittis: `@error.size` is never going to be `<` than 0.

lib/semian/resource.rb
```diff
@@ -51,5 +51,9 @@ def semid
     def key
       '0x00000000'
     end
+
+    def in_use?
```
max611: As discussed with @sirupsen, if `SEMIAN_SEMAPHORES_DISABLED=1`, we won't instantiate the SysV semaphore and will, therefore, use the Ruby class. In this case, it is safe to return `false` since if it's in use, we will be using the C-extension which has the `in_use?` method.

jpittis:
Let's see if b84847e produces too many events / less events than the previous placement. I'm still thinking that adding `count=n` as a named param and only calling notify once might be the best approach.

jpittis:
I've had another look at this PR and I'm no longer concerned with the additional complexity of the code. It seems generally clear, well commented and modular.

I also want to emphasize that I'm not concerned with the performance of this code when the number of resources in the LRU is low. (Think less than 100.) I believe that for low values of N, the performance of this store is negligibly slower than the previous implementation.

Because we do not limit the number of resources that the LRU can hold, just like with the previous hash implementation, there is a risk that a high influx of resource creation could lead to high memory growth. However, my main concern isn't memory usage but is instead the speed of the eviction algorithm for high values of N.

Specifically, when there are a large number of resources in the LRU who are `in_use?` but who are also older than the `MIN_TIME_IN_LRU`, the algorithm will iterate over all of them on every call to `set`.

My measurements show that `clear_unused_resources` takes ~0.1ms for 100 resources, 1ms for 1000 resources and ~10ms for 10000 resources.

The conclusion being that when there are on the order of thousands of long lasting resources found in the LRU, creation of new resources (which triggers eviction) will take on the order of milliseconds longer than the current implementation. It's unclear to me whether this would have serious impact on the request path.

Because of this behaviour I have some suggestions:

1. I suggest if we ship this PR, the LRU implementation should be behind a flag. Existing users with large circuit breaker pools (in the thousands) will get an unexpected slowdown from this change.

2. Users who do take advantage of the LRU implementation will need to monitor the number of existing resources in the LRU. If this number passes the hundreds, care will need to be taken to ensure performance isn't negatively impacted.

3. If we need this thing to perform well for resource pools measured in the thousands, we'll need to tweak and improve this algorithm.

All this being said, I've never seen the pool get larger than 500 on production workloads. This still means that `clear_unused_resources` has a worst case of `0.5` ms. Which I would argue is too high. What is still unclear to me is how likely the algorithm behaves at its worst case in production. In my benchmarks I've purposefully triggered worst case behaviour.

[Here's a gist of my benchmark code.](https://gist.github.com/jpittis/c4bbf78f3d6f4fe4cfef18345de9d025)

To disprove my conclusion, we would need to show that either:

1. The LRU performs nowhere near at worst-case for production workloads.
2. The performance of `set` doesn't matter relative to `get`.

PR #195, opened 2018-09-18
Memcached 1.8.0 bulkhead

maximebedard:
First stab at implementing bulkheading for `memcached@1.8.0`. Since the lib already implements circuit breaking, this only adds bulkheading based on a given name.

If we want to be more granular, I can inject the current peer in the semian identifier, but this will be problematic for multi-* operations.

@csfrancis @dylanahsmith Any thoughts?

### DIFF
--- a/lib/semian/memcached.rb
+++ b/lib/semian/memcached.rb
@@ -15,18 +15,16 @@ module Semian
     ResourceBusyError = Class.new(SemianError)
 
     def semian_identifier
-      name = semian_options && semian_options[:name]
-      ["memcached", name].compact.join("_").to_sym
+      @semian_identifier ||= begin
+        name = semian_options && semian_options[:name]
+        ["memcached", name].compact.join("_").to_sym
+      end
     end
 
     def raw_semian_options
-      # disable semian when host ejection is enabled.
-      return false if !!@options[:auto_eject_hosts]
       @options[:semian].merge(circuit_breaker: false)
     end
 
-    # patch all the methods that actually interacts with memcached, not their higher level abstractions such
-    # as #get or #cas.
     %i(
       set
       add
@@ -37,11 +35,9 @@ module Semian
       prepend
       delete
       exist
-      single_get
-      single_cas
-      multi_get
-      multi_cas
     ).each do |meth|
+      raise "Memcached##{meth} is not defined." unless ::Memcached.method_defined?(meth)
+
       define_method(meth) do |*args|
         acquire_semian_resource(adapter: :memcached, scope: :query) do
           super(*args)
@@ -49,13 +45,22 @@ module Semian
       end
     end
 
-    # Preserve single_get and single_cas private visibility
-    private(
-      :single_get,
-      :single_cas,
-      :multi_get,
-      :multi_cas,
-    )
+    private
+
+    %i(
+      single_get
+      single_cas
+      multi_get
+      multi_cas
+    ).each do |meth|
+      raise "Memcached##{meth} is not defined." unless ::Memcached.private_method_defined?(meth)
+
+      define_method(meth) do |*args|
+        acquire_semian_resource(adapter: :memcached, scope: :query) do
+          super(*args)
+        end
+      end
+    end
   end
 end
 
--- a/lib/semian/memcached.rb
+++ b/lib/semian/memcached.rb
@@ -1,0 +1,67 @@
+require 'semian/adapter'
+require 'memcached'
+
+module Semian
+  module Memcached
+    include Semian::Adapter
+
+    class SemianError < ::Memcached::Error
+      def initialize(semian_identifier, *args)
+        super(*args)
+        @semian_identifier = semian_identifier
+      end
+    end
+
+    ResourceBusyError = Class.new(SemianError)
+
+    def semian_identifier
+      name = semian_options && semian_options[:name]
+      ["memcached", name].compact.join("_").to_sym
+    end
+
+    def raw_semian_options
+      # disable semian when host ejection is enabled.
+      return false if !!@options[:auto_eject_hosts]
+      @options[:semian].merge(circuit_breaker: false)
+    end
+
+    # patch all the methods that actually interacts with memcached, not their higher level abstractions such
+    # as #get or #cas.
+    %i(
+      set
+      add
+      increment
+      decrement
+      replace
+      append
+      prepend
+      delete
+      exist
+      single_get
+      single_cas
+      multi_get
+      multi_cas
+    ).each do |meth|
+      define_method(meth) do |*args|
+        acquire_semian_resource(adapter: :memcached, scope: :query) do
+          super(*args)
+        end
+      end
+    end
+
+    # Preserve single_get and single_cas private visibility
+    private(
+      :single_get,
+      :single_cas,
+      :multi_get,
+      :multi_cas,
+    )
+  end
+end
+
+if Memcached::VERSION == "1.8.0"
+  # Hack to go around the DEFAULTS arguments check
+  Memcached::DEFAULTS.merge!(semian: nil)
+
+  Memcached.prepend(Semian::Memcached)
+end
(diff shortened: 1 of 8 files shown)


lib/semian/memcached.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    def semian_identifier
+      name = semian_options && semian_options[:name]
+      ["memcached", name].compact.join("_").to_sym
+    end
+
+    def raw_semian_options
+      # disable semian when host ejection is enabled.
+      return false if !!@options[:auto_eject_hosts]
+      @options[:semian].merge(circuit_breaker: false)
```
csfrancis: I guess this is okay because we'll still be protected by the bulkheads. Is this implicitly relying on the internal memcached circuit breakers to disable bad hosts?

lib/semian/memcached.rb
```diff
@@ -0,0 +1,67 @@
...
+    private(
+      :single_get,
+      :single_cas,
+      :multi_get,
+      :multi_cas,
+    )
+  end
+end
+
+if Memcached::VERSION == "1.8.0"
```
csfrancis: Do you know if there's anything specific to 1.8.0? Could this apply to older/other versions as well?

test/memcached_test.rb
```diff
@@ -0,0 +1,101 @@
...
+    client.set("foo", "bar")
+
+    Semian[:memcached_testing].acquire do
+      assert_raises(Memcached::ResourceBusyError) do
+        client.get("foo")
+      end
+    end
+  end
+
+  # def test_resource_timeout_on_query
```
csfrancis: How come this test is commented out?

lib/semian/memcached.rb
maximebedard: It could possibly work for previous versions, but not `memcached@master`.

lib/semian/memcached.rb
maximebedard: Correct 👍

test/memcached_test.rb
maximebedard: It was flaky, currently investigating why :/

dylanahsmith:
> If we want to be more granular, I can inject the current peer in the semian identifier, but this will be problematic for multi-* operations.

Note that it looks like we could do this with libmemcached, where retrieving multiple values is done using `memcached_mget` then results are retrieved using multiple calls to `memcached_fetch_result`, such that we could decrement semaphores as the results are coming in.  However, this wouldn't work with semian's acquire_semian_resource method or with the memcached gem's multi_get method that retrieves all the results.

lib/semian/memcached.rb
```diff
@@ -0,0 +1,67 @@
...
+      append
+      prepend
+      delete
+      exist
+      single_get
+      single_cas
+      multi_get
+      multi_cas
+    ).each do |meth|
+      define_method(meth) do |*args|
```
dylanahsmith: I think we should raise `unless Memcached.method_defined?(meth)` so that things fail in an obvious way if one of these private methods gets removed or renamed in the memcached gem.  That seems quite likely given that the upstream master branch is doing a big refactor to remove the use of swig to generate the C extension.

lib/semian/memcached.rb
```diff
@@ -0,0 +1,67 @@
...
+    ResourceBusyError = Class.new(SemianError)
+
+    def semian_identifier
+      name = semian_options && semian_options[:name]
+      ["memcached", name].compact.join("_").to_sym
+    end
+
+    def raw_semian_options
+      # disable semian when host ejection is enabled.
+      return false if !!@options[:auto_eject_hosts]
```
dylanahsmith: Why does this matter? Since the ticket is taken for the group of servers and not for a specific server with the current implementation.

lib/semian/memcached.rb
```diff
@@ -0,0 +1,67 @@
...
+        super(*args)
+        @semian_identifier = semian_identifier
+      end
+    end
+
+    ResourceBusyError = Class.new(SemianError)
+
+    def semian_identifier
+      name = semian_options && semian_options[:name]
+      ["memcached", name].compact.join("_").to_sym
```
dylanahsmith: Couldn't this be memoized?

lib/semian/memcached.rb
dylanahsmith: Also, I'm not sure why you are using `!!` in an `if` condition.

maximebedard: Yep that's my bad, forgot that from my previous experiments.

lib/semian/memcached.rb
maximebedard: Yep we could. TBH, we are not looking at moving to memcached@master, but actually switching over to a pure ruby client such as Dalli. There's still a bit of work to do in order to acheive this, but it's something we want to do somewhere in the near future.

maximebedard:
> using memcached_mget then results are retrieved using multiple calls to memcached_fetch_result

Do you think it's worth doing something like this? I assume I would have to patch the memcached lib to provide an alternative implementation for get that actually uses both of theses methods. Do you think bulkheading for a group of server like the current implementation would be useful enough ?

dylanahsmith:
Doing bulkheading on a per-server basis would be very useful, since we have definitely run into single servers being overloaded while others have been fine, due to hot keys.

If we are doing bulkheading for the whole cluster of cache servers, then it seems like it will be harder to tune for both the situation where the operations are close to equality distributed across servers and for when the operations are skewed towards a single server due to hot keys.  If we tune it for equal distribution, then the memcached server might get overloaded without running out of semian tickets.  If we tune it for a skewed load towards one server, then we could artificially limit our maximum throughput when handling equality distributed operations. I'm not really sure how much of a problem it will end up being, but it does seem like it will make tuning the configuration harder.

Mostly I was just trying to point out that this wasn't a limitation of libmemcached library, just of the memcached gem.  Ideally this would be something we could refactor upstream so that we could then wrap the method.  Dalli has a `get_multi_yielder` method that yields each result, so doing something similar in the memcached gem in a private method would allow the method to simply be wrapped.  However, even then the solution would require a significant semian refactor and the code would be a lot more complicated, so it might not be worth it unless we know we need this level of granularity.

At least it will help that libmemcached will do the circuit breakers on a per-server basis.

lib/semian/memcached.rb
dylanahsmith: Well, there is also a memcached 1-0-stable branch.  Also, it doesn't look like the tests actually cover all these methods, so I still think it makes sense to add the simple runtime check.

maximebedard:
@fbogsany @alexsnaps I'd love to have your thoughts around this. This could potentially help for a pool of server, but targeting specific instances would require quite a bit of changes in the memcached gem, which I would try to avoid as much as possible 😄

sirupsen:
I think it would be useful to revisit timeouts as well. I'd expect Memcached to have extremely consistent performance under normal operation, but even a 10ms timeout could be catastrophic for capacity if you're doing a large number of calls. Bulkheading _does_ solve that issue, but it's harder to reason about and I have a feeling it could be equally solved with good timeouts given that Memcached generally doesn't have long-tail responses like MySQL AFAIK? 👂

csfrancis:
> I feel that addressing this at the ring level isn't the way.

I'm inclined to agree with this. I feel Semian should cover protecting a memcached "resource".  I don't think that Semian should care if that resource actually maps to multiple instances. Really, I think this is where we're looking to go in the mid-long term: we'll have a proxy that is aware of and controls access to the backend memcached instances. That proxy should have its own resiliency and health checking. 

Ideally, our Shopify core configuration will evolve to have a single memcached endpoint per pod that points to a proxy. When that happens all of the code to support multiple memcached instances will be 💀-- I'd be willing to hold off on writing that code in the first place and deal with masking the current memcached ring behind a single resource.

fbogsany:
> Memcached generally doesn't have long-tail responses like MySQL AFAIK?

I *believe* that's true, but we currently see 5 second spikes due to DNS issues and those mask any other problems. We're working on a fix for that. We'll have a better understanding of Memcached response characteristics 🔜

maximebedard:
I will close this PR as we are effectively moving toward using a proxy that will hide all of this complexity from the client.

PR #196, opened 2018-09-18
Addressing comments from #192

max611:
Follow up on https://github.com/Shopify/semian/pull/192

- Test the logging from one state to another.
- Make `open` and `close` consistent with `half_open!`
- Use `Timecop.travel` without `Time.now`

### DIFF
--- a/test/circuit_breaker_test.rb
+++ b/test/circuit_breaker_test.rb
@@ -101,12 +101,10 @@ class TestCircuitBreaker < Minitest::Test
   def test_open_close_open_cycle
     resource = Semian.register(:open_close, tickets: 1, exceptions: [SomeError], error_threshold: 2, error_timeout: 5, success_threshold: 2)
 
-    half_open_future_travel = -> { Time.now + resource.circuit_breaker.error_timeout + 1 }
-
     open_circuit!(resource)
     assert_circuit_opened(resource)
 
-    Timecop.travel(half_open_future_travel.call) do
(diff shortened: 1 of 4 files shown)


max611:
Really not convinced on the `Logger` tests, let me know if you have any ideas on how to make it clearer.

test/circuit_breaker_test.rb
```diff
@@ -20,10 +21,15 @@ def test_acquire_yield_when_the_circuit_is_closed
   end
 
   def test_acquire_raises_circuit_open_error_when_the_circuit_is_open
-    open_circuit!
+    Semian.logger = Logger.new($stderr)
```
pushrax: It should be possible to use a StringIO as the logger and inspect it directly without any fd tricks.

test/circuit_breaker_test.rb
```diff
@@ -117,7 +120,7 @@ def test_open_close_open_cycle
       open_circuit!(resource)
       assert_circuit_opened(resource)
 
-      Timecop.travel(half_open_future_travel.call) do
+      Timecop.travel(resource.circuit_breaker.error_timeout + 1) do
```
jpittis: This isn't equivalent to the previous code right?

`Time.now` was returning two different values for the two `half_open_future_travel.call` calls.

jpittis: TIL!

PR #198, opened 2018-09-28
Add half_open_resource_timeout for net http

max611:
The `half_open_resource_timeout` feature was introduced here https://github.com/Shopify/semian/pull/188 but only for mysql2. 

This PR adds it for the `net_http` adapter. `Net::HTTP` let us assign the timeout by using [`read_timeout=` ](https://github.com/ruby/ruby/blob/trunk/lib/net/http.rb#L769).

### DIFF
--- a/lib/semian/circuit_breaker.rb
+++ b/lib/semian/circuit_breaker.rb
@@ -125,15 +125,14 @@ module Semian
     end
 
     def maybe_with_half_open_resource_timeout(resource, &block)
-      result = nil
-
-      if half_open? && @half_open_resource_timeout && resource.respond_to?(:with_resource_timeout)
-        resource.with_resource_timeout(@half_open_resource_timeout) do
-          result = block.call
+      result =
+        if half_open? && @half_open_resource_timeout && resource.respond_to?(:with_resource_timeout)
+          resource.with_resource_timeout(@half_open_resource_timeout) do
+            block.call
+          end
+        else
+          block.call
         end
-      else
-        result = block.call
-      end
 
       result
     end
--- a/lib/semian/net_http.rb
+++ b/lib/semian/net_http.rb
@@ -90,6 +90,19 @@ module Semian
       end
     end
 
+    def with_resource_timeout(timeout)
+      prev_read_timeout = read_timeout
+      prev_open_timeout = open_timeout
+      begin
+        self.read_timeout = timeout
+        self.open_timeout = timeout
+        yield
+      ensure
+        self.read_timeout = prev_read_timeout
+        self.open_timeout = prev_open_timeout
+      end
+    end
+
     private
 
     def handle_error_responses(result)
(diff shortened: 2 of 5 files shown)


test/net_http_test.rb
```diff
@@ -42,6 +42,40 @@ def test_semian_identifier
     end
   end
 
+  def test_changes_timeout_when_half_open_and_configured
```
max611: I'm not sure if I'm doing this right, is there a better way of half opening the circuit?

lib/semian/circuit_breaker.rb
```diff
@@ -125,15 +125,14 @@ def disabled?
     end
 
     def maybe_with_half_open_resource_timeout(resource, &block)
-      result = nil
-
-      if half_open? && @half_open_resource_timeout && resource.respond_to?(:with_resource_timeout)
-        resource.with_resource_timeout(@half_open_resource_timeout) do
-          result = block.call
+      result =
```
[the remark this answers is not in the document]
max611: Small bikeshed, I can bring back the old syntax. I am used to assigning the value directly from the if, not sure what syntax you guys prefer.

test/net_http_test.rb
max611: I think I'll need help with this test, I can't seem to be able to trigger the right thing. I do hit the `with_resource_timeout` path but `ReadTimeout` does not raise.

lib/semian/net_http.rb
```diff
@@ -90,6 +90,16 @@ def transport_request(*)
       end
     end
 
+    def with_resource_timeout(timeout)
+      prev_read_timeout = self.read_timeout
```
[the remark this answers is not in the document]
max611: `write_timeout` is not out yet :(

test/net_http_test.rb
```diff
@@ -396,6 +432,12 @@ def test_persistent_state_after_server_restart
 
   private
 
+  def half_open_cicuit!(backwards_time_travel = 10)
```
max611: @sirupsen I wanted to use the CircuitBreakerHelper but the `half_open_circuit!` method uses a `open_circuit!` method with an argument and when I call `half_open_circuit!` from the `net_http_test` file, it calls the `open_circuit!` method inside `net_http_test` instead of the `open_circuit!` method inside the helper which raises. To make it work, I would have to change the `open_circuit!` method name everywhere. What do you think of simply keeping the `half_open_cicuit!` method in the `net_http_test` file?

sirupsen: That's fine with me.

lib/semian/circuit_breaker.rb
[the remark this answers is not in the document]
jpittis: Just for context: The reason this kind of thing might frowned upon is because it causes a 7 line diff that we have to review manually to ensure that a typo size bug isn't added to the library.

PR #199, opened 2018-10-09
Add last error message to circuit breaker

max611:
Related to https://github.com/Shopify/cusco/issues/193

As of now, here's what a Semian error will look like while debugging:
error_class: `CircuitOpenError`
error_message: `[service.website.com] CircuitOpenError`

With this change, we would add the last error message to the error_message, it would become something similar to `CircuitOpenError caused by Net::ReadTimeout`

Let me know what you think.

### DIFF
--- a/lib/semian/adapter.rb
+++ b/lib/semian/adapter.rb
@@ -35,8 +35,12 @@ module Semian
         mark_resource_as_acquired(&block)
       end
     rescue ::Semian::OpenCircuitError => error
-      last_error_message = semian_resource.circuit_breaker.last_error_message
-      raise self.class::CircuitOpenError.new(semian_identifier, "#{error.message} caused by #{last_error_message}")
+      begin
+        raise semian_resource.circuit_breaker.last_error
+      rescue
+        last_error_message = semian_resource.circuit_breaker.last_error.message
+        raise self.class::CircuitOpenError.new(semian_identifier, "#{error.message} caused by #{last_error_message}")
+      end
     rescue ::Semian::BaseError => error
       raise self.class::ResourceBusyError.new(semian_identifier, error.message)
     rescue *resource_exceptions => error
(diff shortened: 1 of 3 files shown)


lib/semian/adapter.rb
```diff
@@ -35,7 +35,8 @@ def acquire_semian_resource(scope:, adapter:, &block)
         mark_resource_as_acquired(&block)
       end
     rescue ::Semian::OpenCircuitError => error
-      raise self.class::CircuitOpenError.new(semian_identifier, error.message)
+      last_error_message = semian_resource.circuit_breaker.last_error_message
+      raise self.class::CircuitOpenError.new(semian_identifier, "#{error.message} caused by #{last_error_message}")
```
pushrax: Could we perhaps hold onto the original exception object, and set it as the proper error cause? Then the full stack will show up in error reporting.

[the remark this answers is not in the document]
pushrax: "caused by" is accurate imo, but your alternatives are a bit more precise. Should be perfect after changing the wording here a little, and setting the exception cause if that's possible.

max611: Just to make sure I understand you well, you mean instead of raising `CircuitOpenError` we would raise for example `Net::Timeout` ?

max611: First time working with the `cause` method, so I might be missing something. It seems like I can't set the `cause` value easily. It seems like I need to raise the last error then catch it so that the `cause` value get sets to the `last_error`. I could, however, patch the `SemianError` class to add a `cause` method but I'm not sure if that's the right thing to do.

I've looked into [this](https://ruby-doc.org/core-2.5.0/Exception.html#method-i-cause) and [this](http://www.virtuouscode.com/2013/12/25/exception-causes-in-ruby-2-1/) to get a better idea of how the `cause` method works.

jpittis: We should find out how bugsang displays this before choosing to go with this method.

max611: I tested it on a small rails app and here's how it looks like:
image

I think it make sense and it's very clear. I'll go with this 👍

max611: @pushrax Let me know if the change makes sense!

PR #200, opened 2018-10-12
Add GRPC adapter

max611:
#### Methods
GRPC uses multiple methods to call the server depending on the usage, we need to implement all of them which are `request_response`, `client_streamer`, `server_streamer`, `bidi_streamer`.

To find these methods, I based myself on the Interceptor. You can find the methods here: https://github.com/grpc/grpc/blob/master/src/ruby/lib/grpc/generic/interceptors.rb#L32

#### Errors
GRPC has the BadStatus class that is used for all the errors listed [here](https://github.com/grpc/grpc/blob/master/src/ruby/lib/grpc/errors.rb). As of the `resource_exceptions`, I am still working on it and will update the list with the right erorrs.

#### Test
For now I've only written some basic test to show that the circuit can open on all outbound methods and that it uses the right identifier. I added Mocha as I feel it's easier to stub things with it and I could avoid some flakiness around raising errors.

I based myself on the mysql adapter for the tests but I will certainly add more based on the net_http adapter. Feel free to let me know what kind of tests I might miss that we need.

#### How to
Basically, to use the adapter, we would need to do something like that:
```
ClientStub.new(host, :this_channel_is_insecure, {semian_options: SEMIAN_OPTIONS})
```

I am not sure about the way I am setting up the `semian_options` in the initializer but I can't seem to think about a smarter way, let me know if you have ideas.

Stub creation in Panama: https://github.com/Shopify/panama/blob/master/ruby/lib/shopify_panama/client.rb#L47

#### Alternative
One thing I tried, is using an Interceptor.
As you can read [here](https://github.com/grpc/grpc/blob/618a3f561d4a93f263cca23abad086ed8f4d5e86/src/ruby/lib/grpc/generic/client_stub.rb#L92), _the [`GRPC::ClientInterceptor`](https://github.com/grpc/grpc/blob/master/src/ruby/lib/grpc/generic/interceptors.rb) object will be used for intercepting calls before they are executed_. This allow us to wrap all the calls with semian. You can see the Interceptor spec [here](https://github.com/grpc/grpc/blob/master/src/ruby/spec/generic/client_interceptors_spec.rb) which gives more details on how it works. Since it seems like an experimental API, I don't think we should use it.


max611:
I've updated the PR to use the `interceptor` feature from the GRPC stub.

#### Interceptor
As you can read [here](https://github.com/grpc/grpc/blob/618a3f561d4a93f263cca23abad086ed8f4d5e86/src/ruby/lib/grpc/generic/client_stub.rb#L92), _the [`GRPC::ClientInterceptor`](https://github.com/grpc/grpc/blob/master/src/ruby/lib/grpc/generic/interceptors.rb) object will be used for intercepting calls before they are executed_. This allow us to wrap all the calls with semian. You can see the Interceptor spec [here](https://github.com/grpc/grpc/blob/master/src/ruby/spec/generic/client_interceptors_spec.rb) which gives more details on how it works.

#### Calls
GRPC uses multiple methods to call the server depending on the usage, we need to implement all of them which are `request_response`, `client_streamer`, `server_streamer`, `bidi_streamer`.

#### Errors
GRPC has the BadStatus class that is used for all the errors listed [here](https://github.com/grpc/grpc/blob/master/src/ruby/lib/grpc/errors.rb). As of the `resource_exceptions`, I am still working on it and will update the list with the right erorrs.

#### Test
For now I've only written some basic test to show that the circuit can open and that it uses the right identifier. The goal will be to write test for all the methods and test out the various errors we are expecting.

Let me know what you think and what could be improved. Is `GRPC::ClientInterceptor` the right way to go?

lib/semian/grpc.rb
```diff
@@ -0,0 +1,80 @@
+require 'semian/adapter'
+require 'grpc'
+
+module GRPC
+  GRPC::BadStatus.include(::Semian::AdapterError)
+
+  class SemianError < GRPC::BadStatus
```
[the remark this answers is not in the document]
sirupsen: This seems problematic for Panama because it [assumes a detail is included for BadStatus](https://github.com/Shopify/panama/blob/master/ruby/lib/shopify_panama/errors.rb). So if we choose this error, we need `#details` to be implemented.

sirupsen:
Let us know when you need a new review on this Max!

@ewalk153 @pushrax

lib/semian/grpc.rb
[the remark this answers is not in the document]
sirupsen: Ok, I'll check back when you have :)

lib/semian/grpc.rb
```diff
@@ -0,0 +1,72 @@
...
+    attr_reader :details
+
+    def initialize(semian_identifier, *args)
+      super(*args)
+      @details = args.first
```
pushrax: Maybe just delegate `details` to `message`?

lib/semian/grpc.rb
```diff
@@ -0,0 +1,80 @@
...
+      end
+    end
+
+    Semian::GRPC.reset_exceptions
+
+    def resource_exceptions
+      Semian::GRPC.exceptions
+    end
+
+    def request_response(*args)
```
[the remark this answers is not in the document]
pushrax: From a quick read of https://github.com/grpc/grpc/blob/master/src/ruby/ext/grpc/rb_channel.c it appears the connection can be established in the background in another thread https://github.com/grpc/grpc/pull/9986, possibly also via `Core::Channel.new` somehow.

Given calls don't trigger connections, if they only ever happen in another thread, maybe Semian doesn't need to care since slow or failing connect attempts won't influence call latency. There seems to be [a TODO](https://github.com/grpc/grpc/blob/a786b7415aaaf7247293f4116007aa451652be06/src/ruby/ext/grpc/rb_channel.c#L733) to change this though.

lib/semian/grpc.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+
+    def resource_exceptions
+      [
+        ::GRPC::DeadlineExceeded,
+        ::GRPC::InvalidArgument,
+        ::GRPC::Cancelled,
+        ::GRPC::Unknown,
+        ::GRPC::NotFound,
+        ::GRPC::PermissionDenied,
```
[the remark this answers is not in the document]
jpittis: To emphasize what Justin is saying, `InvalidArgument`, `Cancelled`, `Unknown`, `NotFound`, `PermissionDenied` (And maybe others?) are not signs that the upstream is failing. The general idea of circuit breakers is we want to use errors returned by the client library as heuristics to trigger fast failure and decrease load on the upstream.

IMO we should filter this list to only include errors that are obvious failure scenarios of the upstream.

test/grpc_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    services.each do |s|
+      server.handle(s)
+    end
+    t = Thread.new { server.run }
+    server.wait_till_running
+
+    yield
+
+    server.stop
+    t.join
```
jpittis: What happens when `yield` throws an exception? Do we leak server threads?

test/grpc_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    @hostname = '0.0.0.0'
+    @server = new_rpc_server_for_testing({poll_period: 1}.merge(server_opts))
+    @port = @server.add_http2_port("#{@hostname}:0", :this_port_is_insecure)
+    @host = "#{@hostname}:#{@port}"
+    @client_opts = client_opts
+    populate_toxiproxy
+    @server
+  end
+
+  def populate_toxiproxy
```
jpittis: Why does this need to be dynamically populated vs using [hosts.yml.](https://github.com/Shopify/semian/blob/22e6dcf4d4425752a71db5e6284046cdc03fa1c4/test/config/hosts.yml)

test/grpc_test.rb
max611: I took that part from the GRPC test helper: https://github.com/grpc/grpc/blob/master/src/ruby/spec/support/helpers.rb

What do you think I should do to make sure we won't leak server threads?

test/grpc_test.rb
max611: `@server.add_http2_port("#{@hostname}:0", :this_port_is_insecure)` returns a new port everytime I run it. I tried setting up a port directly but it does not work and had to populate toxiproxy on the go.

max611: Played with it a bit and I can use the `hosts.yml` but I am getting an error when trying to initialize the server in the test. `RuntimeError: could not add port 127.0.0.1:50051 to server, not sure why` I'll look into it

max611: Okay all good, fixed 👍 Thank you for the comment, forgot about that file!

test/grpc_test.rb
jpittis: Stick an ensure after the yield. IMO. @pushrax agree?

test/grpc_test.rb
jpittis: Awesome!

max611:
Hey @pushrax and @jpittis I pushed a small change to the PR. In order to test the adapter without implementing it directly into Panama, we figured it would be easier to use the same system as the `net_http` adapter. (https://github.com/Shopify/panama/issues/447#issuecomment-447119558) This will allow us to initialize the GRPC adapter on another app before core and we will avoid touching Panama for now.

sirupsen:
We'll need to find an alternative reviewer for @pushrax.

@byroot do you have cycles? ❤️

lib/semian/grpc.rb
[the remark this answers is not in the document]
pushrax: Great explanation Jake.

test/grpc_test.rb
pushrax: Plus a `begin` before the thread allocation 👍

lib/semian/grpc.rb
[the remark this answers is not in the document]
casperisfine: Isn't there some kind of `BaseConnectionError`, it's usually the ones for the best semantic for semian.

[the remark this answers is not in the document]
max611: From looking at https://github.com/grpc/grpc/blob/master/src/ruby/lib/grpc/errors.rb there is nothing similar to `BaseConnectionError` 😞 . The main reason I picked `BadStatus` is that it seemed to be the base class of all the errors, I am not sure which error would be the best based on the error list. From the list, are you seeing a better one?

lib/semian/grpc.rb
```diff
@@ -26,14 +26,40 @@ module GRPC
...
+
+      def retrieve_semian_configuration(host)
+        @semian_configuration.call(host) if @semian_configuration.respond_to?(:call)
+      end
+    end
+
+    def raw_semian_options
+      @raw_semian_options ||= begin
+        if @host.empty?
+          host = @ch.target
```
jpittis: I'm finding this stuff pretty confusing. I have no idea what `@ch` is for example. Why do we go there to get the host? Why is `@host` not present? Could we add a comment?

test/config/hosts.yml
```diff
@@ -12,3 +12,7 @@ redis_toxiproxy_port: 16379
 http_host: 127.0.0.1
 http_port: 31050
 http_toxiproxy_port: 31051
+
+grpc_host: 127.0.0.1
+grpc_port: 50051
+grpc_toxiproxy_port: 50052
```
ewalk153: don't forget to end with \n

test/grpc_test.rb
```diff
@@ -0,0 +1,204 @@
...
+  end
+
+  def build_insecure_stub(klass, host: nil, opts: nil)
+    host ||= @host
+    opts ||= @client_opts
+    klass.new(host, :this_channel_is_insecure, **opts)
+  end
+
+  def build_rpc_server(server_opts: {}, client_opts: {})
+    @hostname = '127.0.0.1'
```
ewalk153: This should probably be `SemianConfig['grpc_host']`

test/grpc_test.rb
```diff
@@ -0,0 +1,204 @@
...
+    @server = new_rpc_server_for_testing({poll_period: 1}.merge(server_opts))
+    @port = @server.add_http2_port("#{@hostname}:50051", :this_port_is_insecure)
```
ewalk153: This should probably be `SemianConfig['grpc_port']`

lib/semian/grpc.rb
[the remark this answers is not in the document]
casperisfine: I'm not totally sure of the semantic of each, but `Unavailable` seems like the best candidate.

For context we want Semian exceptions to be handled by the caller like they would handle connection issues. 

So if they have code like this:

```ruby
begin
  grpc.call
rescue GRPC::NotFound
  return nil
rescue GRPC::Unavailable
  # we want Semian exceptions to trigger that handler.
end
```

This way people don't have to explicitly handle semian, it's just a special case of a regular network issue.

Now it's best to check whatever error the library send if it can't connect, but I suppose it's `Unavailable`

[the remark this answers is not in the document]
max611: I've updated it to `Unavailable` If the client can't connect to the server, it will, in fact, return an `Unavailable` error. 👍

max611:
Ready for a second round of reviews @jpittis @byroot

lib/semian/grpc.rb
```diff
@@ -0,0 +1,94 @@
+require 'semian/adapter'
+require 'grpc'
+
+module GRPC
+  GRPC::Unavailable.include(::Semian::AdapterError)
```
casperisfine: It's possible we'd want to include it to some other errors, if they are indicative of a bad health of the remote service.

For example I see there is an `Internal` error class. If as I understand it it's similar to an HTTP 500, it might make sense to include it.

Not sure about the other ones, I know too little about the protocol to answer.

lib/semian/grpc.rb
```diff
@@ -0,0 +1,94 @@
...
+    def semian_identifier
+      @semian_identifier ||= raw_semian_options[:name]
+    end
+
+    def resource_exceptions
+      [
+        ::GRPC::DeadlineExceeded,
+        ::GRPC::ResourceExhausted,
+        ::GRPC::Unavailable,
+        ::GRPC::BadStatus,
```
casperisfine: Since `BadStatus` is the parent of all GRPC errors, we likely don't want to list it here.

This list should only contain "transient" errors, e.g. errors we have reason to believe are due to general bad health of the service, and are likely to be resolved later on.

So this shouldn't include things like validation errors, bad requests, authentication error, etc.

Additionally, ideally all classes in the list should include `Semian::AdapterError`. It's not absolutely mandatory, but it's useful to distinguish which backend is actually in bad health.

max611: I'll take a deeper look on each error. Thanks!

max611: Finally found the right documentation for errors: https://grpc.io/grpc/core/md_doc_statuscodes.html

max611: For reference, there's more details about each errors here: https://grpc.io/grpc/core/md_doc_statuscodes.html

lib/semian/grpc.rb
```diff
@@ -0,0 +1,96 @@
...
+    end
+
+    def resource_exceptions
+      [
+        ::GRPC::DeadlineExceeded,
+        ::GRPC::ResourceExhausted,
+        ::GRPC::Unavailable,
+        ::GRPC::Unknown,
+      ]
+    end
```
max611: Took there errors from https://grpc.io/grpc/core/md_doc_statuscodes.html

lib/semian/grpc.rb
```diff
@@ -0,0 +1,96 @@
...
+    def request_response(*args)
+      acquire_semian_resource(adapter: :grpc, scope: :request_response) { super(*args) }
+    end
+
+    def client_streamer(*args)
+      acquire_semian_resource(adapter: :grpc, scope: :client_streamer) { super(*args) }
+    end
+
+    def server_streamer(*args)
+      acquire_semian_resource(adapter: :grpc, scope: :server_streamer) { super(*args) }
```
casperisfine: Minor nitpick. To forwards all params like this you can do:

```ruby
 def server_streamer(*)
      acquire_semian_resource(adapter: :grpc, scope: :server_streamer) { super }
 end
```

ewalk153:
A word on that @byroot and @jpittis
We will first test with the theme store, and then look at testing the App Store as well. We will be able to test some failure modes with the theme store staging.

PR #204, opened 2018-11-07
Also support a variant of resolve errors (hiredis)

maximebedard:
Hiredis also returns a different error message in case of resolving errors. This make sure this use case is also handled.


maximebedard:
Actually, I don't think this is necessary as the error message is specific to OSX.

PR #209, opened 2018-12-04
Instrument success on `acquire_circuit_breaker`

max611:
We currently only emit a `success` event on the `acquire_bulkhead` method, this causes a problem when we want to deactivate the bulkheads and still keep track of the successes.

[Event](https://github.com/Shopify/semian/blob/47bee8756cc84a198073acc0e148baed1b57d6fa/lib/semian/protected_resource.rb#L41) on circuit open for `acquire_circuit_breaker`.

Success [event](https://github.com/Shopify/semian/blob/47bee8756cc84a198073acc0e148baed1b57d6fa/lib/semian/protected_resource.rb#L50) on `acquire_bulkead`.

I've added `if @bulkhead.nil?` to avoid emitting two successes.

Related graph that went to 0 once bulkhead were disabled:
image

### DIFF
--- a/lib/semian/protected_resource.rb
+++ b/lib/semian/protected_resource.rb
@@ -21,7 +21,8 @@ module Semian
 
     def acquire(timeout: nil, scope: nil, adapter: nil, resource: nil)
       acquire_circuit_breaker(scope, adapter, resource) do
-        acquire_bulkhead(timeout, scope, adapter) do
+        acquire_bulkhead(timeout, scope, adapter) do |_, wait_time|
+          Semian.notify(:success, self, scope, adapter, wait_time || 0)
           yield self
         end
       end
@@ -34,7 +35,6 @@ module Semian
         yield self
       else
         @circuit_breaker.acquire(resource) do
-          Semian.notify(:success, self, scope, adapter) if @bulkhead.nil?
           yield self
         end
       end
@@ -48,8 +48,7 @@ module Semian
         yield self
       else
         @bulkhead.acquire(timeout: timeout) do |wait_time|
-          Semian.notify(:success, self, scope, adapter, wait_time)
-          yield self
+          yield self, wait_time
         end
       end
     rescue ::Semian::TimeoutError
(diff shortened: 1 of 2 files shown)


lib/semian/protected_resource.rb
```diff
@@ -34,6 +34,7 @@ def acquire_circuit_breaker(scope, adapter, resource)
         yield self
       else
         @circuit_breaker.acquire(resource) do
+          Semian.notify(:success, self, scope, adapter) if @bulkhead.nil?
```
pushrax: Would it make sense to delete the two current `Semian.notify` calls and put just one in the main `acquire` method?

max611: `acquire_bulkhead` adds the `wait_time`, `Semian.notify(:success, self, scope, adapter, wait_time)`. Because of that I don't think it's possible to add only one `Semian.notify` :(

pushrax: Why not use `wait_time || 0` in the common codepath?

max611: `wait_time` is added here: https://github.com/Shopify/semian/blob/d74c809554c33e452310e91bcbbf05ae07dd7aa6/lib/semian/protected_resource.rb#L50-L53

I don't think I could access `wait_time` from the common code path without adding a bit of complexity. Maybe I'm missing something, let me know.

pushrax: Right, you could `yield self, wait_time`, this seems not overly complex to me.

max611: I'm lost 🤦‍♂️ I don't know why I did not think about that. Done!

lib/semian/protected_resource.rb
```diff
@@ -47,8 +48,7 @@ def acquire_bulkhead(timeout, scope, adapter)
         yield self
```
pushrax: Maybe `yield self, 0` here to make the output consistent? Then you can remove `|| 0` above

PR #215, opened 2019-02-25
Add metadata links

viraptor:
Add direct links in metadata. This helps with automated dependency updates/management.


viraptor:
CLA's too much work for this :(

Issue #211, opened 2018-12-07
mysql2: Exceptions are "wrapped" infinitely


tsauerwein:
We are trying to use Semian for our application. The `Net::HTTP` adapter works perfectly and I also wrote a custom adapter for [logstash-logger](https://github.com/dwbutler/logstash-logger).

Unfortunately, the MySQL adapter is behaving strangely. I was able to reproduce it in a small script:

```
require 'active_record'
require 'semian'
require 'semian/mysql2'

db_config = {
  adapter: 'mysql2',
  pool: 8,
  timeout: 2,
  host: 'toxiproxy',
  port: 3307,
  database: '...',
  username: '...',
  password: '...',
  reconnect: true,
  connect_timeout: 10,
  read_timeout: 10,
  semian: {
    name: 'test-db',
    bulkhead: false,
    success_threshold: 1,
    error_threshold: 3,
    error_timeout: 10
  }
}

ActiveRecord::Base.establish_connection(db_config)

loop do
  begin
    sleep 1
    puts ActiveRecord::Base.connection.execute('select now()')
  rescue StandardError => exc
    puts exc
  end
end
```

The output is:
```
# ruby test-semian.rb 
[mysql_test-db] Can't connect to MySQL server on 'toxiproxy' (111)
[mysql_test-db] Can't connect to MySQL server on 'toxiproxy' (111)
I, [2018-12-07T09:59:32.641238 #65]  INFO -- : [Semian::CircuitBreaker] State transition from closed to open. success_count=0 error_count=3 success_count_threshold=1 error_count_threshold=3 error_timeout=10 error_last_at="1544173172"
[mysql_test-db] Can't connect to MySQL server on 'toxiproxy' (111)
[mysql_test-db] Semian::OpenCircuitError caused by [mysql_test-db] Can't connect to MySQL server on 'toxiproxy' (111)
[mysql_test-db] Semian::OpenCircuitError caused by [mysql_test-db] Semian::OpenCircuitError caused by [mysql_test-db] Can't connect to MySQL server on 'toxiproxy' (111)
[mysql_test-db] Semian::OpenCircuitError caused by [mysql_test-db] Semian::OpenCircuitError caused by [mysql_test-db] Semian::OpenCircuitError caused by [mysql_test-db] Can't connect to MySQL server on 'toxiproxy' (111)
[...]

PR #216, opened 2019-03-06
Don't modify the last error if it already exist

max611:
Fixes https://github.com/Shopify/semian/issues/211

Since a circuit open error can trigger a circuit open error, the problem was that we were always adding a Circuit open error as a cause to new circuit open error which was getting in an infinite loop and it made reading logs really hard.

### DIFF
--- a/lib/semian/circuit_breaker.rb
+++ b/lib/semian/circuit_breaker.rb
@@ -109,7 +109,7 @@ module Semian
     end
 
     def push_error(error)
-      @last_error = error
+      @last_error ||= error
     end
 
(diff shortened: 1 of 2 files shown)


cjgu:
@casperisfine https://github.com/Shopify/semian/pull/250 should cover this right?

casperisfine:
Yes and no. It fixes #211 but it doesn't change `last_error` behavior.

Honestly I don't think keeping the original error is a good idea, if you go half open and get a different error from the original one, might as well update it.

Closing this.

PR #219, opened 2019-04-16
A small fix to NetHTTP example

dziemid:
At least in my testing the port is a number, not a string.


miry:
Sorry, just found this PR.
It is already fixed inhttps://github.com/Shopify/semian/commit/2c94cc55bccdb0136fad0d536e1469b42964f5f1

PR #220, opened 2019-04-19
Shorten long line

ewalk153:
Extract error pattern into constant to shorten long line.

Quiets rubocop.

### DIFF
--- a/lib/semian/redis.rb
+++ b/lib/semian/redis.rb
@@ -85,13 +85,15 @@ module Semian
         begin
           raw_connect
         rescue SocketError, RuntimeError => e
-          raise ResolveError.new(semian_identifier) if (e.cause || e).to_s =~ /(can't resolve)|(name or service not known)|(nodename nor servname provided, or not known)/i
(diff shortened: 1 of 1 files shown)


lib/semian/redis.rb
```diff
@@ -85,14 +85,16 @@ def connect
...
-          raise ResolveError.new(semian_identifier) if (e.cause || e).to_s =~ /(can't resolve)|(name or service not known)|(nodename nor servname provided, or not known)/i
+          raise ResolveError.new(semian_identifier) if (e.cause || e).to_s =~ RESOLVE_ERROR_PATTERN
           raise
         end
       end
     end
 
     private
 
+    RESOLVE_ERROR_PATTERN = /(can't resolve)|(name or service not known)|(nodename nor servname provided, or not known)/i
```
kirs: Would you need something like `private_constant : RESOLVE_ERROR_PATTERN`?

ewalk153: Yep

miry:
@ewalk153 Is it still relevant? There was another change to extract pattern to separate method: https://github.com/Shopify/semian/commit/190ea6e54288da74d1046dae1fb45b90da55e7fe

PR #223, opened 2019-05-26
Instrument the LRU

michaelkipper:
## What

Our current hypothesis is that on average, `clear_unused_resources` behaves like an amortized O(1) operation, rather than an O(n) one. Without a robust simulator, however, we need to collect some runtime data to disprove the hypothesis.

## How

This PR adds to the `implement_lru_cache` branch by instrumenting the `lru_hash_gc` event with stats about the garbage collection operation.

This PR also changes the signature of the `notify` method to have `payload` as the final argument so we can pass arbitrary data to subscribers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ vendor/
 nohup.out
 .vagrant
 .byebug_history
+.idea
--- a/lib/semian/lru_hash.rb
+++ b/lib/semian/lru_hash.rb
@@ -55,12 +55,15 @@ class LRUHash
 
   # This method uses the property that "Hashes enumerate their values in the
   # order that the corresponding keys were inserted." Deleting a key and
-  # re-inserting it effectively moves it to the back of the cache.
+  # re-inserting it effectively moves it to the front of the cache.
+  # Update the `updated_at` field so we can use it later do decide if the
+  # resource is "in use".
   def get(key)
     @lock.synchronize do
       found = @table.delete(key)
       if found
         @table[key] = found
+        found.updated_at = Time.now
       end
       found
     end
@@ -98,7 +101,13 @@ class LRUHash
       stop_time = Time.now - MINIMUM_TIME_IN_LRU # Don't process resources updated after this time
       @table.each do |_, resource|
         payload[:examined] += 1
+
+        # The update times of the resources in the LRU are monotonically increasing,
+        # time, so we can stop looking once we find the first resource with an
+        # update time after the stop_time.
         break if resource.updated_at > stop_time
+
+        # TODO(michaelkipper): Should this be a flag?
         next if resource.in_use?
 
         resource = @table.delete(resource.name)
--- a/lib/semian/lru_hash.rb
+++ b/lib/semian/lru_hash.rb
@@ -53,6 +53,9 @@ class LRUHash
     clear_unused_resources
   end
 
+  # This method uses the property that "Hashes enumerate their values in the
+  # order that the corresponding keys were inserted." Deleting a key and
+  # re-inserting it effectively moves it to the back of the cache.
   def get(key)
     @lock.synchronize do
       found = @table.delete(key)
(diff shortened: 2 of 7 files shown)


lib/semian/lru_hash.rb
```diff
@@ -83,16 +86,30 @@ def clear_unused_resources
...
+          examined: 0,
+          cleared: 0,
+          elapsed: nil,
+      }
+
+      stop_time = Time.now - MINIMUM_TIME_IN_LRU # Don't process resources updated after this time
       @table.each do |_, resource|
-        break if resource.updated_at + MINIMUM_TIME_IN_LRU > Time.now
+        payload[:examined] += 1
+        break if resource.updated_at > stop_time
```
csfrancis: I know you're just moving this code around, but why is this a `break` rather than a `next`?

michaelkipper: This is an LRU cache, so the table should be ordered by update time.
I was going to say that if we see a resource that's newer than the stop time, we should assume that all elements after that are also newer, but I'm not 100% convinced of that, since `get` doesn't update the `updated_at` field.
I'm going to look into this.

michaelkipper: Thanks for catching that @csfrancis.
Indeed, we should be updating `updated_at` in the call to `get` when we re-order the LRU.
I've made the fix and added a test for monotonicity.

lib/semian/lru_hash.rb
```diff
@@ -83,16 +89,36 @@ def clear_unused_resources
...
         resource = @table.delete(resource.name)
         if resource
-          Semian.notify(:lru_hash_cleaned, self, :cleaning, :lru_hash)
+          payload[:cleared] += 1
           resource.destroy
         end
       end
+
+      payload[:elapsed] = Process.clock_gettime(Process::CLOCK_MONOTONIC) - timer_start
+      Semian.notify(:lru_hash_gc, self, nil, nil, payload)
```
sirupsen: This event breaks backwards compatibility (as you see in `grpc_test`). The reason is that it's only temporary?

test/lru_hash_test.rb
```diff
@@ -127,6 +133,61 @@ def test_clean_instrumentation
     Semian.unsubscribe(subscriber)
   end
 
+  def test_monotonically_increasing
```
sirupsen: Quite the test, but breaks exactly when it should 👌

.gitignore
```diff
@@ -11,3 +11,4 @@ vendor/
 nohup.out
 .vagrant
 .byebug_history
+.idea
```
sirupsen: 😄

lib/semian/lru_hash.rb
```diff
@@ -83,16 +89,36 @@ def clear_unused_resources
...
+        # time, so we can stop looking once we find the first resource with an
+        # update time after the stop_time.
+        break if resource.updated_at > stop_time
+
+        # TODO(michaelkipper): Should this be a flag?
         next if resource.in_use?
 
         resource = @table.delete(resource.name)
         if resource
-          Semian.notify(:lru_hash_cleaned, self, :cleaning, :lru_hash)
```
thegedge: This would mean that anyone subscribing to this event would no longer get it. If we're okay with that, we should still call that out in a changelog.

lib/semian/lru_hash.rb
thegedge: Nice catch!

lib/semian/lru_hash.rb
michaelkipper: I think we'd have to pass a lot of state around to get around that.
We could optionally pass something other than `nil` for `scope` and `adapter`, if you thought that was necessary, but it would still break those tests.

lib/semian/lru_hash.rb
michaelkipper: This PR is merging to a branch that only Core uses.

thegedge: Ah, I missed we were merging into a branch :+1:

PR #225, opened 2019-05-30
Added a max_size to the LRU cache

michaelkipper:
# What

Implement a `max_size` in the LRU cache.

# Why

The data ([dashboard](https://shopify.datadoghq.com/dashboard/mzg-id4-5rp/semian-lru)) we're seeing from core suggests that the LRU cache has a p99 garbage collection time of 1.5ms, and that garbage collection is run quite frequently.

Screen Shot 2019-05-30 at 11 04 16 AM

Screen Shot 2019-05-30 at 11 04 40 AM

Screen Shot 2019-05-30 at 11 04 31 AM

(Note: I'm not sure how frequently, but the average number of evictions is 5 while the average size of the cache is about 150 and we have to examine over 100 of them every time.)

# How

This PR adds a `max_size` parameter with a sensible default based on what we've observed in core. It's my expectation that we'll do fewer garbage collections until the cache fills, and then amortize one large garbage collection. This should lead to better Semian performance on average.

### DIFF
--- a/lib/semian/lru_hash.rb
+++ b/lib/semian/lru_hash.rb
@@ -8,8 +8,6 @@ class LRUHash
   extend Forwardable
   def_delegators :@table, :size, :count, :empty?, :values
   attr_reader :table
-  MAXIMUM_SIZE_OF_LRU = 1000
-  MINIMUM_TIME_IN_LRU = 300
 
   class NoopMutex
     def synchronize(*)
@@ -49,7 +47,7 @@ class LRUHash
   #   circuits to disparate endpoints (or your circuit names are bad).
   #   If the max_size is 0, the garbage collection will be very aggressive and
   #   potentially computationally expensive.
-  def initialize(max_size: MAXIMUM_SIZE_OF_LRU, min_time: MINIMUM_TIME_IN_LRU)
+  def initialize(max_size: Semian.maximum_lru_size, min_time: Semian.minimum_lru_time)
     @max_size = max_size
     @min_time = min_time
     @table = {}
--- a/lib/semian/lru_hash.rb
+++ b/lib/semian/lru_hash.rb
@@ -8,6 +8,7 @@ class LRUHash
   extend Forwardable
   def_delegators :@table, :size, :count, :empty?, :values
   attr_reader :table
+  MAXIMUM_SIZE_OF_LRU = 1000
   MINIMUM_TIME_IN_LRU = 300
 
   class NoopMutex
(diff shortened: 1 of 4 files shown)


michaelkipper:
@sirupsen: I'm not sure how best to "set" the default `MAXIMUM_SIZE_OF_LRU` better than this, given that Semian has no constructor, per se. The cache is created when `resources` is first requested, or `reset!` is explicitly called.

For this branch, having it in `MAXIMUM_SIZE_OF_LRU` works but I'm not sure how that will upstream.

lib/semian/lru_hash.rb
```diff
@@ -34,7 +35,23 @@ def locked?
...
+  #
+  # Note:
+  #   The +min_time+ is a stronger guarantee than +max_size+. That is, if there are
+  #   more than +max_size+ entries in the cache, but they've all been updated more
+  #   recently than +min_time+, the garbage collection will not remove them and the
+  #   cache can grow without bound. This usually means that you have many active
+  #   circuits to disparate endpoints (or your circuit names are bad).
+  #   If the max_size is 0, the garbage collection will be very aggressive and
+  #   potentially computationally expensive.
+  def initialize(max_size: MAXIMUM_SIZE_OF_LRU, min_time: MINIMUM_TIME_IN_LRU)
```
sirupsen: I would say that this should be a global configuration option on `Semian`, documented in the README. `Semian.maximum_lru_size = 1000`, with this as the default when you call this class method.

lib/semian/lru_hash.rb
```diff
@@ -50,7 +67,7 @@ def set(key, resource)
       @table[key] = resource
       resource.updated_at = Time.now
     end
-    clear_unused_resources
+    clear_unused_resources if @table.length > @max_size
```
sirupsen: This prevents the clearing from being run unnecessary, however, this doesn't actually prevent it from growing big. If I query 100,000 resources (for the sake of example) within 300 seconds, they won't get cleared. Isn't the point here to actually limit it at the top of the set and then force removals? This may not be blocking now (since it doesn't seem to be the case), but I'd expect that invariant to be preserved.

michaelkipper: I called this out specifically:

```
# Note:
  #   The +min_time+ is a stronger guarantee than +max_size+. That is, if there are
  #   more than +max_size+ entries in the cache, but they've all been updated more
  #   recently than +min_time+, the garbage collection will not remove them and the
  #   cache can grow without bound. This usually means that you have many active
  #   circuits to disparate endpoints (or your circuit names are bad).
```

It's a longer conversation about what to do in this case, because it means evicting an `in_use` resource.

sirupsen: I missed that, sorry. I think it's acceptable then 👍 Having used 1,000 different resources for 300 seconds is a different one every 3 seconds. That seems pretty high and very reasonable.

sirupsen:
As part of the rollout make sure to run a production profile. I believe the handbook describes how. Pretty easy

lib/semian/lru_hash.rb
michaelkipper: Done.
I made `Semian.maximum_lru_size` and `Semian.minimum_lru_time` global configuration options.

michaelkipper:
I ran a production profile on the front page of my personal store:
https://speedscope.shopify.io/#profileURL=https%3A%2F%2Fspeedscope.shopify.io%2Fp%2Fshopify-profile%2Fshopify-production%2F2a46853c-b806-48fb-9c58-29eaeccbb756.json

`Semian::ProtectedResource#acquire_bulkhead` was the only Semian span to show up, and took 1.61ms (0.5% of the total request time of 318ms.

michaelkipper:
This is deployed to Shopify core at 100%

Screen Shot 2019-06-12 at 3 32 17 PM
Screen Shot 2019-06-12 at 3 33 32 PM
Screen Shot 2019-06-12 at 3 33 11 PM
Screen Shot 2019-06-12 at 3 32 32 PM

It's pretty clear that the `max_size` of 500 is only ever reached overnight, when we're not actively deploying `Shopify/shopify`. Even when it is, it's so infrequent that the 3ms garbage collection time is amortized to nearly zero.

I'm going to merge this PR to the branch, and next steps will be to merge this to `master`.

cc: @jpittis <-- You missed it!
cc: @Shopify/servcomm

PR #227, opened 2019-06-12
[DRAFT] Share circuit breakers between workers

### LINKED ISSUE #27: Share circuit breaker between workers
We should use `SysV` shared memory of `SysV` semaphores to share the state of the circuit breakers between all the workers on the host.

I want to take a crack at it this weekend

\cc @csfrancis @byroot

michaelkipper:
This PR fixes #27.

## What

Moves the implementation of circuit breakers to shared memory so circuit errors can be shared between workers on the same host.

## Why

Currently, for single-threaded workers, we need to see `error_threshold` consecutive failures in order to open a circuit. For applications where timeouts are necessarily high (e.g. 25 seconds for MySQL) that translates into up to 75 seconds of blocking behaviour when a resource is unhealthy.

If all workers on a host experience the outage simultaneously, that information can be shared and the circuit can be tripped in a single timeout iterations, increasing the speed of detecting the unhealthy resource and opening the circuit.

In addition, in the case of a variable latency outage of an upstream resource, the collective hivemind can share information about the increased latency and detect trends earlier and more reliably. This feature is not implemented yet, but becomes possible after this change.

## How

Note: This is **very much** a draft, and needs a lot of refactoring, but it's a start.

The main idea is to move the data for `Simple::Integer` and `Simple::SlidingWindow` to shared memory. The shared memory is accessed via a named key and is protected by a semaphore of the same name.

`Simple::State` simply uses the shared `Simple::Integer` as its backing store to easily share state between workers on the same host.

Feel free to take a look around, but I've still got some refactoring to do.

In particular, this implementation isn't thread-safe yet and requires some _serious_ locking.

cc: @Shopify/servcomm 
cc: @sirupsen, @csfrancis, @byroot

### DIFF
--- a/ext/semian/semian.c
+++ b/ext/semian/semian.c
@@ -66,5 +66,6 @@ void Init_semian()
   rb_define_const(cSemian, "MAX_TICKETS", INT2FIX(system_max_semaphore_count));
 
   Init_SimpleInteger();
+  Init_SlidingWindow();
   Init_CircuitBreaker();
 }
--- a/ext/semian/semian.c
+++ b/ext/semian/semian.c
@@ -64,4 +64,7 @@ void Init_semian()
 
   /* Maximum number of tickets available on this system. */
   rb_define_const(cSemian, "MAX_TICKETS", INT2FIX(system_max_semaphore_count));
+
+  Init_SimpleInteger();
+  Init_CircuitBreaker();
 }
--- a/ext/semian/simple_integer.c
+++ b/ext/semian/simple_integer.c
@@ -92,7 +92,10 @@ VALUE semian_simple_integer_value_get(VALUE self) {
 
 VALUE semian_simple_integer_value_set(VALUE self, VALUE val) {
   int *data = get_value(self);
-  *data = RB_NUM2INT(val);
+
+  // TODO(michaelkipper): Check for respond_to?(:to_i) before calling.
+  VALUE to_i = rb_funcall(val, rb_intern("to_i"), 0);
+  *data = RB_NUM2INT(to_i);
 
   return RB_INT2NUM(*data);
 }
--- a/ext/semian/simple_integer.c
+++ b/ext/semian/simple_integer.c
@@ -8,19 +8,8 @@
 
 static const rb_data_type_t semian_simple_integer_type;
 
-static const char* get_string(VALUE obj) {
-  if (RB_TYPE_P(obj, T_STRING)) {
-    return RSTRING_PTR(obj);
-  } else if (RB_TYPE_P(obj, T_SYMBOL)) {
-    return rb_id2name(SYM2ID(obj));
-  }
-
-  rb_raise(rb_eArgError, "could not convert object to string");
-  return NULL;
-}
-
 static int* get_value(VALUE self) {
-  const char* name = get_string(rb_iv_get(self, "@name"));
+  const char* name = to_s(rb_iv_get(self, "@name"));
   if (name == NULL) {
     rb_raise(rb_eArgError, "could not get object name");
   }
--- a/ext/semian/simple_integer.c
+++ b/ext/semian/simple_integer.c
@@ -1,0 +1,98 @@
+#include "simple_integer.h"
+
+#include <errno.h>
+#include <sys/ipc.h>
+#include <sys/shm.h>
+#include "types.h"
+#include "util.h"
+
+static const rb_data_type_t semian_simple_integer_type;
+
+static const char* get_string(VALUE obj) {
+  if (RB_TYPE_P(obj, T_STRING)) {
+    return RSTRING_PTR(obj);
+  } else if (RB_TYPE_P(obj, T_SYMBOL)) {
+    return rb_id2name(SYM2ID(obj));
+  }
+
+  rb_raise(rb_eArgError, "could not convert object to string");
+  return NULL;
+}
+
+static int* get_value(VALUE self) {
+  const char* name = get_string(rb_iv_get(self, "@name"));
+  if (name == NULL) {
+    rb_raise(rb_eArgError, "could not get object name");
+  }
+
+  key_t key = generate_key(name);
+
+  const int permissions = 0664;
+  int shmid = shmget(key, sizeof(int), IPC_CREAT | permissions);
+  if (shmid == -1) {
+    rb_raise(rb_eArgError, "could not create shared memory (%s)", strerror(errno));
+  }
+
+  void *val = shmat(shmid, NULL, 0);
+  if (val == (void*)-1) {
+    rb_raise(rb_eArgError, "could not get shared memory (%s)", strerror(errno));
+  }
+
+  return (int*)val;
+}
+
+void Init_SimpleInteger()
+{
+  VALUE cSemian = rb_const_get(rb_cObject, rb_intern("Semian"));
+  VALUE cSimple = rb_const_get(cSemian, rb_intern("Simple"));
+  VALUE cSimpleInteger = rb_const_get(cSimple, rb_intern("Integer"));
+
+  rb_define_alloc_func(cSimpleInteger, semian_simple_integer_alloc);
+  rb_define_method(cSimpleInteger, "initialize_simple_integer", semian_simple_integer_initialize, 0);
+  rb_define_method(cSimpleInteger, "increment", semian_simple_integer_increment, 1);
+  rb_define_method(cSimpleInteger, "reset", semian_simple_integer_reset, 0);
+  rb_define_method(cSimpleInteger, "value", semian_simple_integer_value_get, 0);
+  rb_define_method(cSimpleInteger, "value=", semian_simple_integer_value_set, 1);
+}
+
+VALUE semian_simple_integer_alloc(VALUE klass)
+{
+  semian_simple_integer_t *res;
+  VALUE obj = TypedData_Make_Struct(klass, semian_simple_integer_t, &semian_simple_integer_type, res);
+  return obj;
+}
+
+
+VALUE semian_simple_integer_initialize(VALUE self)
+{
+  int* data = get_value(self);
+  *data = 0;
+
+  return self;
+}
+
+VALUE semian_simple_integer_increment(VALUE self, VALUE val) {
+  int *data = get_value(self);
+  *data += rb_num2int(val);
+
+  return RB_INT2NUM(*data);
+}
+
+VALUE semian_simple_integer_reset(VALUE self) {
+  int *data = get_value(self);
+  *data = 0;
+
+  return RB_INT2NUM(*data);
+}
+
+VALUE semian_simple_integer_value_get(VALUE self) {
+  int *data = get_value(self);
+  return RB_INT2NUM(*data);
+}
+
+VALUE semian_simple_integer_value_set(VALUE self, VALUE val) {
+  int *data = get_value(self);
+  *data = RB_NUM2INT(val);
+
+  return RB_INT2NUM(*data);
+}
--- a/ext/semian/sliding_window.c
+++ b/ext/semian/sliding_window.c
@@ -1,0 +1,4 @@
+#include "sliding_window.h"
+
+void Init_SlidingWindow() {
+}
--- a/ext/semian/sliding_window.c
+++ b/ext/semian/sliding_window.c
@@ -1,4 +1,212 @@
 #include "sliding_window.h"
 
+#include <errno.h>
+#include <sys/ipc.h>
+#include <sys/shm.h>
+#include "util.h"
+
+static const rb_data_type_t semian_simple_sliding_window_type;
+
+static semian_simple_sliding_window_shared_t* get_window(uint64_t key) {
+  const int permissions = 0664;
+  int shmid = shmget(key, sizeof(int), IPC_CREAT | permissions);
+  if (shmid == -1) {
+    rb_raise(rb_eArgError, "could not create shared memory (%s)", strerror(errno));
+  }
+
+  void *val = shmat(shmid, NULL, 0);
+  if (val == (void*)-1) {
+    rb_raise(rb_eArgError, "could not get shared memory (%s)", strerror(errno));
+  }
+
+  return (semian_simple_sliding_window_shared_t*)val;
+}
+
+static int check_max_size_arg(VALUE max_size)
+{
+  int retval = -1;
+  switch (TYPE(max_size)) {
+  case T_NIL:
+    retval = SLIDING_WINDOW_MAX_SIZE; break;
+  case T_FLOAT:
+    rb_warn("semian sliding window max_size is a float, converting to fixnum");
+    retval = (int)(RFLOAT_VALUE(max_size)); break;
+  default:
+    retval = RB_NUM2INT(max_size); break;
+  }
+
+  if (retval <= 0) {
+    rb_raise(rb_eArgError, "max_size must be greater than zero");
+  } else if (retval > SLIDING_WINDOW_MAX_SIZE) {
+    rb_raise(rb_eArgError, "max_size must be less than %d", SLIDING_WINDOW_MAX_SIZE);
+  }
+
+  return retval;
+}
+
 void Init_SlidingWindow() {
+  printf("[DEBUG] Init_SlidingWindow\n");
+
+  VALUE cSemian = rb_const_get(rb_cObject, rb_intern("Semian"));
+  VALUE cSimple = rb_const_get(cSemian, rb_intern("Simple"));
+  VALUE cSlidingWindow = rb_const_get(cSimple, rb_intern("SlidingWindow"));
+
+  rb_define_alloc_func(cSlidingWindow, semian_simple_sliding_window_alloc);
+  rb_define_method(cSlidingWindow, "initialize_sliding_window", semian_simple_sliding_window_initialize, 2);
+  rb_define_method(cSlidingWindow, "size", semian_simple_sliding_window_size, 0);
+  rb_define_method(cSlidingWindow, "max_size", semian_simple_sliding_window_max_size, 0);
+  rb_define_method(cSlidingWindow, "values", semian_simple_sliding_window_values, 0);
+  rb_define_method(cSlidingWindow, "last", semian_simple_sliding_window_last, 0);
+  rb_define_method(cSlidingWindow, "<<", semian_simple_sliding_window_push, 1);
+  rb_define_method(cSlidingWindow, "destroy", semian_simple_sliding_window_clear, 0);
+  rb_define_method(cSlidingWindow, "reject!", semian_simple_sliding_window_reject, 0);
+}
+
+VALUE semian_simple_sliding_window_alloc(VALUE klass) {
+  semian_simple_sliding_window_t *res;
+  VALUE obj = TypedData_Make_Struct(klass, semian_simple_sliding_window_t, &semian_simple_sliding_window_type, res);
+  return obj;
+}
+
+VALUE semian_simple_sliding_window_initialize(VALUE self, VALUE name, VALUE max_size) {
+  printf("[DEBUG] semian_simple_sliding_window_initialize\n");
+
+  semian_simple_sliding_window_t *res;
+  TypedData_Get_Struct(self, semian_simple_sliding_window_t, &semian_simple_sliding_window_type, res);
+
+  const char *id_str = check_id_arg(name);
+  res->key = generate_key(id_str);
+
+  semian_simple_sliding_window_shared_t *window = get_window(res->key);
+  window->max_size = check_max_size_arg(max_size);
+  window->length = 0;
+  window->start = 0;
+  window->end = 0;
+
+  printf("[DEBUG]   key:%lu addr:0x%p max_size:%d length:%d start:%d end:%d\n", res->key, window, window->max_size, window->length, window->start, window->end);
+
+  return self;
+}
+
+VALUE semian_simple_sliding_window_size(VALUE self) {
+  printf("[DEBUG] semian_simple_sliding_window_size\n");
+
+  semian_simple_sliding_window_t *res;
+  TypedData_Get_Struct(self, semian_simple_sliding_window_t, &semian_simple_sliding_window_type, res);
+
+  semian_simple_sliding_window_shared_t *window = get_window(res->key);
+  printf("[DEBUG]   key:%lu addr:0x%p max_size:%d length:%d start:%d end:%d\n", res->key, window, window->max_size, window->length, window->start, window->end);
+
+  return RB_INT2NUM(window->length);
+}
+
+VALUE semian_simple_sliding_window_max_size(VALUE self) {
+  printf("[DEBUG] semian_simple_sliding_window_max_size\n");
+
+  semian_simple_sliding_window_t *res;
+  TypedData_Get_Struct(self, semian_simple_sliding_window_t, &semian_simple_sliding_window_type, res);
+
+  semian_simple_sliding_window_shared_t *window = get_window(res->key);
+  printf("[DEBUG]   key:%lu addr:0x%p max_size:%d length:%d start:%d end:%d\n", res->key, window, window->max_size, window->length, window->start, window->end);
+
+  return RB_INT2NUM(window->max_size);
+}
+
+VALUE semian_simple_sliding_window_values(VALUE self) {
+  printf("[DEBUG] semian_simple_sliding_window_values\n");
+
+  semian_simple_sliding_window_t *res;
+  TypedData_Get_Struct(self, semian_simple_sliding_window_t, &semian_simple_sliding_window_type, res);
+  semian_simple_sliding_window_shared_t *window = get_window(res->key);
+
+  VALUE retval = rb_ary_new_capa(window->length);
+  for (int i = 0; i < window->length; ++i) {
+    int index = (window->start + i) % window->max_size;
+    int value = window->data[index];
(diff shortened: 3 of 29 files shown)


ext/semian/simple_integer.c
```diff
@@ -0,0 +1,90 @@
...
+    rb_raise(rb_eArgError, "could not create shared memory (%s)", strerror(errno));
+  }
+
+  void *val = shmat(shmid, NULL, 0);
+  if (val == (void*)-1) {
+    rb_raise(rb_eArgError, "could not get shared memory (%s)", strerror(errno));
+  }
+
+  return (int*)val;
+}
```
thegedge: `@name` doesn't look to be changing after initialization. Could we cache the `shmid` in `semian_simple_integer_t`? How about the `void *val`?

Similarly for `circuit_breaker.c` and `sliding_window.c`.

ext/semian/simple_integer.c
```diff
@@ -0,0 +1,90 @@
+#include "simple_integer.h"
+
+#include <errno.h>
+#include <sys/ipc.h>
+#include <sys/shm.h>
+#include "types.h"
+#include "util.h"
+
+static const rb_data_type_t semian_simple_integer_type;
```
thegedge: We don't have a `free` function, so would we be leaking `shmget` memory that is no longer in use?

Similarly for `circuit_breaker.c.` and `sliding_window.c`.

ext/semian/simple_integer.c
```diff
@@ -0,0 +1,90 @@
...
+VALUE semian_simple_integer_increment(VALUE self, VALUE val) {
+  int *data = get_value(self);
+  *data += rb_num2int(val);
+
+  return RB_INT2NUM(*data);
+}
+
+VALUE semian_simple_integer_reset(VALUE self) {
+  int *data = get_value(self);
+  *data = 0;
```
thegedge: Do we have to worry about race conditions between a processes?

Similarly for `circuit_breaker.c` and `sliding_window.c`.

ext/semian/sliding_window.c
```diff
@@ -0,0 +1,212 @@
...
+    rb_warn("semian sliding window max_size is a float, converting to fixnum");
+    retval = (int)(RFLOAT_VALUE(max_size)); break;
+  default:
+    retval = RB_NUM2INT(max_size); break;
+  }
+
+  if (retval <= 0) {
+    rb_raise(rb_eArgError, "max_size must be greater than zero");
+  } else if (retval > SLIDING_WINDOW_MAX_SIZE) {
+    rb_raise(rb_eArgError, "max_size must be less than %d", SLIDING_WINDOW_MAX_SIZE);
```
thegedge: `less than or equal to %d` or `not be greater than %d`

ext/semian/sliding_window.c
```diff
@@ -0,0 +1,212 @@
...
+
+  const int index = window->end;
+  window->length++;
+  window->data[index] = RB_NUM2INT(value);
+  window->end = (window->end + 1) % window->max_size;
+
+  printf("[DEBUG]   Pushed val:%d index:%d max_size:%d length:%d start:%d end:%d\n", RB_NUM2INT(value), index, window->max_size, window->length, window->start, window->end);
+
+  return self;
+}
```
thegedge: Circular buffers 👏 🎉

ext/semian/simple_integer.c
michaelkipper: Yes, I actually addressed this in a later commit.
Basically, we want to store the `key` generated from the name on the object itself.
I'll double check I've done this everywhere.

ext/semian/simple_integer.c
michaelkipper: Yes, this is one of the outstanding issues with this PR, along with the race conditions.
@spike01 is working on the race conditions, but I don't have a solution for the memory leak right now.

ext/semian/simple_integer.c
michaelkipper: See above. @spike01 is working on that.

ext/semian/sliding_window.c
michaelkipper: Yeah, it's a little complex working with shared memory because we can't have dynamic data structures. So the window size has to be fixed, and I used a circular buffer to work around the memory management issues.

ext/semian/sliding_window.c
michaelkipper: Done.

pushrax:
If we have this, I'm fairly sure we should also decouple bulkheads from circuits. At the moment (or at least as of the end of 2018), a bulkhead acquisition failure marks an error in the same resource's circuit. There's a lengthy discussion in https://github.com/Shopify/shopify/pull/178393. I'll reproduce the big points for open source sake. At the end, my conclusion was that sharing circuits between processes and decoupling bulkheads would be an attractive option, which is possible after this PR!

@ashcharles 
> If Semian is unable to acquire a ticket due to bulkheading, an error is thrown.  Multiple errors may exceed the configured `error_threshold` for the resource causing the circuit to open. While this [behaviour is tested](https://github.com/Shopify/semian/blob/master/test/instrumentation_test.rb#L19) in Semian, document it explicitly here with a test for the case of Redis to make it clear that repeated bulkheading on Redis may lead to open-circuit errors.
> 
> I found this behaviour surprising as did others in the ensuing discussion.  I chose code > documentation. I thought that adding a test to core would make it clear for folks who aren't necessarily looking at the Semian code-base.

@jpittis 
> **TLDR: I can't think of a reason why (besides backward compatibility) that we wouldn't want to remove this behaviour from Semian.**
> 
> This is a really good test case to articulate an important property of Semian.
> 
> Bulkheads triggering does not mean a protected service is experiencing any kind of issue. But we still open a circuit and fast fail requests for the duration of error timeout.
> 
> I suspect that this has caused us to fail requests to Redis (and other protected services) when the service was not under duress.

@pushrax 
> Thanks for bringing this up. I agree it may not be desired behaviour.
> 
> The high level goal of a circuit breaker is: "if a request is likely to fail, don't issue it." This is done for two reasons. First, and most importantly, it reduces unnecessary load on the dependency, which is particularly crucial when the dependency is failing due to overload. Second, it reduces unnecessary load on the dependent, which is crucial when the dependency is timing out and taking up too much dependent capacity.
> 
> The relevant property to notice is: a circuit breaker is not useful when the cost of a failed request is similar to the cost of evaluating the circuit breaker. When a bulkhead is full, further requests will fail without hitting the network at all, thus failure is cheap. Circuit breakers aren't providing their intended improvement when applied to bulkhead failure. I agree it makes sense to remove this behaviour, it complicates thinking about tuning. An oscillating distributed bulkhead is even more confusing to reason about than a regular distributed bulkhead.

@sirupsen 
> I am not sure I am sold on removing this behaviour. I've always thought the interaction between them was useful. If tickets are 50% of workers and timeouts are high you'll be "wasting" 50% of workers on timeouts to the resource. After `3 x error_threshold` you'd be back to having 100% of capacity (if the `half_open_resource_timeout` is reasonable).
> 
> Imagine having a 25s timeout on MySQL. When all tickets are taken, clearly that MySQL is a problem and we trigger circuits rapidly. If you waited only for the circuits, you'd have to wait 3 x 25s. If you had only bulkheads, you're "wasting" 50% of capacity. If you do both, you seem to get the best of both worlds (as long as you have the half open timeout).

@pushrax
> I think this is a case where both positions are correct depending on the tuning and scenario. Simulation would be a great way to show exactly how this is true.
> 
> I keep forgetting circuits are not shared between processes, while bulkheads are. Your idea ends up working around this by sharing state between processes through the bulkheads to influence the circuits.
> 
> 6v1ebxycqeq5mpd9t2iteg_thumb_9c3
> 
> Assume t0 is the resource timeout. My earlier comment about tripping the circuit not being useful here was with reference to the overlapping time between the marked t0 interval and the breaker open interval. In that case, the bulkhead will prevent further requests anyway. In the remaining time (if any) of the breaker open interval, we avoid queuing more work. t1 is showing the "half open timeout" logic.
> 
> What about the case when t0 < timeout?
> 
> nq3bolr te6g3p2mdek0cg_thumb_9c3
> 
> > When all tickets are taken, clearly that MySQL is a problem and we trigger circuits rapidly. 
> 
> Downstream latency regression is one reason tickets can be exhausted, but upstream demand and small amounts of queuing can also cause this. If a large amount of load appears, filling tickets, we prematurely assume instability by shutting down the resource entirely. We fail requests that would not have failed.
> 
> When we trip circuits early based on bulkhead errors, we gain capacity for other work. If the other work can't use this capacity more effectively than the original work, we are making an error. This is the part that's difficult to understand: how likely is it that making capacity available for other work will have higher ROI given we just know a bulkhead is full? I don't have a good intuition for this.
> 
> Bulkheads can be tuned such that even when exhausted, there is sufficient capacity for other work, provided other competing bulkheads aren't concurrently exhausted. Is this not the way we use them? 
> 
> Would sharing circuit state between processes be a more direct way of addressing the goals here without needing to understand the statistics of production failure modes?

ext/semian/simple_integer.c
csfrancis: You can set `semian_simple_integer_type.function.dfree` to your free function:
```c
  typedef struct rb_data_type_struct rb_data_type_t;

  struct rb_data_type_struct {
          const char *wrap_struct_name;
          struct {
                  void (*dmark)(void*);
                  void (*dfree)(void*);
                  size_t (*dsize)(const void *);
                  void *reserved[2];
          } function;
          const rb_data_type_t *parent;
          void *data;
          VALUE flags;
  };
```

```
wrap_struct_name is an identifier of this instance of the struct.
It is basically used for collecting and emitting statistics.
So the identifier must be unique in the process, but doesn't need
to be valid as a C or Ruby identifier.

These dmark / dfree functions are invoked during GC execution.  No
object allocations are allowed during it, so do not allocate ruby
objects inside them.

dmark is a function to mark Ruby objects referred from your struct.
It must mark all references from your struct with rb_gc_mark or
its family if your struct keeps such references.

--
Note that it is recommended to avoid such a reference.
++

dfree is a function to free the pointer allocation.
If this is -1, the pointer will be just freed.
```

csfrancis:
My big concern with this implementation is the number of shared memory segments that will be created. It looks like we're using distinct shared memory segments for each instance of `Simple::Integer` and `Simple::SlidingWindow`. For Shopify production, I could see this amounting to (tens of?) thousands of shared memory segments. I can't say for certain that this would be a problem, but it feels a bit sketchy to me and it would be good to verify that it's not.

michaelkipper:
> My big concern with this implementation is the number of shared memory segments that will be created... For Shopify production, I could see this amounting to (tens of?) thousands of shared memory segments.

@csfrancis: The data I have for core suggests that it's hundreds, not thousands:
https://shopify.datadoghq.com/dashboard/mzg-id4-5rp/semian-lru?screenId=mzg-id4-5rp&screenName=semian-lru&tile_size=m&from_ts=1560376147986&to_ts=1560462547986&live=true&fullscreen_widget=812985505059370&fullscreen_section=overview

Our `max_size` LRU cache should put bounds on this.

ext/semian/simple_integer.c
michaelkipper: Thanks for the tip!
I updated the code to use the `dfree` functions to clean up.

ext/semian/simple_integer.c
michaelkipper: Done. Spike added binary sempahore locks everywhere.

ext/semian/sliding_window.c
```diff
@@ -0,0 +1,458 @@
...
+      }
+    }
+  }
+  sem_meta_unlock(res->sem_id);
+
+  return self;
+}
+
+VALUE
+semian_simple_sliding_window_push(VALUE self, VALUE value)
```
sirupsen: Many other circuits just use a number and not a sliding window. What are the advantages of either approach? One huge advantage would be that you could just use the Integer class instead of a 450 line sliding window implementation.

README.md
```diff
@@ -448,6 +448,31 @@ There are three configuration parameters for circuit breakers in Semian:
...
+the total number of errors we'd need to see server-wide is still _E_. In
+this configuration, we can reduce the time-to-open for a circuit from _E * T_
+to simply _T_ (provided that _N_ is greater than _E_).
+
+You should run a simulation with your workloads to determine an efficient
+scaling factor that will produce a time-to-open reduction but isn't too
+sensitive.
+
+To disable host-based circuits, set the environment variable
+`SEMIAN_CIRCUIT_BREAKER_IMPL` to `ruby`.
```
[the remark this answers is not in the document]
michaelkipper: Done.

ext/semian/semian.c
```diff
@@ -64,4 +66,30 @@ void Init_semian()
...
+static int
+use_c_circuits() {
+  char *circuit_impl = getenv("SEMIAN_CIRCUIT_BREAKER_IMPL");
+  if (circuit_impl == NULL) {
+    fprintf(stderr, "Warning: Defaulting to worker-based circuit breaker implementation\n");
+    return 0;
+  } else {
+    if (!strcmp(circuit_impl, "ruby") || !strcmp(circuit_impl, "worker")) {
+      return 0;
+    } else if (!strcmp(circuit_impl, "c") || !strcmp(circuit_impl, "host")) {
```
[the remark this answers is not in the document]
michaelkipper: Done.

test/simple_integer_test.rb
```diff
@@ -41,6 +42,24 @@ def test_reset
       @integer.reset
       assert_equal(0, @integer.value)
     end
+
+    if ENV['SEMIAN_CIRCUIT_BREAKER_IMPL'] != 'ruby'
```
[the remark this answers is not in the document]
michaelkipper: Done. This test can run for both implementations.

ext/semian/semian.c
```diff
@@ -64,4 +66,30 @@ void Init_semian()
 
   /* Maximum number of tickets available on this system. */
   rb_define_const(cSemian, "MAX_TICKETS", INT2FIX(system_max_semaphore_count));
+
+  if (use_c_circuits()) {
+    Init_SimpleInteger();
```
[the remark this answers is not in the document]
michaelkipper: Mostly because that's roughly the way the bulkhead was implemented, and I wanted to keep the codebase familiar. One could argue that the Ruby implementation of the bulkhead is actually just a stub, but then those functions should have `NotImplemented` assertions in them.

ext/semian/simple_integer.c
```diff
@@ -0,0 +1,157 @@
...
+VALUE
+semian_simple_integer_initialize(VALUE self, VALUE name)
+{
+  semian_simple_integer_t *res;
+  TypedData_Get_Struct(self, semian_simple_integer_t, &semian_simple_integer_type, res);
+  res->key = generate_key(to_s(name));
+
+  dprintf("Initializing simple integer '%s' (key: %lu)", to_s(name), res->key);
+  res->sem_id = initialize_single_semaphore(res->key, SEM_DEFAULT_PERMISSIONS);
+  res->shmem = get_or_create_shared_memory(res->key, &init_fn);
```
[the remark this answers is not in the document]
michaelkipper: There was a brief period where there was something else in the shared memory (I can't really remember now). There's no real issue with using a semaphore, and it would get rid of the need for locks, assuming we could guarantee that the operations would not block.

ext/semian/sliding_window.c
michaelkipper: To do circuit breaking without a sliding window, you'd have to use fixed time intervals for error counts, and reset the counter at the end of each fixed-time interval. There are several issues with this, off the top of my head:

  * That's a breaking change, compared to the current implementation.
  * You don't open circuits as fast, if a series of failure straddles a reset boundary. There can be cases when the circuit never opens at all (which you could argue is a benefit).
  * A master/slave election might be necessary to determine which worker resets the counter, which could be more complex than the sliding window implementation.

This approach is admittedly more complex than a single integer counter but I'm not sure that it's worth sacrificing the benefits of a sliding window.

It's hard to reason about the effect of this without millisecond-accurate production data.

ext/semian/semian.c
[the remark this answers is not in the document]
sirupsen: Fine with me. 👍

ext/semian/simple_integer.c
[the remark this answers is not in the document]
sirupsen: I'll think more about this when I review the actual code and not just the approach.

ext/semian/sliding_window.c
sirupsen: Just to ensure we're on the same page, the alternative approach I'm considering is that you have two integers: (1) Number of failures, (2) Timestamp of when to reset the number of failures.  This makes the code significantly simpler (in Ruby, it's a few lines, but it's quite a bit more here). 

I changed this a few years ago in https://github.com/Shopify/semian/pull/24. It's actually addressing a separate point, which was false positives. I think the reason why you didn't arrive at that conclusion was that the previous solution would refresh the TTL every time it was modified.

Overall, likely the way we should look at circuits (I believe @hkdsun suggested this? Or maybe @csfrancis? Not my idea) is to talk about an "error percentage", i.e. if 10% of requests were errors in some window of time, open the circuits.

michaelkipper: @pushrax suggested this in our shared doc as well.

My primary motivation for host-based circuits is to address the variable latency problem, where a single worker doesn't see error rates above a threshold but the superposition of all errors on a host shows an anomaly.

I'm not sure it works without some sort of window, because in the implementation before #24, the previous errors would only be purged when there was a period of time since the last observed error with no new errors occurring. If the rate of new errors is greater than the `error_timeout`, then the pseudo-window would grow without bound and the circuit would eventually open. So if I'm reading that code correctly, the circuit opens quickly when there are a lot of errors, slowly-but-surely when there are some errors, and only remains closed when there are fewer than one error per `error_timeout`.

Using that sort of approach in a host-based circuit implementation likely regresses to the slowly-but-surely case, where the circuit constantly opens at a low frequency. Combined with capacity loss during the `half_open_resource_timeout` period, this would be an overall drop in available capacity.

michaelkipper:
@csfrancis has some legitimate concerns about the number of shared memory segments this PR would create. `SHMMNI` is 4,096 on Linux so with our LRU `max_size` of 500, we'd need 1,500 of them.

Scott suggested storing all the sliding windows in a single shared memory segment, but I have concerns with this approach. Specifically, complexity that would be required to manage the indexing into that data structure. Right now, we generate a key and the IPC subsystem handles the lookup.

Another difficulty is cleanup. Given that we set `IPC_RMID` and call `shmdt` when a sliding window goes out of scope, we can defer cleanup to the kernel. If we had a single memory segment, we'd have to manage all that ourselves.

For `SimpleInteger`, the size of a shared memory segment on Linux is at least one page (4kB) so storing `SimpleInteger` in a shared memory segment is extremely inefficient. `SEMMNI` is 32,000 on Linux, so storing `SimpleIntegers` as a semaphore is a reasonable choice (as @sirupsen questioned earlier). I'll try and implement that ASAP.

michaelkipper:
@csfrancis was concerned last week that the host-based circuits PR would require too many shared memory segments. The default on Linux is 4,096 and we were going to be using somewhere in the neightborhood of 1,500 with a `max_size` LRU of 500. He suggested using a single shared memory segment for all the circuit data.

I took a long stab at it, but ultimately concluded that it was too hard to do garbage collection. With semaphores, it's easy to do garbage collection with `SEM_UNDO` (in SysV IPC). With shared memory, we can set `IPC_RMID` to have the segment be destroyed after the last process detaches from it. But with a single segment for all the circuits, we'd have to manage garbage collection within the segment ourselves. I don't think it's impossible, but it adds a large amount of complexity to an already complex PR.

To alleviate concerns about the number of shared memory segments, I converted the `Simple::Integer` to use a semaphore, since there are 32k semaphore sets available, which leaves the number of shared memory segments required at 500 since only the SlidingWindow is using them.

As far as shipping this is concerned, I have https://github.com/Shopify/shopify/pull/206065 to bump core to use the new branch, with the previous behaviour. Then https://github.com/Shopify/semian/pull/243 adds support for an environment variable `SEMIAN_CIRCUIT_BREAKER_FORCE_HOST` which enables host-based circuits based on machine name (so we can enable it on a single node). Once we validate the behaviour on that node, we can move to a percentage rollout.

ext/semian/simple_integer.c
```diff
@@ -0,0 +1,155 @@
...
+  semian_simple_integer_t *res;
+  TypedData_Get_Struct(self, semian_simple_integer_t, &semian_simple_integer_type, res);
+
+  // This is definitely the worst API ever.
+  // https://silverhammermba.github.io/emberb/c/#parsing-arguments
+  VALUE val;
+  rb_scan_args(argc, argv, "01", &val);
+
+  int value = check_increment_arg(val);
+  if (perform_semop(res->sem_id, 0, value, SEM_UNDO, NULL) == -1) {
```
csfrancis: I don't think you want `SEM_UNDO` here?

ext/semian/types.h
```diff
@@ -38,4 +40,28 @@ typedef struct {
...
+  int sem_id;
+} semian_simple_integer_t;
+
+// Shared simple sliding window structure
+typedef struct {
+  int max_size;
+  int length;
+  int start;
+  int data[SLIDING_WINDOW_MAX_SIZE];
+} semian_simple_sliding_window_shared_t;
```
csfrancis: The size of this structure is larger than `SHM_DEFAULT_SIZE` (and also larger than the default `PAGE_SIZE` of 4k).

ext/semian/sliding_window.c
```diff
@@ -0,0 +1,485 @@
...
+    res->shmem->start = 0;
+  }
+  sem_meta_unlock(res->sem_id);
+
+  return self;
+}
+
+// Handy for debugging the sliding window, but too noisy for regular debugging.
+/*
+static void dprint_window(semian_simple_sliding_window_shared_t *window)
```
csfrancis: Maybe add an `#ifdef` around this or something so you don't have to leave the code commented out?

ext/semian/sliding_window.c
```diff
@@ -0,0 +1,485 @@
...
+  window->max_size = new_max_size;
+  window->length = new_length;
+
+  return RB_INT2NUM(new_max_size);
+}
+
+static VALUE
+resize_window(int sem_id, semian_simple_sliding_window_shared_t* window, int new_max_size)
+{
+  if (new_max_size > SLIDING_WINDOW_MAX_SIZE) return Qnil;
```
csfrancis: Should this be raising?

ext/semian/sliding_window.c
```diff
@@ -0,0 +1,485 @@
...
+    rb_raise(rb_eArgError, "scale_factor cannot be greater than 1.0");
+  }
+
+  return retval;
+}
+
+static VALUE
+grow_window(int sem_id, semian_simple_sliding_window_shared_t* window, int new_max_size)
+{
+  if (new_max_size > SLIDING_WINDOW_MAX_SIZE) return Qnil;
```
csfrancis: Raise here?

ext/semian/sysv_semaphores.c
```diff
@@ -218,28 +204,22 @@ initialize_new_semaphore_values(int sem_id, long permissions)
...
-  int i;
-  int sem_id = -1;
-  union semun sem_opts;
   struct semid_ds sem_ds;
-
+  union semun sem_opts;
   sem_opts.buf = &sem_ds;
-  sem_id = semget(key, 1, permissions);
 
+  int sem_id = semget(key, 1, permissions);
```
csfrancis: Does this produce a warning? I thought that `key_t` was a 32-bit value and you've changed the type of 64 bits.

ext/semian/simple_integer.c
michaelkipper: No, I don't think I do.

ext/semian/types.h
michaelkipper: Good point. I changed the maximum size to `1,000` (which is plenty) and wrote a test to ensure that the entire shared memory structure is less than or equal to a page.

ext/semian/sliding_window.c
michaelkipper: I don't see the difference, because the compiler won't see it anyways, but sure.

ext/semian/sliding_window.c
michaelkipper: Done.

ext/semian/sliding_window.c
michaelkipper: Done.

ext/semian/sysv_semaphores.c
michaelkipper: I didn't see one, and I compile with `-Wall`.
All the values are generated with:

```c
key_t generate_key(const char *name);
```

I was following what was done in `semian_resource_t`.

ext/semian/sysv_shared_memory.c
```diff
@@ -0,0 +1,54 @@
...
+
+#define TIMEOUT_MS (5 * 1e6)
+#define WAIT_MS (10)
+#define RETRIES (TIMEOUT_MS / WAIT_MS)
+
+void*
+wait_for_shared_memory(uint64_t key)
+{
+  for (int i = 0; i < RETRIES; ++i) {
+    int shmid = shmget(key, SHM_DEFAULT_SIZE, SHM_DEFAULT_PERMISSIONS);
```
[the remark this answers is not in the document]
michaelkipper: The first try uses `IPC_CREAT | IPC_EXCL`. If that fails, then we assume it's because another process has created the segment. I think my intention was to check if that other process had run the `shared_memory_init_fn` but I've tried to optimize that out by building data structures that reset to zero'ed out memory.

I removed the wait loop - we can add that feature if we need it.

PR #229, opened 2019-06-13
WIP: Locks

spike01:
This only puts locking around SimpleInteger, would like some 👀 before continuing because I'm not super confident about the approach. If it's ok then I can add locks to the circuit breaker and sliding window.

Ideally, the logic for semaphore set creation shouldn't be repeated, and I'm wondering if I should be pushing the locking down a layer. Don't know what 'the proper C way' is!

(this also includes #228)

### DIFF
--- a/ext/semian/semian.c
+++ b/ext/semian/semian.c
@@ -65,7 +65,9 @@ void Init_semian()
   /* Maximum number of tickets available on this system. */
   rb_define_const(cSemian, "MAX_TICKETS", INT2FIX(system_max_semaphore_count));
 
-  Init_SimpleInteger();
-  Init_SlidingWindow();
-  Init_CircuitBreaker();
+  if (getenv("USE_RUBY_CIRCUITS") == NULL || strcmp(getenv("USE_RUBY_CIRCUITS"), "false")) {
+    Init_SimpleInteger();
+    Init_SlidingWindow();
+    Init_CircuitBreaker();
+  }
 }
--- a/ext/semian/semian.c
+++ b/ext/semian/semian.c
@@ -1,5 +1,7 @@
 #include "semian.h"
 
+static int use_c_circuits();
+
 void Init_semian()
 {
   VALUE cSemian, cResource;
@@ -65,9 +67,19 @@ void Init_semian()
   /* Maximum number of tickets available on this system. */
   rb_define_const(cSemian, "MAX_TICKETS", INT2FIX(system_max_semaphore_count));
 
-  if (getenv("USE_RUBY_CIRCUITS") == NULL || strcmp(getenv("USE_RUBY_CIRCUITS"), "false")) {
+  if (use_c_circuits()) {
     Init_SimpleInteger();
     Init_SlidingWindow();
     Init_CircuitBreaker();
+  }
+}
+
+static int
+use_c_circuits() {
+  char *circuit_impl = getenv("CIRCUIT_IMPL");
+  if (circuit_impl == NULL || strcmp(circuit_impl, "ruby") != 0) {
+    return 1;
+  } else {
+    return 0;
   }
 }
--- a/ext/semian/semian.c
+++ b/ext/semian/semian.c
@@ -76,7 +76,7 @@ void Init_semian()
 
 static int
 use_c_circuits() {
-  char *circuit_impl = getenv("CIRCUIT_IMPL");
+  char *circuit_impl = getenv("SEMIAN_CIRCUIT_BREAKER_IMPL");
   if (circuit_impl == NULL || strcmp(circuit_impl, "ruby") != 0) {
     return 1;
   } else {
--- a/ext/semian/simple_integer.c
+++ b/ext/semian/simple_integer.c
@@ -28,6 +28,8 @@ static semian_simple_integer_shared_t* get_value(VALUE self) {
 
 void Init_SimpleInteger()
 {
+  printf("[DEBUG] Init_SimpleInteger\n");
+
   VALUE cSemian = rb_const_get(rb_cObject, rb_intern("Semian"));
   VALUE cSimple = rb_const_get(cSemian, rb_intern("Simple"));
   VALUE cSimpleInteger = rb_const_get(cSimple, rb_intern("Integer"));
--- a/ext/semian/simple_integer.c
+++ b/ext/semian/simple_integer.c
@@ -1,28 +1,38 @@
 #include "simple_integer.h"
 
-#include <errno.h>
-#include <sys/ipc.h>
-#include <sys/shm.h>
+#include "sysv_semaphores.h"
+#include "sysv_shared_memory.h"
 #include "types.h"
 #include "util.h"
-#include "sysv_semaphores.h"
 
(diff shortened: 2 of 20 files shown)


ext/semian/semian.c
```diff
@@ -65,7 +65,9 @@ void Init_semian()
   /* Maximum number of tickets available on this system. */
   rb_define_const(cSemian, "MAX_TICKETS", INT2FIX(system_max_semaphore_count));
 
-  Init_SimpleInteger();
-  Init_SlidingWindow();
-  Init_CircuitBreaker();
+  if (getenv("USE_RUBY_CIRCUITS") == NULL || strcmp(getenv("USE_RUBY_CIRCUITS"), "false")) {
```
michaelkipper: Probably should hide this in a function. Might get more complicated than `false` (e.g. `0`, `disabled`, etc.)

ext/semian/simple_integer.c
```diff
@@ -5,13 +5,11 @@
...
 #include "util.h"
+#include "sysv_semaphores.h"
 
 static const rb_data_type_t semian_simple_integer_type;
 
-static semian_simple_integer_shared_t* get_value(VALUE self) {
-  semian_simple_integer_t *res;
-  TypedData_Get_Struct(self, semian_simple_integer_t, &semian_simple_integer_type, res);
-
+static semian_simple_integer_shared_t* get_value(semian_simple_integer_t* res) {
```
michaelkipper: This is a great example of where C++ function overloading makes a lot of sense... :(

ext/semian/simple_integer.c
```diff
@@ -54,9 +54,12 @@ VALUE semian_simple_integer_initialize(VALUE self, VALUE name)
   TypedData_Get_Struct(self, semian_simple_integer_t, &semian_simple_integer_type, res);
   res->key = generate_key(to_s(name));
 
-  semian_simple_integer_shared_t* data = get_value(self);
+  semian_simple_integer_shared_t* data = get_value(res);
   data->val = 0;
 
+  const int permissions = 0664;
```
michaelkipper: We should probably move these permissions out somewhere...

ext/semian/simple_integer.c
```diff
@@ -66,35 +69,68 @@ VALUE semian_simple_integer_increment(int argc, VALUE *argv, VALUE self) {
   VALUE val;
   rb_scan_args(argc, argv, "01", &val);
 
-  semian_simple_integer_shared_t *data = get_value(self);
+  semian_simple_integer_t *res;
+  TypedData_Get_Struct(self, semian_simple_integer_t, &semian_simple_integer_type, res);
+
+  sem_meta_lock(res->sem_id);
```
michaelkipper: It might be helpful visually to create a new scope here. E.g: 

```
sem_meta_lock(res->sem_id);
{
  // Do work in here
}
sem_meta_unlock();
```

It doesn't change anything functionally, but it makes it clear what the critical section is.
The challenge is that you need to pre-declare variables.

test/simple_integer_test.rb
```diff
@@ -41,6 +41,26 @@ def test_reset
...
+        process_count = 255
+        100.times do
+          process_count.times do
+            fork do
+              value = @integer.increment(1)
+              exit!(value)
+            end
+          end
+          exit_codes = Process.waitall.map { |_, status| status.exitstatus }
+          # No two processes should exit with the same exit code
```
michaelkipper: Neat approach!

ext/semian/sysv_semaphores.c
```diff
@@ -240,3 +241,38 @@ diff_timespec_ms(struct timespec *end, struct timespec *begin)
   long begin_ms = (begin->tv_sec * 1e3) + (begin->tv_nsec / 1e6);
   return end_ms - begin_ms;
 }
+
+int
+initialize_single_semaphore(key_t key, long permissions)
+{
+
```
michaelkipper: Nit: Blank line

ext/semian/simple_integer.c
spike01: TIL!

ext/semian/simple_integer.c
spike01: I've pulled this out into the header for now, looking a bit more closely I wonder if it makes more sense to have this follow whatever permissions are passed from when the resource is created... and also manage the destruction of circuit breakers/sliding windows/simple integers there?

ext/semian/simple_integer.c
```diff
@@ -65,69 +65,77 @@ VALUE semian_simple_integer_increment(int argc, VALUE *argv, VALUE self) {
   // This is definitely the worst API ever.
   // https://silverhammermba.github.io/emberb/c/#parsing-arguments
   VALUE val;
+  semian_simple_integer_t *res;
+  VALUE rb_val;
```
michaelkipper: Convention is usually to use `retval` for the name here, to clarify that it's the thing that will ultimately be returned.

ext/semian/semian.c
```diff
@@ -65,7 +67,19 @@ void Init_semian()
...
+  if (use_c_circuits()) {
+    Init_SimpleInteger();
+    Init_SlidingWindow();
+    Init_CircuitBreaker();
+  }
+}
+
+static int
+use_c_circuits() {
+  char *circuit_impl = getenv("CIRCUIT_IMPL");
```
michaelkipper: Nit: Since Semian is a library, running in the context of a larger program, I think it makes sense to namespace this variable. Perhaps something like `SEMIAN_CIRCUIT_BREAKER_IMPL`?

ext/semian/sliding_window.c
```diff
@@ -164,27 +196,33 @@ VALUE semian_simple_sliding_window_reject(VALUE self) {
...
+
+    int cleared = 0;
+    for (int i = 0; i < length; ++i) {
+      int index = (start + i) % length;
+      int value = window->data[index];
+      dprintf("  i:%d index: %d value:%d max_size:%d length:%d start:%d end:%d", i, index, value, window->max_size, window->length, window->start, window->end);
+      VALUE y = rb_yield(RB_INT2NUM(value));
+      if (RTEST(y)) {
+        if (cleared++ != i) {
+          sem_meta_unlock(res->sem_id);
```
spike01: @michaelkipper It took me a while to figure this one out, is this an ok place to release the lock in the event of an exception? `rb_raise` means that the loop exits so `sem_meta_unlock` doesn't get called unless it's here.

michaelkipper: If it happens enough, it might be worth defining:
```c
void raise_arg_error(uint64_t sem_id, msg, ...) {
  if (sem_id) sem_meta_unlock(sem_id);
  rb_raise(rb_eArgError, msg, __VA_ARGS__); // <-- can't remember the syntax of this...
}
```
and then calling it like:
```c
if (RTEST(y)) {
  if (cleared++ != i) {
    raise_arg_error(res->sem_id, "something...");
  }
}
```

michaelkipper:
Merging this into `mkipper/global-circuit-breakers` so I can continue working on per-worker scaling factors.

PR #230, opened 2019-06-14
'Temporary failure in name resolution' is a DNS lookup failure

pushrax:
Ran into an issue today where DNS was down, and the message that we got was "Temporary failure in name resolution" (which is EAGAIN on getaddrinfo). This wasn't opening circuits, leading to cascading failure.

cc @inf-rno @stefanmb @daniellaniyo @csfrancis @hkdsun

### DIFF
--- a/lib/semian/redis.rb
+++ b/lib/semian/redis.rb
@@ -85,7 +85,7 @@ module Semian
         begin
           raw_connect
         rescue SocketError, RuntimeError => e
-          raise ResolveError.new(semian_identifier) if (e.cause || e).to_s =~ /(can't resolve)|(name or service not known)|(nodename nor servname provided, or not known)/i
+          raise ResolveError.new(semian_identifier) if dns_resolve_failure?(e.cause || e)
           raise
         end
       end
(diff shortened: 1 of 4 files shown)


stefanmb:
```

File: lib/semian/redis.rb [hunk 1]
```diff
@@ -111,6 +111,10 @@ def raise_if_out_of_memory(reply)
       return unless reply.message =~ /OOM command not allowed when used memory > 'maxmemory'\.\s*\z/
       raise ::Redis::OutOfMemoryError.new(reply.message)
     end
+
+    def dns_resolve_failure?(e)
+      e.to_s.match?(/(can't resolve)|(name or service not known)|(nodename nor servname provided, or not known)|(failure in name resolution)/i)
```
pushrax: Note, #match? is a 2.4 feature, but 2.4 is from 2016 and 2.3 is end-of-life now.

tylermercier:
Nice fix!

pushrax:
CI is super flaky, I dug a bit and think it's significantly due to eventmachine (via thin). @paranoidaditya is working on replacing that with a different server.

Merging this now since master CI is already broken, all new tests pass and this PR does not introduce new test failures.

sirupsen:
Thanks for the quick fix! 👍 

What happens on MySQL when this happens, or with Net/HTTP? Is that covered?

PR #231, opened 2019-06-15
Use a TCPServer instead of Thin

paranoidaditya:
## What

Use WEBrick instead of https://github.com/macournoyer/thin/ in tests. 

## Why 

Thin relied on eventmachine, which had segfaults pretty often in tests. 
The bug in eventmachine has been fixed in https://github.com/eventmachine/eventmachine/pull/835 but there has not been a new release in a long time 

CI is still flaky, the tests pass locally:
```
# Running:

.....................................................................................................................................................................................

Finished in 87.918143s, 2.0587 runs/s, 5.0274 assertions/s.

```

Merge after #193


.travis.yml
```diff
@@ -6,9 +6,9 @@ before_install:
   - scripts/install_toxiproxy.sh
 
 rvm:
-  - '2.4'
```
michaelkipper: Ruby 2.4 is still in security maintenance. Do we want to remove it from this open-source project?
https://www.ruby-lang.org/en/downloads/branches/

paranoidaditya: Added it back!

paranoidaditya:
It seems like WEBrick is too slow to startup/shutdown, and the tests here can probably make do with a just a TCP server.
Exploring..

paranoidaditya:
Before https://github.com/Shopify/semian/pull/231/commits/1406238259d08a0c22c20057e9c73bf03b3f7a4b:

```
bundle exec ruby -Itest test/net_http_test.rb
Forked test server 26911
Forked test server 26914
Run options: --seed 51025

# Running:

.......................

Finished in 11.892186s, 1.9340 runs/s, 6.8112 assertions/s.

23 runs, 81 assertions, 0 failures, 0 errors, 0 skips
Killed test server 26911.
Killed test server 26914.
```

After:
```
bundle exec ruby -Itest test/net_http_test.rb
Forked test server 28808
Forked test server 28811
Run options: --seed 14284

# Running:

.......................

Finished in 0.005257s, 4375.3353 runs/s, 951.1598 assertions/s.

23 runs, 5 assertions, 0 failures, 0 errors, 0 skips
Killed test server 28808.
Killed test server 28811.
``` 

🤯🤯🤯

Gemfile
```diff
@@ -1,10 +1,6 @@
 source 'https://rubygems.org'
 gemspec
 
-group :debug do
-  gem 'byebug'
```
michaelkipper: This seems generally useful, no?

ext/semian/sysv_semaphores.c
```diff
@@ -107,11 +107,11 @@ perform_semop(int sem_id, short index, short op, short flags, struct timespec *t
   buf.sem_op  = op;
   buf.sem_flg = flags;
 
-  int num_retries = 3;
+  // int num_retries = 3;
```
michaelkipper: Either put this back, or remove the comment.

Gemfile
michaelkipper: Edit: Unless it's replaced with the development dependency below.

test/helpers/mock_server.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+require 'webrick'
+
+module MockServer
+  extend self
+
+  def start(port)
+    pid = Thread.new do
```
michaelkipper: Nit: Technically this is a `tid`, not a `pid`.

test/helpers/mock_server.rb
```diff
@@ -0,0 +1,53 @@
...
+      TestServer.new(hostname: '0.0.0.0', port: port).start
+    end
+    sleep(1)
```
michaelkipper: Sleep statements are gross. Is there another way to check for server health?

test/helpers/mock_server.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def cleanup(pid)
+    Thread.kill(pid)
+    puts "Killed test server #{pid}."
+  end
+end
+
+class TestServer
+  def initialize(hostname:, port:)
+    dev_null = WEBrick::Log.new("/dev/null", 7)
```
michaelkipper: Is there a way to put this into an artifact directory, so it's available from Travis?
https://docs.travis-ci.com/user/uploading-artifacts/

test/helpers/mock_server.rb
paranoidaditya: There isn't a helper function to check server health in WEBrick, so had to put back the old `poll_until_ready`.

test/helpers/mock_server.rb
paranoidaditya: Seems like we'll need to bring in our on S3 bucket for that

thegedge: Any reason to not have this output to stderr (i.e., use `nil` for the filename)? I'm assuming those logs show up in the Travis UI.

test/helpers/mock_server.rb
```diff
@@ -0,0 +1,66 @@
...
+    status, content_type, body = process(request)
+    response.status = status
+    response['Content-Type'] = content_type
+    response.body = body
+  end
+
+  def process(env)
+    response_code = env.path.delete("/")
+    response_code = '200' if response_code == ""
+    [response_code, 'text/html', 'Success']
```
thegedge: nit: I know this is copied from previous code, but technically just "Success" isn't a valid `text/html` document, at least not in modern HTML. Is just `text/plain` sufficient?

test/test_helper.rb
```diff
@@ -48,6 +49,16 @@
   },
 ])
 
+tids = []
+tids << MockServer.start(SemianConfig['http_port'])
+tids << MockServer.start(SemianConfig['http_port'] + 100)
```
thegedge: What is this 100? Avoid magic numbers! :)

test/net_http_test.rb
```diff
@@ -511,42 +492,15 @@ def close_circuit!(hostname: SemianConfig['toxiproxy_upstream_host'], toxic_port
...
-          poll_until_ready(hostname: hostname, port: port)
-        rescue RuntimeError
-          server_thread.kill
-          server_thread.join if server_threw_error
-          raise
-        end
-
-        assert(server.running?)
-        reset_semian_resource(hostname: SemianConfig['toxiproxy_upstream_host'], port: port) if reset_semian_state
+      if port == SemianConfig['http_port'] || SemianConfig['http_port'] + 100 # rubocop:disable all
```
thegedge: Similar to another comment: what is this 100? We should avoid magic numbers, putting them into something that is named (constant, function, etc).

test/net_http_test.rb
```diff
@@ -511,42 +492,15 @@ def close_circuit!(hostname: SemianConfig['toxiproxy_upstream_host'], toxic_port
...
+        reset_semian_resource(port: port) if reset_semian_state
         @proxy = Toxiproxy[:semian_test_net_http]
-        yield(hostname, port.to_i)
-      ensure
-        server&.stop
-        server_thread.kill
-        poll_until_gone(hostname: hostname, port: port)
+        yield('0.0.0.0', port.to_i)
```
thegedge: This is technically an implicit coupling to the `0.0.0.0` value used in `TestServer.start`. I'd recommend calling this out by putting it in a constant.

Arguably, though, it's not super useful to yield. `0.0.0.0` basically says that any local address that can route to this process will be accepted. I would think the more useful thing to yield would be an address that can be used to talk to the server (e.g., `127.0.0.1` in this case).

michaelkipper: I'll let @paranoidaditya comment on this, but in my experience with Docker, `0.0.0.0` is generally preferred to `127.0.0.1`.
But I agree that it should be moved into a named constant.

thegedge: For binding, or making a request?

I'll clarify my comment too: if the intent is to actually yield the address we're bound to, sticking with `0.0.0.0` is a-o-k. If we'd like to yield an address that routes to the server referenced by `with_server`, `0.0.0.0` would go through the "default" route, which I doubt we can guarantee would route to the server (but it probably generally does).

test/helpers/mock_server.rb
paranoidaditya: Just wanted to avoid lots of logs in the CI, primarily the reason why I went with no logs.
Took the silent logging out in the recent commit

test/net_http_test.rb
paranoidaditya: Changed it get the value from the yaml, this was copied from previous code

test/helpers/mock_server.rb
thegedge: If there's way too many logs, we could always have our own custom logger that logs only the interesting things like errors :)

Up to you. If you think there's way too many log lines, feel free to keep it `/dev/null`. Maybe just the access log is `/dev/null`? I'll leave all of that to your discretion.

test/net_http_test.rb
```diff
@@ -511,42 +492,15 @@ def close_circuit!(hostname: SemianConfig['toxiproxy_upstream_host'], toxic_port
...
-          poll_until_ready(hostname: hostname, port: port)
-        rescue RuntimeError
-          server_thread.kill
-          server_thread.join if server_threw_error
-          raise
-        end
-
-        assert(server.running?)
-        reset_semian_resource(hostname: SemianConfig['toxiproxy_upstream_host'], port: port) if reset_semian_state
+      if port == SemianConfig['http_port'] || SemianConfig['http_port_2'] # rubocop:disable all
```
thegedge: Previously, we'd reset regardless of port. Now we only reset if it's one of these two ports. Was this necessary to get things working?

My suspicion is that this is due to the fact that we only know, with certainty, that these are the two ports started. If that's the case, again we're implicitly coupled to other code (specifically, the code that starts the servers in `test/test_helper.rb`). Is there any way we can prevent this coupling?

paranoidaditya: Previously we would start and shutdown a server on every call to `with_server`. 
And also reset the semians on them. 

These were the only two ports we'd start a server on, so now instead of spinning up a new server I just spin them beforehand and only reset the semians in `with_server`.

paranoidaditya: I could make the `http_port` in the yaml an array, spin up server for each in `test/test_helper.rb` and do check for the ports here. 

Arguably we can get rid of this if clause and reset semian resources if we make  `http_port` in the yaml and array

ext/semian/sysv_semaphores.c
```diff
@@ -107,11 +107,9 @@ perform_semop(int sem_id, short index, short op, short flags, struct timespec *t
   buf.sem_op  = op;
   buf.sem_flg = flags;
 
-  int num_retries = 3;
-
   do {
     result = semtimedop(sem_id, &buf, 1, ts);
-  } while (result < 0 && errno == EINTR && num_retries-- > 0);
+  } while (result < 0 && errno == EINTR);
```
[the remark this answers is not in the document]
paranoidaditya: `semtimedop` when interrupted by signal, does not get restarted automatically, Initially in https://github.com/Shopify/semian/pull/233
 I set the retry count to `3` but that only (partially) fixed `Semian::InternalError: error setting ticket count, errno: 4 (Interrupted system call)` in CI (VM), it was still happening but a lot less

Running it locally in docker would work without any errors but running docker in CI would again result in system calls being interrupted.

test/net_http_test.rb
paranoidaditya: It does just work if I take out the `if` clause.

ext/semian/sysv_semaphores.c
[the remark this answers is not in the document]
paranoidaditya: in theory it would be fine? unless each syscall is constantly getting interrupted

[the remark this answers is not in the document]
pushrax: As we discussed it still scares me to introduce the possibility of an infinite loop in hard-to-debug C extensions. If there needs to be a higher limit, we should just increase the limit IMO.

I'd very much like to know the source of these signals, given the high number of occurrences. In GUI environments, a common cause is SIGWINCH from resizing a window, however that's not possible here. It could be a profiler attached that's using SIGALRM on a timer, but we don't run stackprof in this project. Could attach a signal handler to every signal and log.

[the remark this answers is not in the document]
thegedge: I guess it should be fine, since I can't think of a situation where result would forever be less than 0 with errno also remaining `EINTR`.

That being said, I have two issues here. First, we seem to be solving a CI problem, not a production problem. Second, but related to the first, is that this feels like the wrong place to solve this. If there were a signal, we completely suppress it. Perhaps the application would want to handle the signal differently? I feel like this logic should exist at the Ruby level with the ability to be configured by the application that's using Semian. We can default it to "forever retry". Something like this, for example:

```ruby
Semian.interrupt_handler = -> { |signal| Signal.signame(signal) != "TERM" }
```

If the result of the block is `true`, it retries the `semtimedop` call, otherwise it fails. Another option would be to have something like `Semian.interrupt_retry_count`. We could set it to 3 for production, and some really high value for CI.

What does everyone else think?

test/config/hosts.yml
```diff
@@ -11,6 +11,7 @@ redis_toxiproxy_port: 16379
 
 http_host: 127.0.0.1
 http_port: 31050
+http_port_2: 31150
```
pushrax: Can you think of a more descriptive name?

ext/semian/sysv_semaphores.c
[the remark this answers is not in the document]
michaelkipper: I'm OK leaving it at 3 for prod/CI.
If CI requires a much larger number, then we should revisit (exponential backoff?).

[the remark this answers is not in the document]
pushrax: Retrying on EINTR is correct and you could consider it a bug that we were not doing so. This does not suppress any signal handlers. EINTR is an implementation detail of signals. This is my understanding:

1. Thread gets a signal, but previously issued a blocking syscall and is currently blocked.
2. Signal handlers don't get their own parallel threads. The syscall needs to exit to unblock and let the signal handler run. It returns EINTR.
3. The signal handler is run instead of returning to the syscall caller.
4. After the signal handler returns, control flow resumes at the syscall. The program can choose to retry.

However, we should really just be retrying up until the timeout. Retries == 3 is probably not going to blow any time budgets, but if we increase it much further, it starts to become a concern.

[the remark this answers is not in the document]
pushrax: https://www.gnu.org/software/libc/manual/html_node/Interrupted-Primitives.html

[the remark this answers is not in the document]
michaelkipper: For the record, that's:

```c
# define TEMP_FAILURE_RETRY(expression) \
(__extension__                                                              \
({ long int __result;                                                      \
do __result = (long int) (expression);                                      \
while (__result == -1L && errno == EINTR);                              \
__result; }))
```

Which is guarded by `__USE_GNU`. I recently made a commit that switched to `gnu99` format, so we should be OK to use `TEMP_FAILURE_RETRY` if you want (although it's unbounded).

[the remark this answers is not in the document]
pushrax: Some other thinking I didn't include above:

When handling a signal, an application can either cancel execution, or let it continue. If it cancels it (with `exit` or whatever), it doesn't matter that we would have retried, as the process will run atexit handlers and die before we can. If it doesn't cancel it, the signal should not raise an error in whatever random syscall was being made at the time. That would break the correctness of the program.

However, a common pattern is to handle a signal, and mark a flag to gracefully terminate. In this case, we still need to make sure we run all logic correctly. In semian's case, this is really important to avoid corrupting semaphore state. However, we don't want to block forever. That's one reason why it's important to respect the timeout. I wonder if we should remove the null timeout path entirely 🤔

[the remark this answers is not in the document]
thegedge: > However, we should really just be retrying up until the timeout

This seems like the best option, since that can be configured at the application level. The only function that doesn't use a timed operation is [`sem_meta_unlock `](https://github.com/Shopify/semian/blob/fc4f4e9d95215e3d45e61730adea0bff1913ddb2/ext/semian/sysv_semaphores.c#L141).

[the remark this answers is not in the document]
pushrax: The unlock path shouldn't ever actually block, maybe let's just give it a long timeout and require some timeout is passed.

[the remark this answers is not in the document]
pushrax: > TEMP_FAILURE_RETRY

Yeah, let's not use that as (like Michael pointed out) it's unbounded, but I quite like all the context provided on that doc page.

test/config/hosts.yml
paranoidaditya: Made this an array instead

paranoidaditya:
> > It seems like WEBrick is too slow to startup/shutdown, and the tests here can probably make do with a just a TCP server.
> 
> What were the results of looking into that?

https://gist.github.com/paranoidaditya/012e75155ff895c570dac4f397304633

I had 

> > It seems like WEBrick is too slow to startup/shutdown, and the tests here can probably make do with a just a TCP server.
> 
> What were the results of looking into that?

I had this, but could not get it working

```rb
class TestServer
  def initialize(hostname, port)
    @hostname = hostname
    @port = port
  end

  def start
    server = TCPServer.new(@hostname, @port)
    loop do                          
      client_connection = server.accept
      Thread.start(client_connection) do |session|
        request = session.gets
        
        _, full_path = request.split(' ')
        path, _ = full_path.split('?')
  
        response_code = path.delete("/")
        response_code = '200' if response_code == ""
  
        session.print "HTTP/1.1 #{response_code}\r\n"
  
        session.print "Content-Type: text/html\r\n"
  
        session.print "\r\n"
  
        session.close
      end
    end
  end
end
```

ext/semian/sysv_semaphores.c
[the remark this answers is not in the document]
paranoidaditya: I reverted this change in this PR.

pushrax:
The webrick parsing code is pretty convenient and you could still use that with a custom server, which is what I ended up with for the shop mover https://github.com/Shopify/shopify/blob/85e51d23997ab71faa7beded7273457ecb7a38be/components/pod_orchestrator/eagerlib/podding/shop_mover/ghostferry.rb#L248

Hard to say what might be wrong with that code without trying it out. I don't see any obvious flaws. I'd probably remove the threading code entirely because it's not necessary here (don't need to have more than 1 request being handled concurrently), and add a `Connection: close` response header to avoid keepalive issues.

paranoidaditya:
Replaced the WEBrick server with a TCP server in https://github.com/Shopify/semian/pull/231/commits/60384ea12119937f23094d320ee311fc6880ab54

```
bundle exec ruby -Itest test/net_http_test.rb
Finished in 11.548113s, 1.9917 runs/s, 7.0141 assertions/s.
```

test/helpers/mock_server.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+require 'webrick'
+
+module MockServer
```
thegedge: My only issue here is poor encapsulation. `cleanup` technically expects a thread id for something started by `MockServer`, but really we could give it any thread id and things will be fine.

A much better design is to return instances of `MockServer`. You can simplify construction through a class method, but hide the thread id in the class:

```ruby
class MockServer
  class << self
    def start(hostname:, port:)
      new.tap do |server|
        server.start
        server.poll_until_ready
        puts "Created test server #{tid}, port: #{port}"
      end
    end
  end

  def initialize(hostname:, port:)
    @hostname = hostname
    @port = port
    @tid = nil
  end

  def start
    @tid = Thread.new do
      start_server(hostname, port)
    end
  end

  def stop
    Thread.kill(@tid)
  end

  ...
end
```

test/helpers/mock_server.rb
```diff
@@ -0,0 +1,57 @@
...
+        req = WEBrick::HTTPRequest.new(config)
+        req.parse(sock)
+
+        response_code = req.path_info.delete("/")
+        response_code = 200 if response_code == ""
+
+        res.status = response_code
+        res.content_type = 'text/html'
+      rescue WEBrick::HTTPStatus::EOFError, WEBrick::HTTPStatus::BadRequest
+        res.status = 200
```
pushrax: Is 200 the most useful status in this case?

test/helpers/mock_server.rb
```diff
@@ -0,0 +1,57 @@
...
+
+module MockServer
+  extend self
+
+  def start(hostname:, port:)
+    tid = Thread.new do
+      start_server(hostname, port)
+    end
+    poll_until_ready(hostname: hostname, port: port)
+    puts "Created test server #{tid}, port: #{port}"
```
pushrax: Was this intended to be left in?

paranoidaditya: I think you had this file rendered pre https://github.com/Shopify/semian/pull/231/commits/d3c6d5e907ce1ee8b294b43cc7df4b6cb0fb7272

test/helpers/mock_server.rb
paranoidaditya: Switched it out to return 400 in #235

test/helpers/mock_server.rb
pushrax: Ah, well there's still a couple `puts` in the newer code as well.

paranoidaditya: I changed this up a bit but yeah the `puts` for starting & killing the servers is intentional.

PR #233, opened 2019-06-18
Retry syscalls in perform_semop if interrupted

paranoidaditya:
## What 

Retry `semop` and `semtimedop` syscalls in `perform_semop`. 

## Why 

When `semop` and/or `semtimedop` system calls are interrupted by a signal, they are not automatically retried, resulting in failure to acquire/set semian resources. 

```
Semian::InternalError: error setting ticket count, errno: 4 (Interrupted system call)
```

### DIFF
--- a/ext/semian/sysv_semaphores.c
+++ b/ext/semian/sysv_semaphores.c
@@ -100,17 +100,20 @@ set_semaphore_permissions(int sem_id, long permissions)
 int
 perform_semop(int sem_id, short index, short op, short flags, struct timespec *ts)
 {
+  int result;
   struct sembuf buf = { 0 };
 
   buf.sem_num = index;
   buf.sem_op  = op;
   buf.sem_flg = flags;
 
-  if (ts) {
-    return semtimedop(sem_id, &buf, 1, ts);
-  } else {
-    return semop(sem_id, &buf, 1);
-  }
+  int num_retries = 3;
+
+  do {
+    result = semtimedop(sem_id, &buf, 1, ts);
+  } while (result < 0 && errno == EINTR && num_retries-- > 0);
+
+  return result;
 }
 
 int


ext/semian/sysv_semaphores.c
```diff
@@ -100,17 +100,21 @@ set_semaphore_permissions(int sem_id, long permissions)
...
   buf.sem_num = index;
   buf.sem_op  = op;
   buf.sem_flg = flags;
 
-  if (ts) {
-    return semtimedop(sem_id, &buf, 1, ts);
-  } else {
-    return semop(sem_id, &buf, 1);
-  }
+  int syscall_retry = 3;
```
michaelkipper: Readability nit: Either use `num_retries`, starting at zero (with `num_retries < 3`) or `retries_remaining` starting at 3 (`retries_remaining > 0`).

```c
int num_retries = 3;
do {
  result = semtimedop(sem_id, &buf, 1, ts);
} while (result < 0 && errno == EINTR && num_retries-- > 0);
```

ext/semian/sysv_semaphores.c
```diff
@@ -100,17 +100,20 @@ set_semaphore_permissions(int sem_id, long permissions)
...
 
-  if (ts) {
-    return semtimedop(sem_id, &buf, 1, ts);
-  } else {
-    return semop(sem_id, &buf, 1);
-  }
+  int num_retries = 3;
+
+  do {
+    result = semtimedop(sem_id, &buf, 1, ts);
```
pushrax: Is passing a null timespec to semtimedop equivalent to calling semop? That codepath seems to be gone.

paranoidaditya: Yes! `If the timeout argument is NULL, then
       semtimedop() behaves exactly like semop().`

PR #235, opened 2019-06-20
Docker based development environment

paranoidaditya:
~Requires #231~

## What 

- Replace Vagrant with a docker-compose based development environment.
- Change CI to use the docker-compose based environment.
- Use Visual Studio Code with https://code.visualstudio.com/docs/remote/containers   (optional, not forced on the end user)

### Why Visual Studio Code?

- https://code.visualstudio.com/docs/remote/containers really simplifies the development workflow on machines not running Linux, where one can do everything within inside the IDE and not have to switch between terminal windows (docker container shell & host shell) 

## TODOs

- Re-enable rubocop tests after the repo is formatted. (Should be a separate PR, given it touches everything)
- CI/Tests Fixes
	- ~Fix `Semian::InternalError: error setting ticket count, errno: 4 (Interrupted system call)`~ Done in https://github.com/Shopify/semian/pull/235/commits/cfd38482dd1824d09386f733bf8a1bc1bb55486e
	- ~Fix `TestGRPC#test_timeout_opens_the_circuit`~ Done in https://github.com/Shopify/semian/pull/235/commits/abdd35a247649e506253d6be98d95f85c4b107c3

cc: @Shopify/servcomm


michaelkipper:
Added @sirupsen and @csfrancis since this is a major change to workflow.

dalehamel:
My two cents here is that the only reason we used vagrant for this was that things like docker for mac (or windows) didn't exist at the time.

Vagrant was, at the time, the easiest way to get a Linux environment up and running. Now that the both windows and OS X have good hypervisor support to run Linux in order to get docker up and running easily, I think that this change makes a lot of sense.

I see that the dev and CI containers use different dockerfiles though. Looks like the CI container aims to be a bit thinner, but is it worth the divergence?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,34 @@
+{
+    // See https://aka.ms/vscode-remote/devcontainer.json for format details
```
thegedge: JSON can't have comments (unless some parser supports it, but then it's nonstandard)

README.md
```diff
@@ -681,37 +681,54 @@ non-IO.
 
 # Developing Semian
 
-Semian requires a Linux environment.  We provide a [Vagrantfile](Vagrantfile)
-that installs MySQL, Redis and Ruby in an Ubuntu 16.04 virtual machine.  Use
+Semian requires a Linux environment.  We provide a [docker-compose](https://docs.docker.com/compose/) file
+that runs MySQL, Redis, ToxiProxy and Ruby in containers.  Use
```
thegedge: nit: Toxiproxy doesn't use a capital `P`

.devcontainer/devcontainer.json
paranoidaditya: This from the reference file at
https://github.com/microsoft/vscode-dev-containers/blob/master/container-templates/docker-compose/.devcontainer/devcontainer.json

paranoidaditya:
> I'm pretty hesitant to include all of this VSCode-specific configuration. I don't think we should be including editor-specific things in our repos, even if any other workflow can safely ignore it. I'll let @sirupsen or @csfrancis make that call though :)

We do have editor specific configurations in several repos within Shopify:
- https://github.com/Shopify/shopify/tree/master/.vscode
- https://github.com/Shopify/web/tree/master/.vscode

That being said, visual-studio-code here is as essential of a change as swapping out vagrant with docker, it greatly reduces the complexity of getting up and running with developing Semian. Basically abstracting away the containers from an end developer.

.devcontainer/devcontainer.json
thegedge: I can't see anyone else needing to read this file, so if this is what VSCode does then we can keep it that way.

csfrancis:
> Vagrant was, at the time, the easiest way to get a Linux environment up and running. Now that the both windows and OS X have good hypervisor support to run Linux in order to get docker up and running easily, I think that this change makes a lot of sense.

👍 

I don't feel super strongly about the VSCode configuration living in the repo, so long as it's not mandatory. It doesn't seem to be adding a lot of cruft, so it's fine with me.

README.md
```diff
@@ -681,37 +681,40 @@ non-IO.
...
-$ toxiproxy-server > /dev/null &
-# TODO: provide a systemd unit for toxiproxy
-
-# run the tests inside the virtual machine
-$ cd /vagrant && bundle exec rake
-```
+- Everything else
+  - `$ docker-compose -f .devcontainer/docker-compose.yml run semian bash`
 
 Be careful not to run `bundle install` from Mac OS.  The folder is shared with
```
michaelkipper: Is this still true?
If so, why? Can't we just run `bundle install` inside the container, to avoid polluting the external environment?

paranoidaditya: Took this out as its no longer affects the container.
Also changed the docker-compose file to start in `/workspace`, run `bundle install` and then `sleep infinity`.

README.md
```diff
@@ -681,37 +681,57 @@ non-IO.
...
+  
+  See https://code.visualstudio.com/docs/remote/containers for more details
 
-# start the toxiproxy-server inside the virtual machine
-$ toxiproxy-server > /dev/null &
-# TODO: provide a systemd unit for toxiproxy
+- Everything else
+  - `$ docker-compose -f .devcontainer/docker-compose.yml up -d`
+  - `$ docker exec -it semian-dev bash`
+  - `$ cd /workspace && bundle install`
```
thegedge: What if we added `WORKDIR /workspace` to the `Dockerfile`?

README.md
```diff
@@ -681,37 +681,57 @@ non-IO.
...
+  -  Use vim or your TUI editor of choice inside the container 
+  
+  OR
+  
+  - Make changes outside the container and run tests inside the container
```
thegedge: nit: I'd either combine the `OR` instructions into one (since editing inside/outside the container doesn't matter) or drop the `OR` instructions completely, since people know how to edit files :)

Basically, just trying to simplify the instructions as much as possible.

README.md
```diff
@@ -681,37 +681,57 @@ non-IO.
...
+  - `$ docker exec -it semian-dev bash`
+  - `$ cd /workspace && bundle install`
+  - `$ bundle exec rake`
+
+
+If you make any changes to `.devcontainer/..`, you'd need to recreate the containers:
+  - `$ docker-compose -f .devcontainer/docker-compose.yml up -d --force-recreate` 
+
+
+  OR
```
thegedge: nit: All of these `OR` instructions complicates the flow. It may be better organized into two sections: one dedicated to VSCode and another for everyone else.

scripts/run_tests.sh
```diff
@@ -0,0 +1,43 @@
...
+  if (( attempts > 60 )); then
+    echo "ERROR: mysql was not started." >&2
+    exit 1
+  fi
+done
+echo "MySQL has started!"
+
+echo "Running bundle install"
+bundle install --jobs=3 --retry=3 2>&1
+if [[ $? -ne 0 ]]; then
```
thegedge: You can simplify:

```sh
my_command
if [[ $? -ne 0 ]]; then
  echo "..."
  exit 1
fi
```

to

```sh
if ! my_command; then
  echo "..."
  exit 1
fi
```

scripts/run_tests.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+echo "Waiting for MySQL to start."
+attempts=0
+while ! nc -w 1 mysql 3306 | grep -q "mysql"; do
+  sleep 1
```
thegedge: Were you aiming for 60 seconds? If so, you can drop this sleep, since the `-w 1` act almost as a sleep (while mysql isn't running).

test/redis_test.rb
```diff
@@ -295,9 +295,9 @@ def test_circuit_breaker_on_query
   private
 
   def new_redis(options = {})
+    options[:host] = SemianConfig['toxiproxy_upstream_host'] if options[:host].nil?
```
thegedge: nit: Why this change? The benefit of the previous approach is that it didn't mutate the options hash, so callers of `new_redis` won't be surprised that it was changed :)

paranoidaditya: `options[:host]` was always truthy when I was playing around with this, would result in `host` being empty and `Redis.new` falling back to `localhost`

README.md
paranoidaditya: I tried that, appears like `docker-compose` overrides that so added this in the compose file 🎉

scripts/run_tests.sh
paranoidaditya: Done!

paranoidaditya: Seems like it would not sleep with `-w 1` and would error out after 60 continuous attempts. Added it back.
```
semian_1     | Waiting for MySQL to start.
.
.
.
.
semian_1     | mysql [172.18.0.3] 3306 (mysql) : Connection refused
semian_1     | ERROR: mysql was not started.
```

paranoidaditya:
> A few nitpicks to help polish, but my 🎩 was ✅ 🎉
> 
> The one test that was failing though, was this:
> 
> ```
> TestNetHTTP#test_5xxs_dont_raise_exceptions_unless_fatal_server_flag_enabled:
> Net::CircuitOpenError: [nethttp_semian_31050] Semian::OpenCircuitError caused by Internal Server Error
> ```

Its one of the tests I marked as flaky in https://github.com/Shopify/semian/pull/235/commits/6d0f38e66af8f07b2652da9dd3649ace728e7fa1

I'll follow up with a PR which where testing uses https://github.com/Shopify/ci-queue/ and only retries the flaky tests than retrying the whole suite if one test fails

paranoidaditya:
> My two cents here is that the only reason we used vagrant for this was that things like docker for mac (or windows) didn't exist at the time.
> 
> Vagrant was, at the time, the easiest way to get a Linux environment up and running. Now that the both windows and OS X have good hypervisor support to run Linux in order to get docker up and running easily, I think that this change makes a lot of sense.
> 
> I see that the dev and CI containers use different dockerfiles though. Looks like the CI container aims to be a bit thinner, but is it worth the divergence?

I kept the CI container slim because Travis has been really slow to work with.

Another reason was trying to keep the test runs in an immutable environment in between test retries.

scripts/run_tests.sh
thegedge: Interesting. It must be able to fail the connection right away, instead of hang on waiting for it to establish :+1:

README.md
```diff
@@ -681,37 +681,54 @@ non-IO.
...
+$ brew cask install visual-studio-code
 
 # clone Semian
 $ git clone https://github.com/Shopify/semian.git
 $ cd semian
+```
 
-# build the virtual machine (can take a while for ruby to compile)
-$ vagrant up
+# Developing :
```
thegedge: nit: Drop this. We're already under the `Developing` header :)

README.md
```diff
@@ -681,37 +681,54 @@ non-IO.
 
 # Developing Semian
 
-Semian requires a Linux environment.  We provide a [Vagrantfile](Vagrantfile)
-that installs MySQL, Redis and Ruby in an Ubuntu 16.04 virtual machine.  Use
+Semian requires a Linux environment.  We provide a [docker-compose](https://docs.docker.com/compose/) file
+that runs MySQL, Redis, Toxiproxy and Ruby in containers.  Use
 the steps below to work on Semian from a Mac OS environment.
 
+Prerequisites :
```
thegedge: Perhaps make this a `## Prerequisites` section?

paranoidaditya:
Two changes since @thegedge reviewed:

- Poll for MySQL readiness after running `bundle install` & `bundle exec rake build` instead of running it as the first thing in the test script.

- Retry `bundle exec rake test` 5 times in the script.

PR #238, opened 2019-06-25
Notify on circuit breaker state transitions

michaelkipper:
Send `:state_change` notifications when circuit breaker state changes.

This is useful to monitor the current Semian state in the simulator we're building.
We monitor state as a Prometheus gauge, with `closed=0` and `open=1`.

**Note**: This further shows the weakness of the notification API as currently written. The `(event, resource, scope, adapter, payload)` signature doesn't apply universally throughout the codebase. In this case, the `resource` is actually the `CircuitBreaker` instance, not the `ProtectedResource` like the other notifications. A better API would simply be `(event, payload)` where the payload is context dependent on the event type.

cc: @Shopify/servcomm

### DIFF
--- a/lib/semian/circuit_breaker.rb
+++ b/lib/semian/circuit_breaker.rb
@@ -6,7 +6,8 @@ module Semian
 
     attr_reader :name, :half_open_resource_timeout, :error_timeout, :state, :last_error
 
-    def initialize(name, exceptions:, success_threshold:, error_threshold:, error_timeout:, implementation:, half_open_resource_timeout: nil)
+    def initialize(name, exceptions:, success_threshold:, error_threshold:,
+                         error_timeout:, implementation:, half_open_resource_timeout: nil)
       @name = name.to_sym
       @success_count_threshold = success_threshold
       @error_count_threshold = error_threshold
@@ -17,6 +18,8 @@ module Semian
       @errors = implementation::SlidingWindow.new(max_size: @error_count_threshold)
       @successes = implementation::Integer.new
       @state = implementation::State.new
+
+      reset
     end
 
     def acquire(resource = nil, &block)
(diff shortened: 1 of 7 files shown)


michaelkipper:
> the resource is actually the CircuitBreaker instance

I should note that it has to be this way, since the first state transition to `:closed` is during circuit breaker construction, which happens before the `ProtectedResource` is created. A previous version of this code attempted to use `Semian[@name]` to use the `ProtectedResource` as the `resource` parameter, but this meant missing the initial state. Since we're only really interested in `resource.name`, it's not a big deal here but still requires some care with type-safety by consumers.

test/grpc_test.rb
```diff
@@ -85,8 +85,9 @@ def test_timeout_opens_the_circuit
   def test_instrumentation
     notified = false
     subscriber = Semian.subscribe do |event, resource, scope, adapter|
+      return if event != :success
```
thegedge: `return` in a block applies to the function it's in, which in this case would be `test_instrumentation`. If you just want to not have the block executed for the notification but the rest of the function to continue, `next` is what you're looking for.

Similar for some other tests.

michaelkipper: Done.

lib/semian/circuit_breaker.rb
```diff
@@ -127,6 +133,10 @@ def log_state_transition(new_state)
       Semian.logger.info(str)
     end
 
+    def notify_state_transition(new_state)
+      Semian.notify(:state_change, self, nil, nil, state: new_state)
```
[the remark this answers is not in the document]
michaelkipper: I originally had that, but it requires the consumer to think about `nil` as the previous state (e.g. the initial transition to `closed`) which seemed unnecessary for my usecase. Always possible to add in later, with the `payload` system (rather than explicit parameters).

PR #239, opened 2019-06-25
Add circuit name to state transition log message

spike01:
It's useful to see which circuit is currently changing state!

### DIFF
--- a/lib/semian/circuit_breaker.rb
+++ b/lib/semian/circuit_breaker.rb
@@ -124,6 +124,7 @@ module Semian
       str << " success_count=#{@successes.value} error_count=#{@errors.size}"
       str << " success_count_threshold=#{@success_count_threshold} error_count_threshold=#{@error_count_threshold}"
       str << " error_timeout=#{@error_timeout} error_last_at=\"#{@errors.last}\""
+      str << " name=\"#{@name}\""
       Semian.logger.info(str)
     end
 


lib/semian/circuit_breaker.rb
```diff
@@ -124,6 +124,7 @@ def log_state_transition(new_state)
       str << " success_count=#{@successes.value} error_count=#{@errors.size}"
       str << " success_count_threshold=#{@success_count_threshold} error_count_threshold=#{@error_count_threshold}"
       str << " error_timeout=#{@error_timeout} error_last_at=\"#{@errors.last}\""
+      str << " name=#{@name.inspect}"
```
thegedge: Could `@name` have spaces or other weird characters? Might be a good idea to take the approach of `error_last_at` and surround it with quotes.

lib/semian/circuit_breaker.rb
```diff
@@ -124,6 +124,7 @@ def log_state_transition(new_state)
       str << " success_count=#{@successes.value} error_count=#{@errors.size}"
       str << " success_count_threshold=#{@success_count_threshold} error_count_threshold=#{@error_count_threshold}"
       str << " error_timeout=#{@error_timeout} error_last_at=\"#{@errors.last}\""
+      str << " name='#{@name.inspect}'"
```
thegedge: If I had to nit hard, I'd be consistent with `error_last_at` and use `\"` instead of `'`, and also not use `inspect`. I think knowing symbol versus string here shouldn't be important. If it does end up being a string we'd see `'"resource_name"'`.

PR #242, opened 2019-06-27
CI/Test Suite Improvements

epk:
## What

- Add back Ruby 2.4, 2.5 in CI
- Remove redundant `gemfiles/mysql2-0-5-0.gemfile` (`mysql2 ~> 0.5` is already required by the gemspec).
- ~Instead of running the whole test suite on failure just rerun the failed test.~

~Also instead of doing 15 full retries (5 retries if bundle exec rake test returned non zero x 3 times travis retrying the build), just run the failed test 15 times, sleeping for 10s in between retries~

Update: 

In https://github.com/Shopify/semian/pull/242/commits/a324308b7d2fb0f00b0c0be3f2ad9b7cba337d1c we are completely skipping tests as the failures appear to deterministic/leaky and retrying them in the same process does not help.

cc @Shopify/servcomm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,12 @@ before_install:
   - curl -L https://github.com/docker/compose/releases/download/1.24.0/docker-compose-`uname -s`-`uname -m` > docker-compose
   - chmod +x docker-compose
   - sudo mv docker-compose /usr/local/bin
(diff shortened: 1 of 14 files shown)


thegedge:
tl;dr I'm 👍 for this change.

Not sure if I'm in the majority or minority, but I lean towards always skipping flaky tests until they can be made not flaky. This may be a little aggressive, but I find flaky tests cause more pain than value (and can often be fixed fairly easily).

PR #243, opened 2019-07-01
Force-enable host-based circuits

michaelkipper:
## What

This PR allows `SEMIAN_CIRCUIT_BREAKER_FORCE_HOST` to force-enable host-based circuits.

## Why

To deploy #227 on core, we want to enable host-based circuits on a single node, so we can observe their behaviour through a Rails console.

cc: @Shopify/servcomm


ext/semian/semian.c
```diff
@@ -73,9 +73,62 @@ void Init_semian()
...
+    if (strcmp(hostname, cursor) == 0) {
+      fprintf(stderr, "Info: Semian forcing host-based circuit implementation\n");
+      retval = 1;
+      break;
+    }
+  }
+
+  free(hostnames);
+  return retval;
+}
```
thegedge: My own preference would be for more of this logic to be in Ruby land (more familiar, easier to test, etc), and we'd swap out the implementations there, but this works ;)

ext/semian/sliding_window.c
```diff
@@ -400,6 +437,45 @@ semian_simple_sliding_window_clear(VALUE self)
...
+
+  int cleared = 0;
+  for (int i = 0; i < length; ++i) {
+    int index = (start + i) % length;
+    int value = window->data[index];
+    VALUE y = rb_yield(RB_INT2NUM(value));
+    if (RTEST(y)) {
+      if (cleared++ != i) {
+        sem_meta_unlock(sem_id);
+        rb_raise(eInternal, "reject must delete monotonically");
```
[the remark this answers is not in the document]
michaelkipper: You are 100% right. It must delete contiguously, which is only possible when the data is monotonic.

ext/semian/semian.c
michaelkipper: Me too, but this code is called when the module is being loaded and I'm not confident what Ruby code is available at this point.

thegedge: :+1: We can always come back and factor it out to Ruby land later. If it can be the same behaviourally, it shouldn't be an issue.

ext/semian/sliding_window.c
```diff
@@ -409,50 +485,40 @@ semian_simple_sliding_window_reject(VALUE self)
...
-    res->shmem->data[res->shmem->end] = RB_NUM2INT(value);
-    dprintf("Pushed %d onto data[%d] (length %d)", RB_NUM2INT(value), res->shmem->end, res->shmem->length);
-    res->shmem->end = (res->shmem->end + 1) % res->shmem->max_size;
+    int timestamp = check_timestamp_arg(time);
+    push(res->shmem, timestamp);
+  }
+  sem_meta_unlock(res->sem_id);
+
+  return self;
+}
```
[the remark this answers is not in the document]
michaelkipper: I refactored the code to remove `reject_and_push!`.
I was thinking about the wrong ends of the sliding window. The reject operates on the older end, and the push operates on the newer end. So there's no coupling between the two that requires an atomic operation. I think...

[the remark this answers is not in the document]
thegedge: > I think...

That's what I was thinking too 👍

ext/semian/semian.c
csfrancis: I also agree with doing this in Ruby. I'm fairly sure the C extension is loaded here: https://github.com/Shopify/semian/blob/056cd1092d071ba5a970c8fc646e5525715fba77/lib/semian.rb#L270

michaelkipper: Done.

lib/semian.rb
```diff
@@ -296,7 +296,19 @@ def require_keys!(required = [], **options)
   end
 end
 
+def hostname
+  v = %w(KUBE_HOSTNAME KUBE_HOST_NAME KUBE_NODENAME KUBE_NODE_NAME NODENAME NODE_NAME HOSTNAME HOST_NAME)
+  var = v.filter { |x| ENV.include?(x) }.first
```
thegedge: `filter { ... }.first` is the same as `find { ... }`, minus the need to construct the intermediate array.

lib/semian.rb
```diff
@@ -296,7 +296,19 @@ def require_keys!(required = [], **options)
   end
 end
 
+def hostname
+  v = %w(KUBE_HOSTNAME KUBE_HOST_NAME KUBE_NODENAME KUBE_NODE_NAME NODENAME NODE_NAME HOSTNAME HOST_NAME)
```
thegedge: small optimization: Looks like `KUBE_HOSTNAME` is the thing for shopify core. Outside of that, I've typically seen `HOSTNAME`. Perhaps make `HOSTNAME` the second entry in this list?

lib/semian.rb
michaelkipper: Nice!

lib/semian.rb
michaelkipper: I wanted to leave things that are commonly defined to the end of the list, to allow the custom ones to be selected.

```
root@web-5b44f8fbbd-27bh5:/app# set | grep -i -e HOST -e NODE
HOSTNAME=web-5b44f8fbbd-27bh5
HOSTTYPE=x86_64
HOST_IP=10.192.1.141
KUBERNETES_SERVICE_HOST=10.188.0.1
KUBE_HOSTNAME=gke-tier1-us-east-4-default-2-c7098a55-z3jr
PROXYSQL_ADMIN_MASTER_SERVICE_HOST=10.188.5.35
PROXYSQL_MASTER_SERVICE_HOST=10.188.11.180
SMTP_SERVICE_HOST=smtp.smtp-gke-production
WEB_PERF_SERVICE_HOST=10.188.11.159
WEB_SERVICE_HOST=10.188.15.146
```

Notice that `HOSTNAME=web-5b44f8fbbd-27bh5` but `KUBE_HOSTNAME=gke-tier1-us-east-4-default-2-c7098a55-z3jr` is what we're really after.

lib/semian.rb
```diff
@@ -296,7 +296,19 @@ def require_keys!(required = [], **options)
...
+  var = v.find { |x| ENV.include?(x) }
+  ENV[var] if var
+end
+
+def force_host_circuits
```
csfrancis: This should be `force_host_circuits?` to be more Ruby-like.

lib/semian.rb
thegedge: Sounds good :+1:

lib/semian.rb
michaelkipper: Done.

PR #246, opened 2019-07-03
Don't force sliding window entries to be monotonic

michaelkipper:
The code in #227 is not passing tests, because it forces an invariant on the `SlidingWindow` that entries be monotonic. This doesn't work, unfortunately, with host-based circuits, since individual processes perform a two-step operation of:

1. Getting the current timestamp, and
1. Pushing the timestamp onto the sliding window

These operations, performed without a lock, can be interleaved and violates the invariant.

I tried to make that operation atomic, but it breaks with tests that want to use `Timecop.travel` to modify the system time. Therefore, the only option I had was to relax the invariant.

This does change the runtime of the `reject!` function to Ω(n) from O(n). Given that n tends to be less than 100, even in host-based circuits with a large `scale_factor`, this shouldn't be a big deal.

cc: @Shopify/servcomm

### DIFF
--- a/ext/semian/sliding_window.c
+++ b/ext/semian/sliding_window.c
@@ -34,7 +34,6 @@ static void init_fn(void* ptr)
   res->max_size = 0;
   res->length = 0;
   res->start = 0;
-  res->end = 0;
 }
 
 static int
@@ -92,16 +91,17 @@ check_scale_factor_arg(VALUE scale_factor)
 }
 
 static VALUE
-grow_window(semian_simple_sliding_window_shared_t* window, int new_max_size)
+grow_window(int sem_id, semian_simple_sliding_window_shared_t* window, int new_max_size)
 {
   if (new_max_size > SLIDING_WINDOW_MAX_SIZE) return Qnil;
 
+  int end = window->max_size ? (window->start + window->length) % window->max_size : 0;
+  dprintf("Growing window - sem_id:%d start:%d end:%d length:%d max_size:%d new_max_size:%d", sem_id, window->start, end, window->length, window->max_size, new_max_size);
+
   if (window->length == 0) {
     window->start = 0;
-    window->end = 0;
-  } else if (window->end > window->start) {
+  } else if (end > window->start) {
     // Easy case - the window doesn't wrap around
-    window->end = window->start + window->length;
   } else {
     // Hard case - the window wraps, and data might need to move
     int offset = new_max_size - window->max_size;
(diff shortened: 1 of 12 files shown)


thegedge:
Some food for thought: when I was going through the code of `SlidingWindow`, my thought was that it doesn't feel like what I'd expect out of a sliding window data structure.

In particular, the fact that the caller has to manage timestamps doesn't make sense to me, because I would expect the sliding window to manage that. For example, I would expect to be able to write code like this:

```ruby
# Width being time in seconds, and max_size being the size of the underlying ring buffer / storage data structure
window = SlidingWindow.new(width, max_size: 100)

# Can push arbitrary data, SlidingWindow manages timestamps
window.push("value")

# Lazily do reject, instead of caller having to decide how to reject
window.size
```

If it completely managed timestamps, I think it would be far easier to avoid these monotonicity issues.

michaelkipper:
@thegedge: I agree that this is a very opinionated SlidingWindow. In particular, it's a window of timestamps, not data attached to timestamps. The key is that the `max_size` defines the number of timestamps that can be in the window, which makes it simple to figure out when to open the circuit.

    # Can push arbitrary data, SlidingWindow manages timestamps
    window.push("value")

I suppose this might work with `Timecop.travel` if you have the underlying C implementation use the Ruby version of time instead of primitives like `time(2)`.

In any event, I still think that this PR is a _simplification_ on #227, so a net win.

ext/semian/sliding_window.c
```diff
@@ -409,24 +418,30 @@ semian_simple_sliding_window_reject(VALUE self)
...
+      for (int i = 0; i < length; ++i) {
+        const int rptr = (start + length + max_size - i - 1) % max_size;
+
+        const int value = window->data[rptr];
+        if (RTEST(rb_yield(RB_INT2NUM(value)))) {
+          window->length--;
+          window->data[wptr] = value;
+        } else {
+          window->data[wptr] = value;
+          wptr = (wptr + max_size - 1) % max_size;
```
thegedge: I really love how Python chose to make the default implementation for `%` be the same as the mathematical definition, which then allows you to express this exactly the way you'd like: `(wptr - 1) % max_size`

ext/semian/sliding_window.c
```diff
@@ -409,24 +418,30 @@ semian_simple_sliding_window_reject(VALUE self)
...
+      int wptr = (start + length + max_size - 1) % max_size;
+
+      dprintf("Before reject! start:%d length:%d max_size:%d", window->start, window->length, window->max_size);
+      for (int i = 0; i < length; ++i) {
+        const int rptr = (start + length + max_size - i - 1) % max_size;
+
+        const int value = window->data[rptr];
+        if (RTEST(rb_yield(RB_INT2NUM(value)))) {
+          window->length--;
+          window->data[wptr] = value;
```
thegedge: Any reason to do this, since we're discarding `value`?

ext/semian/sliding_window.c
```diff
@@ -409,24 +418,30 @@ semian_simple_sliding_window_reject(VALUE self)
...
-        if (cleared++ != i) {
-          sem_meta_unlock(res->sem_id);
-          rb_raise(rb_eArgError, "reject! must delete monotonically");
+    semian_simple_sliding_window_shared_t *window = res->shmem;
+    const int start = window->start;
+    const int length = window->length;
+    const int max_size = window->max_size;
+
+    if (max_size && length) {
+      int wptr = (start + length + max_size - 1) % max_size;
```
thegedge: It takes a little bit of reading to grok exactly the approach you're taking here (move elements to the end of the buffer). Maybe a short comment saying what's happening below at a high level could be useful? Up to you.

ext/semian/sliding_window.c
michaelkipper: Nope. Removed.

ext/semian/sliding_window.c
michaelkipper: Done.

michaelkipper:
Merging this to `mkipper/global-circuit-breaker` so it's easier to review and reason about.

ext/semian/sliding_window.c
michaelkipper: Done.

Issue #244, opened 2019-07-02
Feature Request: Throttle half_open -> closed attempts


michaelkipper:
## What

Currently, when the `error_timeout` expires, the next acquisition request for a circuit will cause a transition from `open` to `half_open`. In this state, workers will attempt to access the resource with a modified timeout of `half_open_resource_timeout`. The motivation here is that the modified timeout is much lower than the client timeout so if the resource is still unhealthy, it will fail fast(er).

In the current implementation, every available worker (subject to the bulkhead configuration) will attempt the `half_open` -> `closed` transition. This means that if the resource is still unhealthy, all the workers could potentially block for `half_open_resource_timeout` seconds, reducing overall node capacity.

Mathematically, this means that `t[half-open] / (t[half-open] + t[error_timeout])` will be spent attempting to re-open the circuit. If `t[half-open]` is 1.0s and `t[error-timeout]` is 5.0s (our MySQL defaults) then 16.7% of our capacity will go toward re-opening the circuit. If bulkheads are in place with a quota of 0.5, that number will be 8.3%.

## How

When a circuit opens, the number of available tickets should immediately drop to 1. This shields the rest of the workers from this unhealthy resource. This is marginally faster than the open circuit error, since bulkhead acquisition is attempted before circuit-breaker acquisition, but that's likely not a big deal.

When the transition happens from `open` to `half_open`, we can raise the number of available tickets to `success_threshold`, to allow parallel re-closing of the circuit. Once the circuit is finally re-closed, we can raise the number of available tickets back to the original `tickets/quota` value.

PR #247, opened 2019-07-04
Throttle bulkhead from the circuit breaker on state transition

michaelkipper:
## What

This PR attempts to address #244. It throttles the number of bulkhead tickets to 1 when the circuit opens, throttles to `success_threshold` when the circuit moves to `half_open`, and removes the throttle when the circuit finally closes.

## Why

When the circuit is in an open state, we still allow all the workers that can acquire tickets to attempt the transition to `half_open`. When the circuit is half open, we allow all those workers to attempt the transition to `closed`, which can sap much needed capacity during a sustained outage.

**Note: This PR is still a work in progress, and there are several corner cases that need to be addressed. I'm looking for a high-level review of the approach.**

## Issues
  * It's a little gross to couple the bulkhead and circuit breaker in this way, but I'm not sure how else to approach it.
  * Not sure how `SEM_UNDO` works when one process applies the throttle, and another removes it. Or what happens when the process that applied the throttle terminates. Needs to be thought through more.

cc: @Shopify/servcomm


ext/semian/test/types_test.c
```diff
@@ -0,0 +1,36 @@
...
+  if (actual > expected) {
+    fprintf(stderr, "Error: got %d, which is greater than %d (%s)\n", actual, expected, message);
+    exit(EXIT_FAILURE);
+  }
+}
+
+void test_sliding_window()
+{
+  semian_simple_sliding_window_shared_t window;
+  assert_le_int(sizeof(window), 4096, "window size is greater than a page");
```
thegedge: How about `sysconf(_SC_PAGESIZE)` instead of 4096?

michaelkipper:
> Have you had any success with semsim?

No, I don't think this works at all...
`SI_SEM_CONFIGURED_TICKETS` is basically a bookkeeping value. The real value we want to modify is `SI_SEM_TICKETS`.
The problem is that at the point when we most likely want to reduce `SI_SEM_TICKETS` to 1, it's already close to zero. If we attempt to do a `-throttle` operation, that worker would block until the rest of the workers timed out on their requests. Given that we know we're opening the circuit, it would be awesome if we could terminate those requests immediately, but my brief research into this suggests that we'd have to take down the worker with a `SIGTERM` which is likely a very expensive operation.

PR #254, opened 2019-09-23
close block to avoid syntax errors on older versions

jordanify:
### This PR:
- is related to https://github.com/Shopify/semian/pull/193
- attempts to fix a syntax error when Semian is loaded into older versions of Rails app.

### Background info:
- I'm working on a feature in the [Kit app](https://github.com/Shopify/kit) that requires making graphql calls and thought it would be nice if we could use [Cusco](https://github.com/Shopify/cusco). 
- Kit is running version `2.3.4` of Ruby, and can't upgrade yet as they're still on AWS.
- I added the Cusco gem to Kit, which has a dependency on Semian.
- When I fire up the Kit app with Cusco & Semian installed, the app hits a `SyntaxError` in [lru_hash.rb](https://github.com/Shopify/semian/blob/3fc9810c0f469e13a30cbbd7d3e867132f0e6977/lib/semian/lru_hash.rb#L156):
trace_at_startup

### The Fix:
- Locally I changed the `try_synchronize` method to what's in this PR, mainly ending the initial `Thread.handle_interrupt` block with an `end`, so that the method behaves as if there was a `begin` at the top of the method.
- @casperisfine Looking at #193 I wasn't exactly sure if the `ensure` is there to cover the entire method, or if it was meant for only the block contents, but thought i'd whip this up as a starting point!

### DIFF
--- a/lib/semian/lru_hash.rb
+++ b/lib/semian/lru_hash.rb
@@ -150,11 +150,13 @@ class LRUHash
 
   def try_synchronize
     Thread.handle_interrupt(EXCEPTION_NEVER) do
-      return false unless @lock.try_lock
-      Thread.handle_interrupt(EXCEPTION_IMMEDIATE) { yield }
-      true
-    ensure
-      @lock.unlock if @lock.owned?
+      begin
+        return false unless @lock.try_lock
+        Thread.handle_interrupt(EXCEPTION_IMMEDIATE) { yield }
+        true
+      ensure
+        @lock.unlock if @lock.owned?
+      end
     end
   end
 end


jordanify:
@casperisfine thanks!  I added an explicit `begin` to the start of the block, is this any better?

jordanify:
@casperisfine thanks again!  Will i need to bump the version and cut a new release for this change?

vosechu:
Thank you! I just came here to make the same PR and you're already ahead of me!

casperisfine:
@jordanify really sorry, somehow I never got the notification for your last comment.

Let's merge this, and if it's critical for you we can cut a new release, otherwise it can wait a bit.

jordanify:
@casperisfine no problem - this ended up being a super-low priority for us that we ended up workaround anyways.  Now that it's in we'll come back to it eventually!

vosechu:
@casperisfine This is really important for us. If you're willing, we would really appreciate a release. It doesn't happen a lot for us, but when it does it's a doozy. 

@jordanify What did you do to work around it?

jordanify:
@vosechu i guess you could say it was a replacement and not a workaround...we ended up loading the `graphql-client` gem, which Cusco uses, and writing a very basic interface to fire the requests we needed.  Definitely not ideal, but it tied us over until the app is in a place to update gems.

vosechu:
@jordanify Thank you! Makes sense. 

If we need to, I can monkey patch or install this revision directly. But it'd be nice to close this out!

Thank you for sending in this PR and thank you to Casper for maintaining! This project has been a huge help for us and I really appreciate it.

casperisfine:
I'll cut a new release first thing tomorrow

vosechu:
Thank you!

casperisfine:
0.9.1 is out.

vosechu:
Thank you so much! This made a huge difference for us. We’re going to test
in prod during a maintenance window, but so far we’ve already seen a marked
improvement on circuit close times.

I don’t quite know how to express how much this gem has improved our work.
We see 200k requests per minute, so an outage of one of the shards can
exhaust our unicorn pool pretty quickly. Bulkheading really is the win
button for us. So thankful for your team open sourcing this!

PR #255, opened 2019-11-22
remove call to match? method not present in < ruby 2.4

arsac:
We're running ruby 2.3.4 and just updated to fix the issue with circuit breakers not closing but are now getting `NoMethodError` since `match?` was added in 2.4.  This will keep the fix that was added but restore compatability with 2.3.4


sirupsen:
@arsac you fixed it somehow else?

arsac:
@sirupsen Just needed to update with my github email, the new pr is here: https://github.com/Shopify/semian/pull/256

PR #256, opened 2019-11-25
remove call to match? method not present in < ruby 2.4

arsac:
We're running ruby 2.3.4 and just updated to fix the issue with circuit breakers not closing but are now getting `NoMethodError` since `match?` was added in 2.4. This will keep the fix that was added but restore compatibility with 2.3.4


miry:
Hi. Sorry for so long long reply. I think it is not actual, because only supported 2.7 and 3.0 rubies.

PR #257, opened 2019-12-03
Support half_open_resource_timout for Redis

damianthe:


lib/semian/redis.rb
```diff
@@ -91,6 +91,30 @@ def connect
       end
     end
 
+    def with_resource_timeout(temp_timeout)
+      prev_timeouts = {
+        timeout: options[:timeout],
+        connect_timeout: options[:connect_timeout],
+        read_timeout: options[:read_timeout],
+        write_timeout: options[:write_timeout],
+      }
```
sirupsen: Why not just use local variables?

test/redis_test.rb
```diff
@@ -243,6 +245,49 @@ def test_query_instrumentation
     Semian.unsubscribe(subscriber)
   end
 
+  def test_changes_timeout_when_half_open_and_configured
```
sirupsen: This test only tests the connect or read timeout, but most likely not both (which one requires some more intimacy with the gem, I'm not sure if it kills the connection after a timeout--some clients do). Most likely it maintains the connection and you 'just' need another test that'll disconnect before your second toxiproxy.

test/redis_test.rb
```diff
@@ -243,6 +245,49 @@ def test_query_instrumentation
...
+      end
+
+      assert_raises ::Redis::CircuitOpenError do
+        client.get('foo')
+      end
+    end
+
+    time_circuit_half_open = Time.now + ERROR_TIMEOUT + 1
+    Timecop.freeze(time_circuit_half_open) do
+      @proxy.downstream(:latency, latency: (SEMIAN_OPTIONS[:half_open_resource_timeout] * 1000).to_i).apply do
```
sirupsen: You could change this latency to 'infinite' and this test would still pass. You'll want to do an `assert_in_delta` to do what you're trying to do here.

damianthe: Right, but if Redis timeouts out after `half_open_resource_timeout`, then we know the shorter timeout was successfully applied.

If the shorter timeout wasn't successfully applied, this test would fail which means it is doing its job.

lib/semian/redis.rb
damianthe: Didn't consider it. 💃 My internal heuristic "use hash when lots-o-variables" has failed me.

test/redis_test.rb
damianthe: Amazing suggestion.

test/redis_test.rb
sirupsen: Ah yeah, you're right about that. 👍

damianthe: Added a timing check anyway with `Benchmark`

test/mysql2_test.rb
```diff
@@ -312,32 +314,35 @@ def client.raw_ping
...
     end
 
-    Timecop.travel(ERROR_TIMEOUT + 1) do
+    time_circuit_half_open = ERROR_TIMEOUT + 1
+    Timecop.travel(time_circuit_half_open) do
       @proxy.downstream(:latency, latency: 1500).apply do
         assert_raises Mysql2::Error do
           client.query('SELECT 1 + 1;')
         end
       end
```
damianthe: I updated the mysql2 test to more closely match the Redis test. However, in doing so I noticed some odd behaviour. 

I tried to replace the above four lines with a call `assert_mysql_timeout_in_delta` for a timeout of 1s. The timeout raises successfully, but for some reason it takes `4.6 seconds` to timeout, instead of 1s.

I also tried the same thing, but with a lower timeout, to see if the circuit was just in a bad state, and the timeout doesn't occur. This indicates to me that the timeout is indeed `1 second` but for some reason it takes `mysql2` 4.6 seconds to actually fail. 

This is some odd behaviour and is outside the scope of this PR to fix. It would be good to understand why this happens anyway.

sirupsen:
Needs a README change, otherwise LGTM!

PR #258, opened 2019-12-05
Release v0.10.0

damianthe:
Includes the half open resource timeout support for Redis added by https://github.com/Shopify/semian/pull/257.


damianthe:
Done. No new errors in storefront-renderer production with this change.

I also tested this thoroughly locally with the half open timeout on storefront renderer and the timeout is being applied as expected 👍

sirupsen:
👍

PR #263, opened 2020-03-25
Test ruby 2.7 and fix a thread safety issue

casperisfine:
Since https://github.com/ruby/ruby/commit/36da0b3da1aed77e0dffb3f54038f01ff574972b
Ruby checks more often for interupts in various enumerable methods,
so Hash#count for instance is no longer "atomic".
    
Because of this we need to syncrhonize before calling it on @table,
otherwise if multipel threads use the LRUHash, it migth fail with:
    
`can't add a new key into hash during iteration (RuntimeError)`
    
Credits goes to Alan Wu for figuring the bug.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,21 @@
 source 'https://rubygems.org'
+
+if RUBY_VERSION >= '2.7'
+  # Hack to get grpc and it's dependencies to install on Ruby 2.7
+  module BundlerHack
+    def __materialize__
+      if name == 'grpc' || name == 'google-protobuf'
+        Bundler.settings.temporary(force_ruby_platform: true) do
+          super
+        end
+      else
+        super
+      end
+    end
+  end
+  Bundler::LazySpecification.prepend(BundlerHack)
+end
+
 gemspec
 
 group :development, :test do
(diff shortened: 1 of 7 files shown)


Gemfile
```diff
@@ -1,4 +1,20 @@
...
+      if name == 'grpc' || name == 'google-protobuf'
+        Bundler.settings.temporary(force_ruby_platform: true) do
+          super
+        end
+      else
+        super
+      end
+    end
+  end
+  Bundler::LazySpecification.prepend(BundlerHack)
```
sirupsen: 🇫🇷

casperisfine:
I forgot, here's Alan's repro script: https://gist.github.com/XrXr/851220a46bf7c8459e633e397da7191b

In theory not all methods have to be synced, only `.count`, but might as well be safe.

lib/semian/lru_hash.rb
```diff
@@ -5,8 +5,6 @@ class LRUHash
   # everytime we set a new resource. A default window of
   # 5 minutes will allow empty item to stay in the hash
   # for a maximum of 5 minutes
-  extend Forwardable
-  def_delegators :@table, :size, :count, :empty?, :values
   attr_reader :table
```
XrXr: This is suspicious for thread safety. One could still grab @table with this and do dangerous stuff. Do we need to maintain the public API for this class?

casperisfine: Yeah, we should remove that method. I'll do that.

Issue #245, opened 2019-07-02
Feature Request: Percentage Based Error Thresholds


michaelkipper:
## What

Currently, we express error thresholds as the number of failures (`error_threshold`) in a certain time period (`error_timeout`). After that threshold is reached, we open the circuit, and only close it again after a certain number of successful requests (`success_threshold`) are reached.

This requires intimate knowledge of your request patterns. A more flexible model is to use an error _percentage_ threshold to determine when to open the circuit. Instead of saying _3 failures in 5 seconds_, one might say _over 10% of requests failed_.

## How

Either add a new parameter, `error_percent_threshold` or allow `error_threshold` to be expressed as a percentage (e.g. `"10%"`).

Maintain either a large sliding window of successes and errors to compute percentages, or perhaps a set of counters to reduce the overall size of the windows.

thegedge:
> This requires intimate knowledge of your request patterns

On top of that, it also assumes a happy path. I like a percentage threshold because it could adapt to a shift in traffic that wasn't anticipated (e.g., flash sale). This would work well with our current setup where bulkheads trip circuits, because bulkhead timeouts are more likely during high RPS, which in turn makes it more likely that we'd open circuits in high traffic situations (which may not be desirable).

PR #264, opened 2020-04-02
Error rate circuit breaker

RyanAD:
This PR adds a circuit breaker that opens when the error percentage crosses some threshold (ex, 25% of calls failed in the last 10 seconds), also discussed here: https://github.com/Shopify/semian/issues/245. Ideally trying to solve for the problem mentioned in this comment: https://github.com/Shopify/semian/issues/245#issuecomment-508565807 . We have unpredictable spikes in traffic, which make tuning an absolute # of errors in a time window difficult. 

There are some methods that technically violate the DRY principle, in particular the `acquire` and `maybe_with_half_open_resource_timeout` methods on [lib/semian/error_rate_circuit_breaker.rb](lib/semian/error_rate_circuit_breaker.rb) and [lib/semian/circuit_breaker.rb](lib/semian/circuit_breaker.rb) are the same. I'm open to suggestions on if this should be fixed, and the best way to do so.

This has not been battle-tested in a high volume production environment yet, and I'm still working on documentation. Also working on using Timecop in the new unit tests.

### DIFF
--- a/lib/semian/error_rate_circuit_breaker.rb
+++ b/lib/semian/error_rate_circuit_breaker.rb
@@ -1,0 +1,167 @@
+module Semian
+  class ErrorRateCircuitBreaker #:nodoc:
+    extend Forwardable
+
+    def_delegators :@state, :closed?, :open?, :half_open?
+
+    attr_reader :name, :half_open_resource_timeout, :error_timeout, :state, :last_error, :error_percent_threshold,
+                :request_volume_threshold, :success_count_threshold
+
+    def initialize(name, exceptions:, error_percent_threshold:, error_timeout:, window_size:,
+                   request_volume_threshold:, success_threshold:, implementation:, half_open_resource_timeout: nil)
+
+      raise 'error_threshold_percent should be between 0.0 and 1.0 exclusive' unless (0.0001...1.0).cover?(error_percent_threshold)
+
+      @name = name.to_sym
+      @error_timeout = error_timeout
+      @exceptions = exceptions
+      @half_open_resource_timeout = half_open_resource_timeout
+      @error_percent_threshold = error_percent_threshold
+      @last_error_time = nil
+      @request_volume_threshold = request_volume_threshold
+      @success_count_threshold = success_threshold
+
+      @results = implementation::TimeSlidingWindow.new(window_size)
+      @state = implementation::State.new
+
+      reset
+    end
+
+    def acquire(resource = nil, &block)
+      return yield if disabled?
+      transition_to_half_open if transition_to_half_open?
+
+      raise OpenCircuitError unless request_allowed?
+
+      result = nil
+      begin
+        result = maybe_with_half_open_resource_timeout(resource, &block)
+      rescue *@exceptions => error
+        if !error.respond_to?(:marks_semian_circuits?) || error.marks_semian_circuits?
+          mark_failed(error)
+        end
+        raise error
+      else
+        mark_success
+      end
+      result
+    end
+
+    def transition_to_half_open?
+      open? && error_timeout_expired? && !half_open?
+    end
+
+    def request_allowed?
+      closed? || half_open? || transition_to_half_open?
+    end
+
+    def mark_failed(error)
+      push_error(error)
+      if closed?
+        transition_to_open if error_threshold_reached?
+      elsif half_open?
+        transition_to_open
+      end
+    end
+
+    def mark_success
+      @results << true
+      return unless half_open?
+      transition_to_close if success_threshold_reached?
+    end
+
+    def reset
+      @last_error_time = nil
+      @results.clear
+      transition_to_close
+    end
+
+    def destroy
+      @state.destroy
+    end
+
+    # TODO understand what this is used for inside Semian lib
+    def in_use?
+      return false if error_timeout_expired?
+      @results.count(false) > 0
+    end
+
+    private
+
+    def current_time
+      Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    end
+
+    def transition_to_close
+      notify_state_transition(:closed)
+      log_state_transition(:closed)
+      @state.close!
+      @results.clear
+    end
+
+    def transition_to_open
+      notify_state_transition(:open)
+      log_state_transition(:open)
+      @state.open!
+    end
+
+    def transition_to_half_open
+      notify_state_transition(:half_open)
+      log_state_transition(:half_open)
+      @state.half_open!
+      @results.clear
+    end
+
+    def success_threshold_reached?
+      @results.count(true) >= @success_count_threshold
+    end
+
+    def error_threshold_reached?
+      return false if @results.empty? or @results.length < @request_volume_threshold
+      @results.count(false).to_f / @results.length.to_f >= @error_percent_threshold
+    end
+
+    def error_timeout_expired?
+      return false unless @last_error_time
+      current_time - @last_error_time >= @error_timeout
+    end
+
+    def push_error(error)
+      @last_error = error
+      @last_error_time = current_time
+      @results << false
+    end
+
+    def log_state_transition(new_state)
+      return if @state.nil? || new_state == @state.value
+
+      str = "[#{self.class.name}] State transition from #{@state.value} to #{new_state}."
+      str << " success_count=#{@results.count(true)} error_count=#{@results.count(false)}"
+      str << " success_count_threshold=#{@success_count_threshold} error_count_percent=#{@error_percent_threshold}"
+      str << " error_timeout=#{@error_timeout} error_last_at=\"#{@last_error_time}\""
+      str << " name=\"#{@name}\""
+      Semian.logger.info(str)
+    end
+
+    def notify_state_transition(new_state)
+      Semian.notify(:state_change, self, nil, nil, state: new_state)
+    end
+
+    def disabled?
+      ENV['SEMIAN_CIRCUIT_BREAKER_DISABLED'] || ENV['SEMIAN_DISABLED']
+    end
+
+    def maybe_with_half_open_resource_timeout(resource, &block)
+      result =
+          if half_open? && @half_open_resource_timeout && resource.respond_to?(:with_resource_timeout)
+            resource.with_resource_timeout(@half_open_resource_timeout) do
+              block.call
+            end
+          else
+            block.call
+          end
+
+      result
+    end
+  end
+end
--- a/lib/semian/error_rate_circuit_breaker.rb
+++ b/lib/semian/error_rate_circuit_breaker.rb
@@ -5,11 +5,11 @@ module Semian
     def_delegators :@state, :closed?, :open?, :half_open?
 
     attr_reader :name, :half_open_resource_timeout, :error_timeout, :state, :last_error, :error_percent_threshold,
-                :request_volume_threshold, :success_count_threshold
+                :request_volume_threshold, :success_threshold, :time_window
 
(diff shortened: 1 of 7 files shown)


lib/semian/error_rate_circuit_breaker.rb
```diff
@@ -0,0 +1,166 @@
...
+  class ErrorRateCircuitBreaker #:nodoc:
+    extend Forwardable
+
+    def_delegators :@state, :closed?, :open?, :half_open?
+
+    attr_reader :name, :half_open_resource_timeout, :error_timeout, :state, :last_error, :error_percent_threshold,
+                :request_volume_threshold, :success_count_threshold
+
+    def initialize(name, exceptions:, error_percent_threshold:, error_timeout:, window_size:,
+                   request_volume_threshold:, success_threshold:, implementation:, half_open_resource_timeout: nil)
```
damianthe, line 11: nit: The name `window_size` generally implies "integer". Like Array#size. How about renaming it to `window_time` like you have  in `TimeSlidingWindow`?

lib/semian/error_rate_circuit_breaker.rb
```diff
@@ -0,0 +1,166 @@
...
+    end
+
+    def request_allowed?
+      closed? || half_open? || transition_to_half_open?
+    end
+
+    def mark_failed(error)
+      push_error(error)
+      if closed?
+        transition_to_open if error_threshold_reached?
```
damianthe, line 61: When `window_size`  > `error_timeout` the window can leak failed requests from the last time the circuit was closed. 
Therefore prematurely opening the circuit again. What do you think about clearing the `@window` when the state transitions from closed  to open and half-open to open?

lib/semian/error_rate_circuit_breaker.rb
```diff
@@ -0,0 +1,166 @@
+module Semian
+  class ErrorRateCircuitBreaker #:nodoc:
```
damianthe, line 2: I don't think this will work as well you think it will.

The error condition we are trying to dampen with a circuit breaker is a service timeout. Timeouts inherently take a long time so each timeout will be consuming a large portion of the allocated `window_size` to measure the percentage in.

So what does that mean? It means a statement like "Open the circuit if 90% of the requests are failed", doesn't actually work intuitively. In fact, if configured incorrectly, the circuit breaker will rarely ever open when you want it too.

For example, consider a `window_size` of 100 seconds, timeout of 10s, and average service response time is 50ms. If you're going for 90% of requests failed, it wont trigger after the intuitively feeling 9 timeouts. Instead, it will fail after 10 (which is 100%) because the 50ms service requests will fill any gap between timeouts with dozens of successful requests.

This rule extends for most percentages, even with a value of 1%, we're still not getting a behaviour we expect.

Based on https://github.com/Shopify/semian/issues/245, the goal of this PR is to conceptually simplify the configuration. I don't think that has been achieved. This still requires a deep understanding of how the circuit break works to configure correctly.

lib/semian/error_rate_circuit_breaker.rb
```diff
@@ -0,0 +1,166 @@
...
+      @state.half_open!
+      @window.clear
+    end
+
+    def success_threshold_reached?
+      @window.count(true) >= @success_count_threshold
+    end
+
+    def error_threshold_reached?
+      return false if @window.empty? or @window.length < @request_volume_threshold
```
damianthe, line 119: `request_volume_threshold` is an interesting parameter because its desired value varies depending on the volume of requests.

If the volume of requests happening in window_size ever drops bellow `request_volume_threshold`, the circuit will never open.

lib/semian/error_rate_circuit_breaker.rb
damianthe: An alternative could be to measure the response time of the service requests and calculate the percentage based on time spent making successful and failed requests rather than request count.

damianthe (changes requested):
I really like the concept of this PR at a high level. It would be amazing to have a more intuitive way to configure the circuit breaker. 

However, I don't think that this change has actually makes it more intuitive (see comments below). I'm open to alternative approaches or suggestions. Please let me know what you think.

lib/semian/error_rate_circuit_breaker.rb
RyanAD: 👍

lib/semian/error_rate_circuit_breaker.rb
RyanAD: Hi @damianthe. I agree, and it's something I've been mulling over to determine how to solve this. I'm coming from having used Hystrix, and they solve this by having a separate thread that monitors how long calls are taking. That approach is difficult in a multiprocess Ruby environment. I like the idea of tracking time spent in calls. While there may not be a good way to interrupt a call, we could at least track how many are effectively timing out. Let me give that some thought and see what I can come up with.

lib/semian/error_rate_circuit_breaker.rb
RyanAD: Good catch. I can look at solving this as part of the timeout issue discussed above.

RyanAD:
> I really like the concept of this PR at a high level. It would be amazing to have a more intuitive way to configure the circuit breaker.
> 
> However, I don't think that this change has actually makes it more intuitive (see comments below). I'm open to alternative approaches or suggestions. Please let me know what you think.



> I really like the concept of this PR at a high level. It would be amazing to have a more intuitive way to configure the circuit breaker.
> 
> However, I don't think that this change has actually makes it more intuitive (see comments below). I'm open to alternative approaches or suggestions. Please let me know what you think.

I was talking about this with a colleague, and we both agreed it would be better to specifically specify which circuit breaker implementation you want instead of determining based on the options passed in. I plan on making that change. I'm currently working on other things at the moment, but should have time to cycle back on this in the next week.

lib/semian/error_rate_circuit_breaker.rb
RyanAD: `def transition_to_half_open` is clearing `@window`. I'm not sure that we need to clear it when the state transitions to open as well, but I don't think it will hurt.

lib/semian/error_rate_circuit_breaker.rb
RyanAD: That's intentional, I modeled it after this property in Hystrix: https://github.com/Netflix/Hystrix/wiki/Configuration#circuitBreaker.requestVolumeThreshold

https://stackoverflow.com/a/38525660

lib/semian/error_rate_circuit_breaker.rb
RyanAD: You bring up a good example. I was picturing a multithreaded environment, and in that scenario 9 timeouts across 4 threads using a 100 second window would not be significant enough to open the circuit.

I like your idea of tracking time spent making successful vs failed requests. I'll work on an implementation that calculates that and will open when that crosses the threshold. I'll leave the ability to open based on % of failed calls as well, I think that is still useful in the scenario where the service is completely down. If that happens we don't spend much time making failed calls, but we also don't want to have 100s of failed calls in the 100 second window before opening the circuit.

I think having both gives flexibility to cover both scenarios.

Thank you for the feedback!

damianthe: Awesome!

> If that happens we don't spend much time making failed calls, but we also don't want to have 100s of failed calls in the 100 second window before opening the circuit.

I'm not sure I fully understand. If the circuit is completely down, it will be continuously in the half open state and all the calls made will be failed calls. Where does the % of failed calls circuit open  solution come into play here?

lib/semian/error_rate_circuit_breaker.rb
damianthe: Interesting. I can see how this is useful now. If you configure it to be above  `time_window / ( timeout * request_volume_threshold)` then it acts as a damper for the problem I was describing in https://github.com/Shopify/semian/pull/264#discussion_r406299735

lib/semian/error_rate_circuit_breaker.rb
RyanAD: > Awesome!
> 
> > If that happens we don't spend much time making failed calls, but we also don't want to have 100s of failed calls in the 100 second window before opening the circuit.
> 
> I'm not sure I fully understand. If the circuit is completely down, it will be continuously in the half open state and all the calls made will be failed calls. Where does the % of failed calls circuit open solution come into play here?

Consider your example of a circuit breaker with a time window  of 100 seconds, where the first 50 seconds (50%) of calls execute successfully. Now say the resource (redis) goes down completely (process dead), you'll now get near instantaneous connection refused errors. Those errors don't take up much time, and therefore contribute very little to the time spent in error. Therefore the circuit may have 100s or 1000s of errors before opening. In this example, you would have 40 seconds of near instant errors. 

I put a failing unit test in showing this example https://github.com/Shopify/semian/pull/264/files#diff-01dd7217b9f4b90e7bfd8d6c1e28686bR116

This is less of an issue with shorter time windows.

RyanAD:
> I really like the concept of this PR at a high level. It would be amazing to have a more intuitive way to configure the circuit breaker.
> 
> However, I don't think that this change has actually makes it more intuitive (see comments below). I'm open to alternative approaches or suggestions. Please let me know what you think.

Thank you for the suggestions. I just pushed an implementation that calculates the time spent in error vs success. Please let me know your thoughts.

test/time_sliding_window_test.rb
```diff
@@ -0,0 +1,74 @@
+require 'test_helper'
+
+class TestTimeSlidingWindow < Minitest::Test
+  def setup
+    @sliding_window = ::Semian::ThreadSafe::TimeSlidingWindow.new(0.5, -> { Time.now.to_f * 1000 }) # Timecop doesn't work with a monotonic clock
+    @sliding_window.clear
+  end
```
damianthe, line 7: I would throw a `Timecop.freeze` in the setup as well as a `Timecop.return` in the teardown to prevent any test flakyness caused by shared machines executing the tests.

damianthe (commented):
Really nice work. I like the direction this is heading. Just a few minor comments

lib/semian/error_rate_circuit_breaker.rb
RyanAD: I agree. I removed the test for now and will leave it as is.

damianthe (approved):
Looks good to me. Thank you for adding this feature! Feel free to keep me posted on how it works out for you  in a live production environment. 👍

RyanAD:
> 

We've been successfully using this within Instacart for a while now. Let me confirm we don't have any changes that should be incorporated here. I'll follow up by the end of next week.

nap:
@miry is this still valid? What's missing, beside the conflict, to push this out?

lib/semian/error_rate_circuit_breaker.rb
```diff
@@ -0,0 +1,191 @@
...
+      str << " name=\"#{@name}\""
+      Semian.logger.info(str)
+    end
+
+    def notify_state_transition(new_state)
+      Semian.notify(:state_change, self, nil, nil, state: new_state)
+    end
+
+    def disabled?
+      ENV['SEMIAN_CIRCUIT_BREAKER_DISABLED'] || ENV['SEMIAN_DISABLED']
```
miry, line 175: Requires rethinking of changes in master branch. Idea to not use ENV inside business logic, only during the configuration phase.

yathi:
Any updates on this work?

PR #266, opened 2020-05-12
Fix positional/keyword arguments deprecation warnings for Ruby 2.7+

DougEdey:
After updating an app to Ruby 2.7, I figured I should contribute to Semian by getting rid of these Deprecation warnings.

You can see these in previous [CI Runs](https://travis-ci.org/github/Shopify/semian/jobs/669129103#L357-L366) for Semian

I avoided doing any autoformatting/linting on these files in order to ensure that I applied the minimal set of changes

In Ruby 3.0, there's a change for Positional & Keyword arguments, you can read all the background [here](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/) as it will do a much better job of explaining it than I will

~I went with the recommended solution of adding the `ruby2_keywords` gem [source](https://github.com/ruby/ruby2_keywords) and [rubygems](https://rubygems.org/gems/ruby2_keywords)~

After discussion with @casperisfine I decided to fix the warnings rather than work around them, there's some warning upstream in grpc still but that'll be an upstream PR by someone:

```
/usr/local/bundle/gems/grpc-1.28.0/src/ruby/lib/grpc/generic/service.rb:175: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/workspace/lib/semian/grpc.rb:87: warning: The called method `client_streamer' is defined here
./usr/local/bundle/gems/grpc-1.28.0/src/ruby/lib/grpc/generic/service.rb:170: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/workspace/lib/semian/grpc.rb:82: warning: The called method `request_response' is defined here
.../usr/local/bundle/gems/grpc-1.28.0/src/ruby/lib/grpc/generic/service.rb:180: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/workspace/lib/semian/grpc.rb:92: warning: The called method `server_streamer' is defined here
..../usr/local/bundle/gems/grpc-1.28.0/src/ruby/lib/grpc/generic/service.rb:185: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/workspace/lib/semian/grpc.rb:97: warning: The called method `bidi_streamer' is defined here
```

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM ruby:2.6.3
+FROM ruby:2.7.1
 
 # Avoid warnings by switching to noninteractive
 ENV DEBIAN_FRONTEND=noninteractive
--- a/lib/semian/resource.rb
+++ b/lib/semian/resource.rb
@@ -4,8 +4,8 @@ module Semian
 
     class << Semian::Resource
       # Ensure that there can only be one resource of a given type
-      def instance(*args)
-        Semian.resources[args.first] ||= ProtectedResource.new(args.first, new(*args), nil)
+      def instance(name, **kwargs)
+        Semian.resources[name] ||= ProtectedResource.new(name, new(name, **kwargs), nil)
       end
     end
 
--- a/test/resource_test.rb
+++ b/test/resource_test.rb
@@ -117,7 +117,7 @@ class TestResource < Minitest::Test
 
   def test_register_with_neither_quota_nor_tickets_raises
     assert_raises ArgumentError do
-      create_resource :testing
+      Semian::Resource.new(:testing)
     end
   end
 
@@ -490,9 +490,9 @@ class TestResource < Minitest::Test
     assert_equal 0, timeouts
   end
 
-  def create_resource(*args)
+  def create_resource(name, **kwargs)
     @resources ||= []
-    resource = Semian::Resource.new(*args)
+    resource = Semian::Resource.new(name, **kwargs)
     @resources << resource
     resource
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# unrealeased
+
+* Fix: Positional/Keyword arguments deprecations warning for Ruby 2.7
+
 # v0.10.1
 
 * Fix: thread safety bug on Ruby 2.7. (#263)
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -287,7 +287,7 @@ module Semian
     Resource.new(name, tickets: options[:tickets], quota: options[:quota], permissions: permissions, timeout: timeout)
   end
 
-  def require_keys!(required = [], **options)
+  def require_keys!(required, options)
     diff = required - options.keys
     unless diff.empty?
       raise ArgumentError, "Missing required arguments for Semian: #{diff}"
--- a/lib/semian/grpc.rb
+++ b/lib/semian/grpc.rb
@@ -79,22 +79,22 @@ module Semian
       raw_semian_options.nil?
     end
 
-    def request_response(*)
+    def request_response(*, **)
       return super if disabled?
       acquire_semian_resource(adapter: :grpc, scope: :request_response) { super }
     end
 
-    def client_streamer(*)
+    def client_streamer(*, **)
       return super if disabled?
       acquire_semian_resource(adapter: :grpc, scope: :client_streamer) { super }
     end
 
-    def server_streamer(*)
+    def server_streamer(*, **)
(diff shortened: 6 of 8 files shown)


casperisfine:
IMO it's the callers that should properly pass keyword arguments.

casperisfine:
e.g.

```
semian_1     | ../app/lib/semian.rb:251: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
semian_1     | /app/lib/semian.rb:290: warning: The called method `require_keys!' is defined here
```

Rather than mark `require_keys!` with `ruby2keywords`, the proper fix it to add `**`:

```ruby
    require_keys!([:success_threshold, :error_threshold, :error_timeout], **options)
```

casperisfine:
Actually that in specific case, it's a bit silly to take `**options` as it's only ever used as a hash. So changing the signature for `require_keys!(required, options)` would be just as good if not better.

DougEdey:
@casperisfine When I read the doc I interpreted it as `that won't work`, but I tried that approach and it works fine on Ruby 2.6, Updating PR

casperisfine:
Yeah, this whole change is a bit hard to understand at first. But when you get it it makes much more sense than the old behavior.

.devcontainer/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ruby:2.6.3
+FROM ruby:2.7.1
```
DougEdey, line 1: Local Dev was not running with the same Ruby version as CI, so I fixed that here

lib/semian/resource.rb
```diff
@@ -5,7 +5,7 @@ class Resource #:nodoc:
     class << Semian::Resource
       # Ensure that there can only be one resource of a given type
       def instance(*args)
-        Semian.resources[args.first] ||= ProtectedResource.new(args.first, new(*args), nil)
+        Semian.resources[args.first] ||= ProtectedResource.new(args.first, new(args.first, **args[1]), nil)
```
DougEdey, line 8: I'm unsure if there's a better way to deal with this

DougEdey:
@casperisfine Updated code, Description, and title

lib/semian/resource.rb
casperisfine: Oh, I missed that one as well. 

Same thing, `*args, **kwargs` is the way to go, unless I'm missing something.

DougEdey: Nope, i didn't know you could do that, switched both over

test/resource_test.rb
```diff
@@ -492,7 +492,7 @@ def test_multiple_register_with_fork
 
   def create_resource(*args)
     @resources ||= []
-    resource = Semian::Resource.new(*args)
+    resource = Semian::Resource.new(args.first, **args[1])
```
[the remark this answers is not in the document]
DougEdey: I didn't know you could do that :) Fixed

lib/semian/resource.rb
```diff
@@ -4,8 +4,8 @@ class Resource #:nodoc:
 
     class << Semian::Resource
       # Ensure that there can only be one resource of a given type
-      def instance(*args)
-        Semian.resources[args.first] ||= ProtectedResource.new(args.first, new(*args), nil)
+      def instance(*args, **kwargs)
+        Semian.resources[args.first] ||= ProtectedResource.new(args.first, new(args.first, **kwargs), nil)
```
casperisfine, line 8: There's still something that isn't quite right here.

You grab all positional arguments with `*args` but only pass the first one.

So `instance(1, 2, 3, foo: 42)` works, but `2, 3` are simply ignored.

If you know that instance only takes a single positional argument, then the proper signature is `(name, **kwargs)`

test/resource_test.rb
```diff
@@ -490,9 +490,9 @@ def test_multiple_register_with_fork
     assert_equal 0, timeouts
   end
 
-  def create_resource(*args)
+  def create_resource(*args, **kwargs)
     @resources ||= []
-    resource = Semian::Resource.new(*args)
+    resource = Semian::Resource.new(args.first, **kwargs)
```
casperisfine, line 495: same here. `*args` not `args.first`.

lib/semian/resource.rb
DougEdey: That's fair, this was just what the old code had, I'll fix it now

test/resource_test.rb
DougEdey: I prefer the `name, **kwargs` format here too

PR #271, opened 2020-08-21
Fix link to 'Release It' book in README.md

scottjanderson:
The link to the first edition is broken, so we should update it to the second edition.


scottjanderson:
I've signed the CLA!

PR #273, opened 2020-10-08
Update README monitoring example

### LINKED ISSUE #253: Broken subscribe interface
Following the [monitoring instructions](https://github.com/Shopify/semian/blob/169aa759639973a6e2d0475e1a94a221d20c7660/README.md#monitoring) in the readme:

```ruby
# `event` is `success`, `busy`, `circuit_open`.
# `resource` is the `Semian::Resource` object
# `scope` is `connection` or `query` (others can be instrumented too from the adapter)
# `adapter` is the name of the adapter (mysql2, redis, ..)
Semian.subscribe do |event, resource, scope, adapter|
  StatsD.increment("Shopify.#{adapter}.semian.#{event}", 1, tags: [
    "resource:#{resource.name}",
    "total_tickets:#{resource.tickets}",
    "type:#{scope}",
  ])
end
```

Results in an Error:

```
NoMethodError: undefined method `tickets' for #<Semian::CircuitBreaker:0x0000558e2ec77ba0>
```

It seems the `resource` is a `Semian::CircuitBreaker`, and not a `Semian::Resource`!

Introduced by: #238

Semian Version: 0.8.9

spike01:
Closes #253 

The old example made an assumption that the resource in question is a bulkhead (referencing tickets), when in some cases the resource will only be protected by a circuit breaker. This feels like a minimal example that will give ALL the event types, with reasonable tags. This is probably too many metrics (for most uses, `circuit_open` and `busy` is probably enough...) but is better than broken?

Tested this locally and it seems to work

### DIFF
--- a/README.md
+++ b/README.md
@@ -648,11 +648,11 @@ Semian internals. For example to instrument just events with
 # `scope` is `connection` or `query` (others can be instrumented too from the adapter)
 # `adapter` is the name of the adapter (mysql2, redis, ..)
 Semian.subscribe do |event, resource, scope, adapter|
-  StatsD.increment("Shopify.#{adapter}.semian.#{event}", 1, tags: [
-    "resource:#{resource.name}",
-    "total_tickets:#{resource.tickets}",
-    "type:#{scope}",
-  ])
+  StatsD.increment("semian.#{event}", 1, tags: {
+    resource: resource.name,
+    adapter: adapter,
+    type: scope,
+  })
 end
 ```
 


README.md
```diff
@@ -648,11 +648,11 @@ Semian internals. For example to instrument just events with
 # `scope` is `connection` or `query` (others can be instrumented too from the adapter)
 # `adapter` is the name of the adapter (mysql2, redis, ..)
 Semian.subscribe do |event, resource, scope, adapter|
-  StatsD.increment("Shopify.#{adapter}.semian.#{event}", 1, tags: [
-    "resource:#{resource.name}",
-    "total_tickets:#{resource.tickets}",
-    "type:#{scope}",
-  ])
+  StatsD.increment("semian.#{event}", 1, tags: {
```
tgwizard, line 651: Is the `1` required here, or is it an optional default?

tgwizard (approved):

README.md
bmansoob: Looks like it is default, at least in the gem we use: https://github.com/Shopify/statsd-instrument/blob/master/lib/statsd/instrument/client.rb#L200

bmansoob (approved):

PR #274, opened 2020-10-13
Try to trigger compaction bug

XrXr:


pushrax:
This seems to have started triggering much more often in Core after I shipped a change that prepends to Semian::TimeoutError. Could try doing that here.

XrXr:
It did trigger in the Travis build.

PR #275, opened 2020-10-13
Fixes for GC.compact

XrXr:
Without doing this the GC could move the objects returned by
rb_const_get and make Semian's globals point to unexpected objects on
the heap.

No tests here because usually compaction bugs are hard to trigger reliably. I'm trying to trigger them in https://github.com/Shopify/semian/pull/274 though. If it goes well maybe we should include that as a best-effort regression test.

### DIFF
--- a/ext/semian/semian.c
+++ b/ext/semian/semian.c
@@ -22,12 +22,14 @@ void Init_semian()
    * Represents a Semian error that was caused by an underlying syscall failure.
    */
   eSyscall = rb_const_get(cSemian, rb_intern("SyscallError"));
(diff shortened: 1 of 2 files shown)


rafaelfranca (approved):

paracycle (approved):

pushrax (approved):
I feel like there must be quite a lot of code in the wild that has this bug...

casperisfine:
> I feel like there must be quite a lot of code in the wild that has this bug...

Yes, plenty of extensions have this issue.

casperisfine (approved):

PR #276, opened 2020-11-04
Allow MySQL queries through Semian regardless of position

insom:
The current Semian behaviour will let [certain allow-listed MySQL queries to go through](https://github.com/Shopify/semian/pull/60), even if we don't have tickets -- but the existing matches are incompatible with using the [Marginalia gem set to `prepend_comment`](https://github.com/basecamp/marginalia/blob/7bf7ec89e18cd4849cd43c24ed931d341363665c/lib/marginalia.rb#L52).

This PR adds extra regular expressions to catch this particular case. We still need to anchor on either `\A` or `\z` because otherwise we would match allow-listed words when they appeared in the body of queries.

### DIFF
--- a/lib/semian/mysql2.rb
+++ b/lib/semian/mysql2.rb
@@ -37,12 +37,9 @@ module Semian
     DEFAULT_PORT = 3306
 
     QUERY_WHITELIST = Regexp.union(
-      /\A\s*ROLLBACK/i,
-      /\A\s*COMMIT/i,
-      /\A\s*RELEASE\s+SAVEPOINT/i,
-      /ROLLBACK[\s;]*\z/i,
-      /COMMIT[\s;]*\z/i,
-      /RELEASE\s+SAVEPOINT[\s;]*\z/i,
+      /\A(?:\/\*.*?\*\/)?\s*ROLLBACK/i,
+      /\A(?:\/\*.*?\*\/)?\s*COMMIT/i,
+      /\A(?:\/\*.*?\*\/)?\s*RELEASE\s+SAVEPOINT/i,
     )
 
     # The naked methods are exposed as `raw_query` and `raw_connect` for instrumentation purpose
--- a/lib/semian/mysql2.rb
+++ b/lib/semian/mysql2.rb
@@ -40,6 +40,9 @@ module Semian
       /\A\s*ROLLBACK/i,
       /\A\s*COMMIT/i,
       /\A\s*RELEASE\s+SAVEPOINT/i,
+      /ROLLBACK[\s;]*\z/i,
+      /COMMIT[\s;]*\z/i,
+      /RELEASE\s+SAVEPOINT[\s;]*\z/i,
     )
 
     # The naked methods are exposed as `raw_query` and `raw_connect` for instrumentation purpose
--- a/test/mysql2_test.rb
+++ b/test/mysql2_test.rb
@@ -200,6 +200,16 @@ class TestMysql2 < Minitest::Test
     end
   end
 
+  def test_semian_allows_rollback_with_marginalia
+    client = connect_to_mysql!
+
+    client.query('START TRANSACTION;')
+
+    Semian[:mysql_testing].acquire do
+      client.query('/*foo:bar*/ ROLLBACK;')
+    end
+  end
+
   def test_semian_allows_commit
     client = connect_to_mysql!
 


lib/semian/mysql2.rb
```diff
@@ -40,6 +40,9 @@ module Mysql2
       /\A\s*ROLLBACK/i,
       /\A\s*COMMIT/i,
       /\A\s*RELEASE\s+SAVEPOINT/i,
+      /ROLLBACK[\s;]*\z/i,
```
casperisfine, line 43: Wouldn't it be better to just ignore the comment?, e.g. `\A(?:\/\*.*?\*\/)?\s* ROLLBACK`.

insom:
👍  -- originally I thought that using a non-greedy `.*?` would be more overhead than a specific match against beginning and end -- but I didn't benchmark it and actually it's much better to only have three complex regex than six simple ones.

Updated the PR.

```
local:~/src/github.com/Shopify/semian/lib/semian % cat aaron.rb
require 'benchmark'

ANCHORED = Regexp.union(
  /\A\s*ROLLBACK/i,
  /ROLLBACK[\s;]*\z/i,
)

UNANCHORED = Regexp.union(
  /\A(?:\/\*.*?\*\/)?\s* ROLLBACK/i
)

n = 100_000
Benchmark.bm(5) do |x|
  x.report("Anchored:") { n.times { ANCHORED =~ '/*foo:' + ('1' * 1000) + 'bar*/ ROLLBACK' } }
  x.report("Unanchored:") { n.times { UNANCHORED =~ '/*foo:' + ('1' * 1000) + 'bar*/ ROLLBACK' } }
end
local:~/src/github.com/Shopify/semian/lib/semian % ruby aaron.rb
            user     system      total        real
Anchored:  2.565520   0.010094   2.575614 (  2.577190)
Unanchored:  1.313081   0.005984   1.319065 (  1.320022)
```

casperisfine:
I didn't know perf was a concern  here, but if it is, here's an improved benchmark:

```ruby
require 'benchmark/ips'

ANCHORED = Regexp.union(
  /\A\s*COMMIT/i,
  /COMMIT[\s;]*\z/i,
  /\A\s*RELEASE\s+SAVEPOINT/i,
  /RELEASE\s+SAVEPOINT[\s;]*\z/i,
  /\A\s*ROLLBACK/i,
  /ROLLBACK[\s;]*\z/i,
)

UNANCHORED = Regexp.union(
  /\A(?:\/\*.*?\*\/)?\s*COMMIT/i,
  /\A(?:\/\*.*?\*\/)?\s*RELEASE\s+SAVEPOINT/i,
  /\A(?:\/\*.*?\*\/)?\s*ROLLBACK/i,
)

COMBINED = /\A(?:\/\*.*?\*\/)?\s*(COMMIT|RELEASE\s+SAVEPOINT|ROLLBACK)/

string = '/*foo:' + ('1' * 1000) + 'bar*/ ROLLBACK'

Benchmark.ips do |x|
  x.report("Anchored:") { ANCHORED.match?(string) }
  x.report("Unanchored:") { UNANCHORED.match?(string) }
  x.report("Combined:") { COMBINED.match?(string) }
  x.compare!
end
```

- Use a static string so we don't bench string concatenation
- Use benchmark/ips so we don't bench the iterations
- Use `Regexp.match?` as we don't care about the matched values
- Added a third method that share the regexp prefix

```
           Anchored:     2.203k i/100ms
         Unanchored:     3.077k i/100ms
           Combined:    11.218k i/100ms
           Anchored:     22.589k (± 2.6%) i/s -    114.556k in   5.074815s
         Unanchored:     30.559k (± 4.5%) i/s -    153.850k in   5.046568s
           Combined:    109.344k (± 4.5%) i/s -    549.682k in   5.037894s

Comparison:
           Combined::   109343.7 i/s
         Unanchored::    30558.6 i/s - 3.58x  (± 0.00) slower
           Anchored::    22589.3 i/s - 4.84x  (± 0.00) slower
```

insom:
> I didn't know perf was a concern here, but if it is, here's an improved benchmark

Thank you for this. To be honest, it's probably not that much of a concern -- I was thinking that this code path gets hit on every query -- but compared to the network round-trip to MySQL or ProxySQL, the three separate regexes are worth it for code clarity.

casperisfine:
Yeah, I don't think it's a big deal either. But if it's only a matter of clarity, you cal also decompose the regexp:

```ruby
COMMENT = /(?:\/\*.*?\*\/)?/
STATEMENTS = Regexp.union(
  /COMMIT/,
  /RELEASE\s+SAVEPOINT/,
  /ROLLBACK/,
)
/\A#{COMMENT}\s*#{STATEMENTS}/i
```

Anyway, I'm good with any solution.

casperisfine (approved):

bmansoob (approved):

insom:
Waiting 2+ hours for Travis is nuts. I've verified locally using the same container:

```
can't be called from trap context
..............................

Finished in 90.615165s, 2.2513 runs/s, 6.0034 assertions/s.

204 runs, 544 assertions, 0 failures, 0 errors, 0 skips
Running rubocop
Running rubocop failed
```

(Rubocop failing is expected). Merging 🚢

PR #278, opened 2020-12-01
Use global registered_workers count to calculate tickets instead of local per resource lazy loaded count

bibstha:
In production, we've seen that when using quota and registered_workers count is low, bulkheads are quite frequent.
Specially with an adapter like mysql where we increase registered_workers lazily (only when we connect, ping or query) this is an issue.

With this PR, Semian keeps track of overall number of processes (global registered worker count) in a host once. Then for any Semian resource configuration with `is_global: true`, it uses the global registered worker count to calculate tickets instead of resource specific registered_worker count.

## How to use global workers?

```ruby
# n separate process will call it once each, therefore global worker count = n
10.times { Semian.register_global_worker }

resource = Semian.register(:mysql_shard0, quota: 0.5, is_global: true)
puts resource.tickets
# this will print 5 instead of 1
```

## Approach

Instead of working with a dedicated semaphore to manage global worker count, I decided to use a global worker, which is an singleton instance of general resource. This saves us from complexity and validate the usefulness of this approach quicker.

### DIFF
--- a/ext/semian/resource.c
+++ b/ext/semian/resource.c
@@ -12,6 +12,9 @@ check_quota_arg(VALUE quota);
 static int
 check_tickets_arg(VALUE tickets);
 
+static int
+check_is_global_arg(VALUE is_global);
+
 static long
 check_permissions_arg(VALUE permissions);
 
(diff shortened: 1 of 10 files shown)


alexrs:
This approach makes sense to me. This is actually the behavior I would expect for MySQL circuit breakers.

PR #280, opened 2020-12-10
Add Semian.namespace to flush all resources

casperisfine:
### Context
I'm trying to rollout a major infrastructure change on an app, and among other things it involve a changed process UID. Because of this, during rollout I get some Semian permission errors:

```
gems/semian-0.10.6/lib/semian/resource.rb:14:in `initialize_semaphore': semget() failed, errno: 13 (Permission denied) (Semian::SyscallError)
```

Because of this I'd like to change the key of all our semaphores. 

### Feature

`Semian.namespace` can be assigned a string that will be prefixed to all semaphores (before the hashing)

### DIFF
--- a/test/resource_test.rb
+++ b/test/resource_test.rb
@@ -415,6 +415,26 @@ class TestResource < Minitest::Test
     end
   end
 
+  def test_namespace
+    previous_namespace = Semian.namespace
+
+    resource = create_resource :testing, permissions: 0o600, tickets: 1
+    key = resource.key
+    Semian.destroy(:testing)
+
+    resource = create_resource :testing, permissions: 0o600, tickets: 1
+    assert_equal key, resource.key
+    Semian.destroy(:testing)
+
+    Semian.namespace = "testing"
+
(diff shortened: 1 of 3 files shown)


DazWorrall (approved):

test/resource_test.rb
```diff
@@ -415,6 +415,26 @@ def test_permissions
...
 
+  def test_namespace
+    previous_namespace = Semian.namespace
+
+    resource = create_resource :testing, permissions: 0o600, tickets: 1
+    key = resource.key
+    Semian.destroy(:testing)
+
+    resource = create_resource :testing, permissions: 0o600, tickets: 1
+    assert_equal key, resource.key
```
hkdsun, line 426: Had to look at why you're not asserting a match / equality with a human-readable string.. Looks like `resource.key` is hex string of a SHA digest 👍

hkdsun (approved):
LGTM 👍 

You'll bump the version in a separate commit?

casperisfine:
> You'll bump the version in a separate commit?

Yes.

PR #281, opened 2020-12-10
Switch to GitHub Actions for CI

DazWorrall:
This is a direct port of the current approach - running the tests inside a container orchestrated by the host. We can probably improve things by switching to using [job containers](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer) directly, and using that to persist bundler caches between builds. For now I only wanted to switch away from Travis - the wait time for jobs is getting pretty excruciating (#280 has been waiting for _hours_). Should I also remove the `docker-compose.ci.yml` file, or is that used for local dev sometimes?

Side note: boy, these tests are flaky, huh?


casperisfine (approved):
MVP

casperisfine:
> Side note: boy, these tests are flaky, huh?

There's a `SKIP_FLAKY_TESTS` variable for CI

casperisfine:
https://github.com/Shopify/semian/blob/master/README.md#everything-else

DazWorrall:
Yeah I saw that @casperisfine, but it is intentionally not used in CI:

> Run with SKIP_FLAKY_TESTS=true to skip flaky tests **(CI runs all tests)**

In favour of several retries I guess:

https://github.com/Shopify/semian/blob/6bb3d41c122dddfcb8a6d4681879f5ea7861a3ea/scripts/run_tests.sh#L29-L35

casperisfine:
AH my bad.

Let's merge!

PR #283, opened 2021-01-27
Resolve common misconception about error_threshold in the Readme

damianthe:
I've had a few instances of confusion around the interaction between error_timeout and error_threshold.
This simple update to the read-me will help but I can still see how this interaction is unintuitive. 

Perhaps it's time we also update error_threshold to work indefinitely despite the time between errors, and only reset the counter on success.

### DIFF
--- a/README.md
+++ b/README.md
@@ -441,7 +441,7 @@ all workers on a server.
 
 There are four configuration parameters for circuit breakers in Semian:
 
-* **error_threshold**. The amount of errors to encounter for the worker before
+* **error_threshold**. The amount of errors a worker encounters within error_timeout amount of time before
   opening the circuit, that is to start rejecting requests instantly.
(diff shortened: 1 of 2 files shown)


tgwizard (approved):
LGTM. Perhaps the inline comments on https://github.com/Shopify/semian/blob/17a55985d5ea2344ec7779179c044a82f96d994c/lib/semian.rb#L146-L149 are a bit clearer than the README, and we can update the docs for `error_timeout` too?

damianthe:
@tgwizard Do the docs auto update or do I have to change something special?

tgwizard:
@damianthe I have no idea, unfortunately.

damianthe:
I think it does since the comments are labeled with some seemingly parsable formatting.

PR #285, opened 2021-01-29
Log last error message on circuit breaker state transition

froot:
This PR will help address: https://github.com/Shopify/shopify-proxysql/issues/129

We are seeing elevated semian errors in core whenever we deploy to ProxySQL. I was unable to find any logs in splunk that noted what caused the circuits to open. This is possibly because the application code that receives the `CircuitOpenError` exception is not choosing to log the error message.

We need to be able to see what errors are causing the circuit to open, so I'm logging the `last_error.message` whenever the state is changed to `open`.

### DIFF
--- a/lib/semian/circuit_breaker.rb
+++ b/lib/semian/circuit_breaker.rb
@@ -136,6 +136,10 @@ module Semian
       str << " success_count_threshold=#{@success_count_threshold} error_count_threshold=#{@error_count_threshold}"
       str << " error_timeout=#{@error_timeout} error_last_at=\"#{@errors.last}\""
       str << " name=\"#{@name}\""
+      if new_state == :open && @last_error
+        str << " last_error_message=#{@last_error.message.inspect}"
+      end
+
       Semian.logger.info(str)
     end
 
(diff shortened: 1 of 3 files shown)


casperisfine:
> This PR will help address: ...

No big deal at all, but keep in mind semian is a public project. So it's preferable to describe issues in a more generic way so that external contributors can better understand what's going on.

lib/semian/circuit_breaker.rb
```diff
@@ -136,6 +136,10 @@ def log_state_transition(new_state)
       str << " success_count_threshold=#{@success_count_threshold} error_count_threshold=#{@error_count_threshold}"
       str << " error_timeout=#{@error_timeout} error_last_at=\"#{@errors.last}\""
       str << " name=\"#{@name}\""
+      if new_state == :open && @last_error
+        str << " last_error_message=\"#{@last_error.message}\""
```
casperisfine, line 140: nitpick: `" last_error_message=#{@last_error.message.inspect}"` would probably handle eventual quote escaping better.

casperisfine (approved):
LGTM

hkdsun (commented):
💯

PR #291, opened 2021-04-09
Update README with :state_change instrumentation event

### LINKED ISSUE #290: Update instrumentation docs
Semian instruments state changes in circuit breakers, but the current documentation does not reflect this:

https://github.com/Shopify/semian/blob/7a5adb74eb2565b6a5041d7033c7b871c7d8e9bc/lib/semian/circuit_breaker.rb#L147

```ruby
# `event` is `success`, `busy`, `circuit_open`.
# `resource` is the `Semian::Resource` object
# `scope` is `connection` or `query` (others can be instrumented too from the adapter)
# `adapter` is the name of the adapter (mysql2, redis, ..)
Semian.subscribe do |event, resource, scope, adapter|
  StatsD.increment("semian.#{event}", 1, tags: {
    resource: resource.name,
    adapter: adapter,
    type: scope,
  })
end
```

I might prepare a PR as soon as I find some spare time.

madejejej:
closes #290

### DIFF
--- a/README.md
+++ b/README.md
@@ -643,16 +643,29 @@ Semian internals. For example to instrument just events with
 [`statsd-instrument`][statsd-instrument]:
 
 ```ruby
-# `event` is `success`, `busy`, `circuit_open`.
+# `event` is `:success`, `:busy`, `:circuit_open`, `:state_change`.
 # `resource` is the `Semian::Resource` object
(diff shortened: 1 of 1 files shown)


miry (changes requested):
@madejejej Sorry for really long reply. I checked the current readme state, it is cover the `:state_change`. Can you check if this PR is still required.

PR #299, opened 2021-10-19
Add a new configuration option to override the error threshold rolling window

franciscojaviermejia:
### What are you trying to accomplish?

Semian currently uses the `error_timeout` configuration parameter for two different purposes:

1. **To open circuits:** The error_timeout is the size of the rolling window used to store errors when the circuit is closed. When <error_threshold> errors occur inside the rolling window, the circuit opens.
2. **To keep the circuits closed**: The amount of time in seconds until trying to query the resource again when the circuit opens.

This PR implements an optional configuration parameter named: `error_threshold_timeout`. When set to true, it overrides the `error_timeout` value for scenario 1.

### Why?

To enhance the flexibility of Semian. We are currently forced to update the `error_timeout` to update the size of the rolling window used to open circuits. This is not ideal since reconfiguring the `error_timeout` has consequences.

For instance, to increase the size of the rolling window used to open circuits we need to increase the error_timeout. Increasing the error_timeout has a side-effect of increasing the time taken for opened circuits to close and continue to serve requests. We are adding unnecessary down time to serve the purpose of scenario 1. This side-effect is amplified when the number of resources is high.

What should reviewers focus on? What could go wrong?

1. This PR is intended to be backwards-compatible. Have I missed something?

2. Is there a better approach to solve this problem?

3. Is the name for the new configuration option appropriate?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# v0.11.8
+
+* Feature: Add error_threshold_timeout configuration parameter (#299)
+
 # v0.11.7
 
 * Fix: ECONNRESET won't trigger circuit open for redis (#306)
--- a/lib/semian/version.rb
+++ b/lib/semian/version.rb
@@ -1,3 +1,3 @@
 module Semian
-  VERSION = '0.11.7'
+  VERSION = '0.11.8'
 end


franciscojaviermejia:
cc @celsodantas @casperisfine

Renamed configuration parameter to `error_window_timeout` and updated the README.

celsodantas (approved):

PR #301, opened 2021-11-22
feat: add force disable semian

SomalianIvan:
In opentelemetry we already implement our own retry mechanism with exponential backoff. Hence we want a way to disable semian temporarily for specific code blocks, to avoid seeing custom errors and overhead of semian.

My intuition is that having ability to disable semian for specific code path has usecases.

### DIFF
--- a/lib/semian/net_http.rb
+++ b/lib/semian/net_http.rb
@@ -61,9 +61,9 @@ module Semian
       end
     end
 
-    def initialize(*args, semian_enabled: true)
+    def initialize(*args, semian: true)
       super(*args)
-      @semian_enabled = semian_enabled
+      @semian_enabled = semian
     end
 
     Semian::NetHTTP.reset_exceptions
@@ -80,7 +80,7 @@ module Semian
     end
 
     def disabled?
-      raw_semian_options.nil? || semian_enabled == false
+      raw_semian_options.nil? || @semian_enabled == false
     end
 
     def connect
@@ -107,8 +107,6 @@ module Semian
         self.open_timeout = prev_open_timeout
       end
     end
-
-    attr_accessor :semian_enabled
 
     private
 
--- a/lib/semian/net_http.rb
+++ b/lib/semian/net_http.rb
@@ -61,6 +61,11 @@ module Semian
       end
     end
 
+    def initialize(*args, semian_enabled: true)
+      super(*args)
+      @semian_enabled = semian_enabled
+    end
+
     Semian::NetHTTP.reset_exceptions
 
     def raw_semian_options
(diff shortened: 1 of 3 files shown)


lib/semian/net_http.rb
```diff
@@ -44,7 +44,7 @@ def semian_identifier
     ].freeze # Net::HTTP can throw many different errors, this tries to capture most of them
 
     class << self
-      attr_accessor :exceptions
+      attr_accessor :exceptions, :disable_semian
```
casperisfine, line 47: That's process global, so not thread safe. 

Would make more sense to have it on the instance or something like that.

SomalianIvan: :nod: for sure

casperisfine:
Works for me, but I wonder if it should be part of `Semian::Adapter`, so that it behave the same on all resources.

casperisfine:
I guess it wouldn't be a problem if this wasn't a mutation of an existing instance, but a constructor argument.

Am I correct that in your case `opentelemetry` has its own Net::HTTP connection, right? So what you really want is more `Net::HTTP.new(..., semian: false)` or something like that.

SomalianIvan:
> Am I correct that in your case opentelemetry has its own Net::HTTP connection, right? So what you really want is more Net::HTTP.new(..., semian: false) or something like that.

Yep , but we also need to be able to tell if semian is enabled or not.

casperisfine:
I'm afraid I don't understand, could you elaborate a bit more, with code samplesperhaps?

SomalianIvan:
https://github.com/open-telemetry/opentelemetry-ruby/blob/e099ecaed370211869b1fb26b2a8abf5d0027b97/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb#L63

```
@http = Net::HTTP.new(@uri.host, @uri.port)
@http.semian_disabled = true if @http.respond_to?(:semian_disabled=)
```

I think I want something like this ^^

SomalianIvan:
> You mean you don't need any change to semian?

does the constructor already accept that change?

lib/semian/net_http.rb
```diff
@@ -103,6 +108,8 @@ def with_resource_timeout(timeout)
       end
     end
 
+    attr_accessor :semian_enabled
```
casperisfine, line 111:
```suggestion
    attr_accessor :semian_enabled
```

Let's not allow to change it after the connection was created, because it opens a huge can of worm.

lib/semian/net_http.rb
```diff
@@ -61,6 +61,11 @@ def reset_exceptions
       end
     end
 
+    def initialize(*args, semian: true)
+      super(*args)
+      @semian_enabled = semian
```
robertlaurin, line 66: Would it be strange to give this an attribute accessor?

SomalianIvan: > Would it be strange to give this an attribute accessor so that we can toggle it if needed?

From what @casperisfine mentioned, http object might be shared between threads so toggling it isn't thread safe

lib/semian/net_http.rb
SomalianIvan: 👍

casperisfine (approved):
LGTM. Please squash your commits though.

PR #303, opened 2021-12-02
Fix disable flag

SomalianIvan:
I goofed and earlier approach in #301 doesn't work. https://github.com/ruby/ruby/blob/1f4af993835219efa8feaf76a0b36252028691f1/lib/net/http.rb#L654 - http class has its `self.new` definition. This should patch the correct method, but I am a little bit apprehensive to use `instance_variable_set`

### DIFF
--- a/lib/semian/net_http.rb
+++ b/lib/semian/net_http.rb
@@ -55,12 +55,6 @@ module Semian
       attr_accessor :exceptions
       attr_reader :semian_configuration
 
-      def prepended(base)
-        class << base
-          prepend ClassMethods
-        end
-      end
-
       def semian_configuration=(configuration)
         raise Semian::NetHTTP::SemianConfigurationChangedError unless @semian_configuration.nil?
         @semian_configuration = configuration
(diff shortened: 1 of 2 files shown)


lib/semian/net_http.rb
```diff
@@ -43,10 +43,24 @@ def semian_identifier
...
+
     class << self
       attr_accessor :exceptions
       attr_reader :semian_configuration
 
+      def prepended(base)
+        class << base
+          prepend ClassMethods
+        end
+      end
```
casperisfine, line 62: Not a fan of this, you can just do `Net::HTTP.singleton_class.prepend(Semian::NetHTTP::ClassMethods)`

lib/semian/net_http.rb
```diff
@@ -119,4 +128,4 @@ def handle_error_responses(result)
   end
 end
 
-Net::HTTP.prepend(Semian::NetHTTP)
+Net::HTTP.prepend(Semian::NetHTTP)
```
casperisfine, line 131: This shouldn't have changed, no big deal but...

casperisfine:
> I am a little bit apprehensive to use instance_variable_set

Yeah it's not great but there isn't much alternatives. Instead you can define a private `semian_disabled=` and then `send()` to bypass the private visibility. It's more or less the same in the end.

PR #306, opened 2022-01-10
Don't open the circuit on redis connection reset

MatthewRBruce:
Connect Reset by Peer is a fast failing error and should not cause an open circuit, unfortunately the Redis gem bundles `ECONNRESET` along with a few other errors as `Redis::ConnectionError`.  This PR implements the `marks_semian_circuits?` method in the Redis::ConnectionError class and checks the error message to determine if the error is an ECONNRESET.  If it is then we don't mark this as a semian failure. 

This PR also update version of toxiproxy used in tests to 2.3.0 which supports a connect reset toxic.

### DIFF
--- a/lib/semian/redis.rb
+++ b/lib/semian/redis.rb
@@ -16,6 +16,14 @@ class Redis
     include ::Semian::AdapterError
   end
 
+  class ConnectionError < Redis::BaseConnectionError
+    # A Connection Reset is a fast failure and we don't want to track these errors in 
+    # semian
+    def marks_semian_circuits?
+      message != "Connection lost (ECONNRESET)"
+    end
+  end
+
   ResourceBusyError = Class.new(SemianError)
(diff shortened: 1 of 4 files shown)


Mangara (approved):

lib/semian/redis.rb
```diff
@@ -16,6 +16,14 @@ class OutOfMemoryError < Redis::CommandError
     include ::Semian::AdapterError
   end
 
+  class ConnectionError < Redis::BaseConnectionError
+    # A Connection Reset is a fast failure and we don't want to track these errors in 
+    # semian
+    def marks_semian_circuits?
+      return message != "Connection lost (ECONNRESET)"
```
casperisfine, line 23:
```suggestion
      message != "Connection lost (ECONNRESET)"
```

casperisfine (approved):

PR #309, opened 2022-02-28
Support new @raw_connection ivar name

nvasilevski:
Rails change - https://github.com/rails/rails/commit/d86fd6415c0dfce6fadb77e74696cf728e5eb76b

Next version of Rails will use a different name for the `@connection` variable and `semian` needs to support this
Since we can't rely on any version check at this moment so I used `instance_variable_defined` to determine which variable to use

Is there a better way to support both ivars?

### DIFF
--- a/lib/semian/rails.rb
+++ b/lib/semian/rails.rb
@@ -2,6 +2,8 @@ require 'active_record/connection_adapters/abstract_adapter'
 
 class ActiveRecord::ConnectionAdapters::AbstractAdapter
   def semian_resource
-    @connection.semian_resource
+    # support for https://github.com/rails/rails/commit/d86fd6415c0dfce6fadb77e74696cf728e5eb76b
+    connection = instance_variable_defined?(:@raw_connection) ? @raw_connection : @connection
+    connection.semian_resource
   end
 end


lib/semian/rails.rb
```diff
@@ -2,6 +2,8 @@
 
 class ActiveRecord::ConnectionAdapters::AbstractAdapter
   def semian_resource
-    @connection.semian_resource
+    # support for https://github.com/rails/rails/commit/d86fd6415c0dfce6fadb77e74696cf728e5eb76b
+    connection = instance_variable_defined?("@raw_connection") ? @raw_connection : @connection
```
paarthmadan, line 6: I don't have an opinion, in fact I'm simply curious:

I wonder if explicitly using the Rails version to discriminate on the ivar has advantages? It certainly makes the intent more clear, in that it's for backwards compatibility. I suppose it's more fragile, though.

Curious what others think.

nvasilevski: I feel like that question comes often and I haven't heard strong preference on this. I believe last time we were discussing something similar we agreed that using explicit Rails version branching more often used but no obvious advantage/disadvantage 

Also curious about other opinions!

casperisfine: When it's not too complicated I prefer feature checking, especially when it's for yet unreleased versions, because I've seen changes like this be reverted in the past, so feature testing is more reliable.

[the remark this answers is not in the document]
nvasilevski: Makes sense 👍  Changed

adrianna-chang-shopify (approved):

PR #311, opened 2022-04-15
Cleanup CI

casperisfine:
Remove some old hacks, and test against multiple ruby versions


.github/workflows/main.yml
```diff
@@ -8,12 +8,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        gemfile:
-          - gemfiles/hiredis-0-6.gemfile
-          - gemfiles/mysql2-0-4-10.gemfile
```
casperisfine, line 13: There's little point testing older mysql2 gems as for Ruby 3.0 compat we need to point at the git repo. So really 0.4.10 and `0.5.0` are basically unusable today.

PR #322, opened 2022-05-20
Improve CI latency work

miry:
Replace usage of Docker containers with github action solution.
Setup a container solution for running environment to have same logic as for docker-compose.
Setup hostname alias `semian`, so services can reach back to test container.

It improved running CI/CD from 15 mins to 4 mins (in case no flaky tests).

I have not add linting tasks for now, because semian requires a lot of upgrades.


tinchi (approved):
Loooks good !

PR #323, opened 2022-05-21
Run test in batches

miry:
Allow to split tests in batches and run.

I could not find a gem for it, so create a new single task base on Rake::TestTask

Currently it does not give huge boost, because I refactored github workflows and those already had improved test performance from 13 mins to 4 mins:

- https://github.com/Shopify/semian/pull/322
- https://github.com/Shopify/semian/pull/327

## Usage

It requires two environment variables:
* *TEST_WORKERS* - Total number of workers or batches. It uses to identify a total number of batches, that would be run in parallel. Default: 1
* *TEST_WORKER_NUM* - Current worker id. The value is between 1 and *TEST_WORKERS*

```shell
$ bundle exec rake test:parallel TEST_WORKERS=5 TEST_WORKER_NUM=1
```

## Alternatives

- [test-queue](https://github.com/tmm1/test-queue)
- [rails_parallel](https://github.com/Shopify/rails_parallel)
- [parallel_tests](https://github.com/grosser/parallel_tests)
- [ci-queue](https://github.com/shopify/ci-queue)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,6 @@ jobs:
     runs-on: ubuntu-latest
     # https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container
     container: ruby:${{ matrix.ruby }}
-
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


miry:
Tested parallel_test and found it is slow than just tests sorted by size.
ci-queue is just working with redis or other communications, that is not working correctly with github actions as it does not support parallel steps run.

tsevan (approved):
:+1:

Issue #315, opened 2022-04-28
Slow GRPC tests


casperisfine:
I started looking at CI a bit, as it's really slow and wonky.

After enabling individual test timing, it seems that just 3 GRPC tests account for over a third of the test suite time:

```
TestGRPC#test_bulkheads_tickets_are_working = 0.01 s = .
TestGRPC#test_unavailable_server_opens_the_circuit = 0.00 s = .
TestGRPC#test_errors_are_tagged_with_the_resource_identifier = 0.00 s = .
TestGRPC#test_circuit_breaker_on_server_streamer = 20.00 s = .
TestGRPC#test_rpc_server = 0.00 s = .
TestGRPC#test_instrumentation = 0.00 s = .
TestGRPC#test_semian_identifier = 0.00 s = .
TestGRPC#test_timeout_opens_the_circuit = 0.11 s = .
TestGRPC#test_circuit_breaker_on_bidi_streamer = 19.88 s = .
TestGRPC#test_circuit_breaker_on_client_streamer = 20.00 s = .
...
Finished in 180.014256s, 1.5388 runs/s, 4.1719 assertions/s.
```

Ref: https://github.com/Shopify/semian/pull/200

cc @max611 (since you added the adapter, maybe you have an idea what may take this long or at least how to debug this?)

PR #347, opened 2022-06-18
#315 fix: slow gRPC test by setting default client timeout

jychen7:
close https://github.com/Shopify/semian/issues/315

### DIFF
--- a/test/grpc_test.rb
+++ b/test/grpc_test.rb
@@ -174,7 +174,7 @@ class TestGRPC < Minitest::Test
 
   def open_circuit!(stub, method, args)
     ERROR_THRESHOLD.times do
-      assert_raises(GRPC::Unavailable) do
+      assert_raises(GRPC::DeadlineExceeded) do
         stub.send(method, args)
       end
     end
--- a/test/grpc_test.rb
+++ b/test/grpc_test.rb
@@ -8,6 +8,7 @@ require "mocha/minitest"
 require "echo_service"
 
 class TestGRPC < Minitest::Test
+  DEFAULT_CLIENT_TIMEOUT_IN_SECONDS = 3
   ERROR_THRESHOLD = 1
   SEMIAN_OPTIONS = {
     name: :testing,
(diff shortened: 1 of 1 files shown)


jychen7:
> [Sign the CLA](https://cla.shopify.com/) to have your PR reviewed

done

test/grpc_test.rb
```diff
@@ -190,7 +191,7 @@ def build_rpc_server(server_opts: {}, client_opts: {})
     @server = new_rpc_server_for_testing({ poll_period: 1 }.merge(server_opts))
     @port = @server.add_http2_port("0.0.0.0:#{SemianConfig["grpc_port"]}", :this_port_is_insecure)
     @host = "#{@hostname}:#{@port}"
-    @client_opts = client_opts
+    @client_opts = { timeout: DEFAULT_CLIENT_TIMEOUT_IN_SECONDS }.merge(client_opts)
```
jychen7, line 194: though gRPC code says default is infinity, but my local test also align with CI, seems that 20s is set
https://github.com/grpc/grpc/blob/53d69cc581c5b7305708587f4f1939278477c28a/src/ruby/lib/grpc/generic/client_stub.rb#L28-L29

20s is somewhat same as the example, maybe an coincide
https://grpc.io/blog/deadlines/#adjusting-deadlines

test/grpc_test.rb
```diff
@@ -8,6 +8,7 @@
 require "echo_service"
 
 class TestGRPC < Minitest::Test
+  DEFAULT_CLIENT_TIMEOUT_IN_SECONDS = 3
```
jychen7, line 11: result
```
TestGRPC#test_circuit_breaker_on_bidi_streamer = 3.00 s = .
TestGRPC#test_circuit_breaker_on_server_streamer = 3.00 s = .
TestGRPC#test_circuit_breaker_on_client_streamer = 3.00 s = .
```
CI log: https://github.com/jychen7/semian/runs/6955785910?check_suite_focus=true

jychen7:
> What do you think to keep at least one test case to cover GRPC::Unavailable?

I think unavailable is covered at
https://github.com/Shopify/semian/blob/40ef4400ab928526413969d446d663d5b6757c39/test/grpc_test.rb#L62-L72

but it is mocked. So if we want a real unavailable error, I think it is fine to keep one test case like original

PR #348, opened 2022-06-19
[CI] Handle GRPC::ActiveCall::Operation

jychen7:


lib/semian/grpc.rb
```diff
@@ -84,25 +84,55 @@ def disabled?
...
-      acquire_semian_resource(adapter: :grpc, scope: :bidi_streamer) { super }
+      scope =:bidi_streamer
+      acquire_semian_resource(adapter: :grpc, scope: scope) do
+        result = super
+        handle_operation(result, scope)
+        result
+      end
+    end
+
+    def handle_operation(result, scope)
```
jychen7, line 128: CI now pass
https://github.com/jychen7/semian/actions/runs/2524645306

miry:
Let me review the changes. Mean time please update Changelog.

jychen7:
> Let me review the changes. Mean time please update Changelog.

sure, added at  2784a2c2116676068e51d19b7c5b621a73328121

jychen7:
> Analysis / Ruby linter (pull_request)

ah, will fix

test/grpc_test.rb
```diff
@@ -53,8 +54,11 @@ def test_errors_are_tagged_with_the_resource_identifier
 
   def test_rpc_server
     run_services_on_server(@server, services: [EchoService]) do
-      GRPC::ActiveCall.any_instance.expects(:request_response)
+      GRPC::ActiveCall.any_instance.expects(:request_response).twice
       @stub.an_rpc(EchoMsg.new)
+      
+      operation = @stub_return_op.an_rpc(EchoMsg.new)
+      operation.execute
```
miry, line 61: There are two comments and suggestions:

1. I would prefer to separate unit test for each scenarios.

2. Move `stub` initialization from `setup` to test method.
It would avoid reling tests on global state and show all dependencies in same test.
That improve readability of the test.

```ruby
def test_rpc_server_with_operation
  subject = build_insecure_stub(EchoStubReturnOp)
  GRPC::ActiveCall.any_instance.expects(:request_response)

  run_services_on_server(@server, services: [EchoService]) do
    operation = subject.an_rpc(EchoMsg.new)
    operation.execute
  end
end
```

lib/semian/grpc.rb
jychen7: > Also do you know what can we do with duplication lines

as `acquire_semian_resource` is defined at [Adapter](https://github.com/Shopify/semian/blob/v0.12.0/lib/semian/adapter.rb#L32) and such `handle_operation` is only valid in gRPC, I think we can define a private method here in `lib/semian/grpc`, then use it.
e.g.
```
def bidi_streamer(*, **)
  return super if disabled?

  acquire_semian_resource_grpc(scope: :bidi_streamer) { super }
end

def acquire_semian_resource_grpc(scope:)
  acquire_semian_resource(adapter: :grpc, scope: scope) do
    result = yield
    handle_operation(result, scope) if result.is_a?(::GRPC::ActiveCall::Operation)
    result
  end
end
```

PR #364, opened 2022-08-01
Handle new Rails connection changes

eileencodes:
The refactoring in https://github.com/rails/rails/pull/44591 we started
seeing errors like `NoMethodError: undefined method 'semian_resource'
for nil:NilClass` coming from Semian. While this is called from
shopify/shopify, the `connection` that's `nil` is the one in the
`rails.rb` monkey patch.

After debugging with matthewd we found that Semian is relying on the
previous connection being available, whereas the new AR behavior means
it might not be.

To fix this we can memoize the `semian_resource` on `connection` in the
adapter, and then when the connection is reset or disconencted, set
`semian_resource` to `nil` and grab the new connection instance next
time we need it.

In addition to these changes I decided to add `activerecord` as a
development dependency to the gemspec so that we are able to test these
changes. Previously you would need to bundle open the gem in
shopify/shopify to test changes before sending a PR. It's important that
we're able to test what Semian is relying on so it makes sense to have a
depenedency on Active Record here.

Related: https://github.com/Shopify/shopify/pull/367400

### DIFF
--- a/test/rails_test.rb
+++ b/test/rails_test.rb
@@ -1,0 +1,78 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class TestRails < Minitest::Test
+  SUCCESS_THRESHOLD = 2
+  ERROR_THRESHOLD = 1
+  ERROR_TIMEOUT = 5
+  SEMIAN_OPTIONS = {
+    name: :testing,
+    tickets: 1,
+    timeout: 0,
+    error_threshold: ERROR_THRESHOLD,
+    success_threshold: SUCCESS_THRESHOLD,
+    error_timeout: ERROR_TIMEOUT,
+  }
+
+  def setup
+    @config = {
+      adapter: "mysql2",
+      connect_timeout: 2,
+      read_timeout: 2,
+      write_timeout: 2,
+      reconnect: true,
+      prepared_statements: false,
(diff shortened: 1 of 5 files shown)


semian.gemspec
```diff
@@ -28,4 +28,6 @@ Gem::Specification.new do |s|
   s.files = ::Dir["{lib,ext}/**/**/*.{rb,h,c}"] +
     ::Dir.glob("*.md")
   s.extensions = ["ext/semian/extconf.rb"]
+
+  s.add_development_dependency("activerecord", ">= 7.0.3")
```
miry, line 32: We handle all development dependency in Gemfile.

test/rails_test.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def test_semian_resource_is_available_after_reconnect
+    assert(@connection.semian_resource, "expected semian_resource to be available on @connection")
+
+    @connection.reconnect!
+
+    assert(@connection.semian_resource, "expected semian_resource to be available on @connection")
+  end
+end
```
miry, line 61: Can you also add test to check the result of `semian_resource`?
I would like to be sure that config is not just `nil`, 
but is equal to `SEMIAN_OPTIONS`

miry (approved):

PR #367, opened 2022-08-10
Check for OOM message with regex instead of start_with

eileencodes:
When I bumped this gem in our app I found the redis OOM tests were
failing due to the message match here being different. In the Rails app
tests messages said:

```
ERR Error running script (call to
f_8f56c8393dbc8bef0948f9c501f4c718c6944a14): @user_script:1:
@user_script: 1: -OOM command not allowed when used memory >
'maxmemory'.
```

These were failing because they don't start with OOM so the right error
is never raised. However when I put the original code back other tests
here started failing so I removed the regex `.\s*\z` and am just
matching on the message to raise the correct error.

I tried writing a test that would fail without the start with but
unfortunately it seemed that these tests always start with OOM.

Note: this is blocking the Rails upgrade because we can't merge/upgrade to #364 until these errors are also fixed.


tgwizard (approved):

miry:
@eileencodes Please update as well https://github.com/Shopify/semian/blob/master/CHANGELOG.md

PR #370, opened 2022-08-16
Release 0.13.1

casperisfine:


miry (approved):

casperisfine:
@miry deploys are failing because `grpc` wont' install on shipit somehow (it's like a running gag...) https://shipit.shopify.io/shopify/semian/rubygems/deploys/1770187

Would you have time to look into it? Otherwise I'll have a look tomorrow. I think we don't need that gem for release, so we can likely customize the `shipit.yml` or `gemfile` a little bit.

miry:
> Would you have time to look into it? Otherwise I'll have a look tomorrow. I think we don't need that gem for release, so we can likely customize the shipit.yml or gemfile a little bit.

I agree. Checking shipit.yml

miry:
@casperisfine I think because of new grpc we had failed tests. 
So I am reverting it back: https://github.com/Shopify/semian/pull/371

I don't remember if it is enough to specify group `deploy` for shipit to avoid install everything else.

miry:
Done: https://rubygems.org/gems/semian

PR #381, opened 2022-08-31
Revert "Bump redis from 4.7.1 to 4.8.0"

shanempope:
Reverts Shopify/semian#377


miry:
@shanempope can you provide description, why do you want to revert PR?

I checked last time all checks were passed, i wonder what is not working?

PR #382, opened 2022-08-31
Release 0.13.3

shanempope:


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Unreleased
 
+# v0.13.3
+
+* Bump redis from 4.7.1 to 4.8.0
```
miry, line 5: There is no semian gem dependencies to redis. We don't need any new version of gem for test environment.

Issue #385, opened 2022-09-01
Compatibility issues with Redis 5.X


muz:
Saw this on an internal-app of ours with Dependabot attempting to bump `redis` from `4.8.0` to `5.0.2`;

```
NameError: method `io' not defined in Redis::Client
--
  |  
  | base.send(:remove_method, :io)
  | ^^^^^
  | /tmp/bundle/ruby/3.1.0/gems/semian-0.13.2/lib/semian/redis.rb:71:in `remove_method'
  | /tmp/bundle/ruby/3.1.0/gems/semian-0.13.2/lib/semian/redis.rb:71:in `included'
  | /tmp/bundle/ruby/3.1.0/gems/semian-0.13.2/lib/semian/redis.rb:153:in `include'
  | /tmp/bundle/ruby/3.1.0/gems/semian-0.13.2/lib/semian/redis.rb:153:in `<main>'
````
  
https://github.com/Shopify/semian/blob/07a2a21b3a7bca412e3f116ba9b2e2672059db5b/lib/semian/redis.rb#L72
  

----

In `redis` 4.8 we can see this method defined here https://github.com/redis/redis-rb/blob/9edb20462fb4578fce9ffa802a56ddb711643716/lib/redis/client.rb#L298-L307

It looks like the method was removed as part of https://github.com/redis/redis-rb/commit/08a21001b5f27088a412fbe7a6fba8d558eade40#diff-5bc007010e6c2e0aa70b64d6f87985c20986ee1b2882b63a89b52659ee9c91f8L296

Semian'll need updating to support Redis 5.X by not trying to remove this now-removed method

miry:
It means redis now uses redis-client in background.
It means we can load redis_client semian instead.

PR #386, opened 2022-09-02
Detect Redis 5

miry:
Print warning message redis 5.x during load time.

```
# gem install pkg/semian-0.13.2.gem
Building native extensions. This could take a while...

==============================================================================

semians is not compatible with redis 5.x.".
Update Gemfile to use older redis version:

    gem "redis", "~> 4.8"

==============================================================================

Successfully installed semian-0.13.2
1 gem installed
````

Ignore Semian settings for redis 5.

```irb
> require "redis"

> Redis::VERSION
=> "5.0.2"

> require "semian/redis"
W, [2022-09-02T16:33:00.722716 #194]  WARN -- : NOTE: Semian is not compatible with Redis 5.x
```

I would like to have this hack.
Mean time work on proper solution in https://github.com/Shopify/semian/issues/385.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -4,7 +4,8 @@ FROM ruby:${RUBY_VERSION} as base
 # Avoid warnings by switching to noninteractive
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update && apt-get install --no-install-recommends -y \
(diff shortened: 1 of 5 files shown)


muz (approved):

casperisfine:
Redis 5 is already somewhat supported since it's a wrapper around `redis-client` which already has a semian adapter.

But I'll do a pass on Monday to make sure it's transparent.

PR #388, opened 2022-09-05
Redis 5.0 support

### LINKED ISSUE #385: Compatibility issues with Redis 5.X

casperisfine:
Since 5.0 use redis-client under the hood, for the most part we only need to translate errors and delegate a few methods.

Closes #385

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,8 +14,8 @@ group :test do
   gem "mysql2", "~> 0.5"
   gem "pry-byebug", require: false
   gem "rake-compiler"
-  gem "hiredis-client"
-  gem "redis", "~> 4.8"
+  gem "hiredis-client", github: "redis-rb/redis-client"
+  gem "redis", github: "redis/redis-rb"
   gem "timecop"
   gem "toxiproxy"
   gem "webrick"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,3 +1,19 @@
+GIT
+  remote: https://github.com/redis-rb/redis-client.git
+  revision: 5dc39f533f96fb94ebd6ef6cf7776b43f6243791
+  specs:
+    hiredis-client (0.7.3)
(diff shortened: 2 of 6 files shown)


casperisfine:
Ah, I was planning to merge just yet, I wanted to cut new releases for `redis` and `redis-client` gems so we wouldn't need to point to the git repos in the Gemfile.

But no big deal I can PR that tomorrow.

PR #411, opened 2022-09-23
Enrich to_s for Resources classes.

miry:
When working with Semian, found to be usefull have light version of Resource.

Provide more informaion for string represintation of `Semian:ProtectedResource`,
  `Semian:Resource`, and `Semian:CircuitBreaker` instances.

Change the return value of `to_s` from class name with address to
class name with some fields.

Example:

* `"#<Semian::ProtectedResource:0x0000ffffa6f38558>"` -> `#<Semian::ProtectedResource name: testing, circuit_breaker: #<Semian::CircuitBreaker name: testing, half_open_resource_timeout: , error_timeout: 5, closed?: true, open?: false, half_open?: false, last_error: >, bulkhead: #<Semian::Resource name: testing>>`
* `"#<Semian::CircuitBreaker:0x0000ffffa6f3a830>"` -> `#<Semian::CircuitBreaker name: testing, half_open_resource_timeout: , error_timeout: 5, closed?: true, open?: false, half_open?: false, last_error: >`
* `"#<Semian::Resource:0x0000ffffa6f38d00>"` -> `#<Semian::Resource name: testing>`

### DIFF
--- a/lib/semian/circuit_breaker.rb
+++ b/lib/semian/circuit_breaker.rb
@@ -89,6 +89,17 @@ module Semian
       !error_timeout_expired? && !@errors.empty?
     end
 
+    def to_s
+      "\#<#{self.class} name: #{@name}, " \
+        "half_open_resource_timeout: #{@half_open_resource_timeout}, " \
+        "error_timeout: #{@error_timeout}, " \
+        "closed?: #{@state.closed?}, " \
+        "open?: #{@state.open?}, " \
+        "half_open?: #{@state.half_open?}, " \
+        "implementation: #{@state.class.to_s.split("::")[-2]}, " \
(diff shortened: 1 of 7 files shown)


tinchi (approved):
👍

lib/semian/circuit_breaker.rb
```diff
@@ -89,6 +89,17 @@ def in_use?
       !error_timeout_expired? && !@errors.empty?
     end
 
+    def to_s
```
pl-at-shopify, line 92: I'd guess you're trying to implement the `inspect` method that is intended to be used during debugging. The `to_s` method is more general purpose and should provide only the information that is meaningful in text context (usually the less, the better to avoid unintentional information leakage if exposed to the user).

PR #418, opened 2022-09-27
Provide optional flag to disable time window based error eviction.

whatisinternet:
Provide a way to disable the time based pruning of errors and allow for "errors in a row" type behaviour. This will allow circuits to trip more consistently when distributed requests may not hit the same host enough times within a timeout.

### DIFF
--- a/test/circuit_breaker_test.rb
+++ b/test/circuit_breaker_test.rb
@@ -13,8 +13,14 @@ class TestCircuitBreaker < Minitest::Test
     rescue
       nil
     end
-    Semian.register(:testing, tickets: 1, exceptions: [SomeError], error_threshold: 2, error_timeout: 5,
-      success_threshold: 1)
+    Semian.register(
+      :testing,
+      tickets: 1,
+      exceptions: [SomeError],
+      error_threshold: 2,
+      error_timeout: 5,
+      success_threshold: 1,
+    )
     @resource = Semian[:testing]
   end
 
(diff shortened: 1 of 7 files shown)


test/circuit_breaker_test.rb
```diff
@@ -63,6 +69,26 @@ def test_once_success_threshold_is_reached_only_error_threshold_will_open_the_ci
     assert_circuit_opened
   end
 
+  def test_once_success_threshold_is_reached_only_error_threshold_will_open_the_circuit_again_without_timeout
+    resource = Semian.register(
+      :three,
+      tickets: 1,
```
miry, line 75: I would use replace `ticket: 1` with `bulkhead: false` to make sure there is no options from "bulkheading", and only test Circuit breaker.

whatisinternet: Yeah good call. Fixed up.

whatisinternet:
🤔 Yeah we could use `0` as a boolean in this case but does make the variable a little overloaded. Commit soon to trial it.

whatisinternet:
@miry I did try it out and found that it wasn't _great_ in terms of long term maintainability or clarity.

We end up with the following in the initializer

```ruby
      if error_threshold_timeout.nil?
        @use_timeout = true
        @error_threshold_timeout = error_timeout
      elsif error_threshold_timeout == 0
        @use_timeout = false
        @error_threshold_timeout = error_timeout
      else
        @use_timeout = true
        @error_threshold_timeout = error_threshold_timeout
      end
```

In fairness I _think_ there is a bug in the existing code because an `error_threshold_timeout` of `0` is possible and will result in evicting errors when the user may not expect them to be evicted.

miry:
Thanks for trying. I think `use_timeout` could be misleaded and used for Bulkheading, because there is option: `timeout` in https://github.com/Shopify/semian#bulkhead-configuration

miry (approved):
👍🏼

whatisinternet:
That's fair. I really struggled with a name so if there is something better without going into a bike shed I'm very happy to change it.

franciscojaviermejia (approved):

PR #420, opened 2022-10-03
WIP: Adapter for GitHub's Trilogy MySQL client

casperisfine:
This puts the basis of an adapter, it's still missing quite a few things.

We might also need to add a few things upstream, such as `Trilogy#closed?` etc.

Also Trilogy raise a lot of different base exceptions, so it's very tricky to list them all. We might need to resort to consider all `IOError` and `SystemCallError` subclasses.

Co-Authored-By: @adrianna-chang-shopify

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,6 +17,7 @@ group :test do
 
   gem "grpc", "1.46.3"
   gem "mysql2", "~> 0.5"
+  gem "trilogy"
   gem "activerecord", ">= 7.0.3"
   gem "hiredis", "~> 0.6"
   gem "hiredis-client"
--- a/Gemfile
+++ b/Gemfile
@@ -32,3 +32,5 @@ group :lint do
 end
 
 gemspec
+
+gem "debug", "~> 1.6"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -84,6 +84,7 @@ GEM
     ruby-progressbar (1.11.0)
     timecop (0.9.5)
     toxiproxy (2.0.2)
+    trilogy (2.1.1)
     tzinfo (2.0.5)
       concurrent-ruby (~> 1.0)
     unicode-display_width (2.3.0)
@@ -114,6 +115,7 @@ DEPENDENCIES
   semian!
   timecop
   toxiproxy
+  trilogy
   webrick
 
 BUNDLED WITH
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -84,7 +84,7 @@ GEM
     ruby-progressbar (1.11.0)
     timecop (0.9.5)
     toxiproxy (2.0.2)
-    trilogy (2.1.1)
+    trilogy (2.2.0)
     tzinfo (2.0.5)
       concurrent-ruby (~> 1.0)
     unicode-display_width (2.3.0)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -23,6 +23,9 @@ GEM
     coderay (1.1.3)
     concurrent-ruby (1.1.10)
     connection_pool (2.3.0)
+    debug (1.6.3)
+      irb (>= 1.3.6)
+      reline (>= 0.3.1)
     google-protobuf (3.21.6)
     google-protobuf (3.21.6-x86_64-darwin)
     googleapis-common-protos-types (1.4.0)
@@ -38,6 +41,9 @@ GEM
       redis-client (= 0.9.0)
     i18n (1.12.0)
       concurrent-ruby (~> 1.0)
+    io-console (0.5.11)
+    irb (1.5.1)
+      reline (>= 0.3.0)
     json (2.6.2)
     memory_profiler (1.0.0)
     method_source (1.0.0)
@@ -62,6 +68,8 @@ GEM
     redis-client (0.9.0)
       connection_pool
     regexp_parser (2.5.0)
+    reline (0.3.1)
+      io-console (~> 0.5)
     rexml (3.2.5)
     rubocop (1.36.0)
       json (~> 2.3)
@@ -97,6 +105,7 @@ PLATFORMS
 DEPENDENCIES
   activerecord (>= 7.0.3)
   benchmark-memory
+  debug (~> 1.6)
   grpc (= 1.46.3)
   hiredis (~> 0.6)
   hiredis-client
--- a/lib/semian/mysql2.rb
+++ b/lib/semian/mysql2.rb
@@ -46,6 +46,7 @@ module Semian
     class << self
       # The naked methods are exposed as `raw_query` and `raw_connect` for instrumentation purpose
       def included(base)
+        require 'debug'; debugger
         base.send(:alias_method, :raw_query, :query)
         base.send(:remove_method, :query)
 
--- a/lib/semian/trilogy.rb
+++ b/lib/semian/trilogy.rb
@@ -21,6 +21,14 @@ class Trilogy
 
   ResourceBusyError = Class.new(SemianError)
   CircuitOpenError = Class.new(SemianError)
+
+  # method aliasing needs to happen first
+  # The naked methods are exposed as `raw_query` and `raw_ping` for instrumentation purpose
+  alias_method(:raw_query, :query)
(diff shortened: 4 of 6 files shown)


adrianna-chang-shopify:
Okay, tests are now passing for the Trilogy test suite 🎉 Handling the different base exceptions Trilogy raises is a bit tricky, we might need to do some work upstream before we can ship the adapter. I made some notes on the different situations we get into, would love feedback if anyone has time.

### Expected cases
#### Query error 
* e.g. `client.query("ERROR!")`
* Produces `Trilogy::Error => trilogy_query_recv: 1064 You have an error in your SQL syntax;`
* Should not open circuit
* See `test_query_errors_does_not_open_the_circuit`

#### Read timeout
*  e.g. `client.query("SELECT sleep(5)")`
* Produces `::Errno::ETIMEDOUT`
* Should open the circuit
* See `test_read_timeout_error_opens_the_circuit`

#### Connection errors due to network latency
* Simulate with proxy latency, e.g. `@proxy.downstream(:latency, latency: 2200).apply`
* Produces `::Errno::ETIMEDOUT`
* Should open the circuit
* See `test_connection_errors_opens_the_circuit`

#### Network errors
* Simulate with `@proxy.down`
* Produces `Trilogy::Error => [mysql_testing] trilogy_query_recv: TRILOGY_CLOSED_CONNECTION`
* Should open the circuit
* See `test_network_errors_are_tagged_with_the_resource_identifier`

#### Errno::ECONNREFUSED
* Occurs if we try to connect to the db when the network is down, e.g. `@proxy.down`
* Should result in the circuit opening

### Edge cases (looking for feedback)
#### Connection reset by peer
* `Errno::ECONNRESET: Connection reset by peer - trilogy_query_send`
* Occurs if too much time elapsed since connection was used
* I don’t think this should cause the circuit to open, since this is a “fail fast” situation
* See `test_changes_timeout_when_half_open_and_configured`:
```diff
diff --git a/test/adapters/trilogy_test.rb b/test/adapters/trilogy_test.rb
index 7b11ae0..8f56c21 100644
--- a/test/adapters/trilogy_test.rb
+++ b/test/adapters/trilogy_test.rb
@@ -356,7 +356,9 @@ class TestTrilogy < Minitest::Test
       # Trilogy has reset the connection request at this point, so need to reconnect
       # Circuit breaker is only created once with half_open_resource_timeout, so
       # no need to pass that through again
-      connect_to_mysql!
+      # connect_to_mysql!
+      # If we try to use the client here, we'll see connection reset by peer error
+      client.query("SELECT 1;")
     end
```
#### IOError
* Occurs when attempting to use a client where the connection to the server has explicitly been closed (via `#close`)
* Don’t think it should result in circuit opening, also a “fast failure” error

#### Trilogy::Error
* `[mysql_testing] trilogy_query_send: TRILOGY_CLOSED_CONNECTION`
* I'm also seeing this error pop up after `::Errno::ETIMEDOUT` exceptions -- seems like Trilogy closes the connection
* The tricky thing is that I don't _think_ this type of error should result in the circuit opening, should it? But I'm not sure how to differentiate between this `TRILOGY_CLOSED_CONNECTION` exception and the one raised above when the network was down.. I haven't dug too deeply into the Trilogy source yet, but maybe we'll need two different types of errors to differentiate these two situations
* See `test_read_timeout_error_opens_the_circuit`:
```diff
diff --git a/test/adapters/trilogy_test.rb b/test/adapters/trilogy_test.rb
index 7b11ae0..2a2eae7 100644
--- a/test/adapters/trilogy_test.rb
+++ b/test/adapters/trilogy_test.rb
@@ -88,7 +88,8 @@ class TestTrilogy < Minitest::Test
     # After Trilogy::CircuitOpenError check regular queries are working fine.
     result = Timecop.travel(ERROR_TIMEOUT + 1) do
       # Reconnect because trilogy closed the connection
-      client = connect_to_mysql!
+      # If we don't reconnect here we'll see Trilogy::Error
+      # client = connect_to_mysql!
       client.query("SELECT 1 + 1;")
     end
```
Also wondering about the difference between this and the "reset by peer" exception 🤔 

One other interesting scenario I encountered:
There's a strange interaction with the circuit being half-opened, a `half_open_resource_timeout` being set, and us needing to re-establish the connection with MySQL. This test demonstrates the issue:
```ruby
def test_changes_timeout_when_half_open_and_configured
    client = connect_to_mysql!(half_open_resource_timeout: 0.5)

    # Timeout is 2 seconds, this causes circuit to open
    @proxy.downstream(:latency, latency: 3000).apply do
      ERROR_THRESHOLD.times do
        assert_raises(Errno::ETIMEDOUT) do
          client.query("SELECT 1 + 1;")
        end
      end
    end

    assert_raises(Trilogy::CircuitOpenError) do
      client.query("SELECT 1 + 1;")
    end

    client = Timecop.travel(ERROR_TIMEOUT + 1) do
      connect_to_mysql! # <=============== THIS LINE HERE
    end

    @proxy.downstream(:latency, latency: 750).apply do
      assert_raises(Errno::ETIMEDOUT) do
        client.query("SELECT 1 + 1;")
      end
    end

    Timecop.travel(ERROR_TIMEOUT + 1) do
      # Need to reconnect again because Trilogy reset conn request
      client = connect_to_mysql!
      client.query("SELECT 1 + 1;")

      # Timeout has reset to the normal 2 seconds now that circuit is closed
      @proxy.downstream(:latency, latency: 1500).apply do
        client.query("SELECT 1 + 1;")
      end
    end

    assert_equal(2, client.read_timeout)
    assert_equal(2, client.write_timeout)
  end
```
If you look at the line I've marked, we're needing to reconnect to MySQL because Trilogy has closed the connection.
The circuit is half open, and so on calling `#initialize` and attempting to connect to the db, we go through `with_resource_timeout`, which uses `self.read_timeout` to fetch the current read timeout for the client. The problem is that we're doing this on the already-closed connection (remember we're in the process of reconnecting), so this produces an `IOError`. 

I think eventually we'll want to check if the connection is `#closed?` and yield the block anways, so that we can always connect via `Trilogy.new` without the previously-closed-connection causing problems. To get around this in the meantime, I've opted to rescue `IOError`, check if the error is related to the connection being closed, and it if is yield the block anyways -- either we're in the process of reconnecting (`Trilogy.new`), so we'll disregard the `IOError` and reconnect properly, or we're acquiring some other resource, and it's likely yielding the block will re-raise:
```ruby
    def with_resource_timeout(temp_timeout)
      prev_read_timeout = read_timeout
      self.read_timeout = temp_timeout
      yield
    rescue IOError => error
      raise unless error.message.match?(/connection closed/)

      yield
    ensure
      self.read_timeout = prev_read_timeout unless prev_read_timeout.nil?
    end
```
Does this make sense?

In general, Trilogy seems to be eager about closing unused connections (unlike Mysql2 AFAICT), so we might need to be more defensive in the adapter code.

Would love your thoughts @paarthmadan @casperisfine

miry:
> I don’t think this should cause the circuit to open, since this is a “fail fast” situation
>
> This tells us the connection was closed on the server side. I'm not certain wether we should consider it as a network error. @miry any opinion?

There are various reasons for TCP/IP sending a reset,
most lucky not related to overloaded DB or network disturbance.
It would open a circuit breaker too often.
I would ignore `Errno::ECONNRESET` in semians.

Something to note is that it could be `ECONNRESET` returns
with Toxiproxy and Docker proxy usage.
Docker proxy always listens open port, 
and if Toxiproxy enables "down" toxic - it could return `ECONNRESET`.

adrianna-chang-shopify:
Thanks for the feedback, both of you! I'll make some adjustments, but does seem like we need to refactor the error classes upstream before proceeding with this.

> Something to note is that it could be ECONNRESET returns
with Toxiproxy and Docker proxy usage.
Docker proxy always listens open port,
and if Toxiproxy enables "down" toxic - it could return ECONNRESET.

Ah, interesting! Am I correct in interpreting this to mean that the  `ECONNRESET` we're seeing in the tests (ie. what was described in `#test_changes_timeout_when_half_open_and_configured`) does not correspond to behaviour we'd see in production, and is purely a side-effect of testing with a Docker-based environment?

PR #423, opened 2022-10-07
Avoid chaining duplicated semian identifiers in error messages

casperisfine:
When errors are created from other errors messages they might already have a semian identifier in their message

### DIFF
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -122,11 +122,16 @@ module Semian
     attr_accessor :semian_identifier
 
     def to_s
+      message = super
       if @semian_identifier
-        "[#{@semian_identifier}] #{super}"
-      else
-        super
+        prefix = "[#{@semian_identifier}] "
+        # When an error is created from another error's message it might
+        # already have a semian identifier in their message
+        unless message.start_with?(prefix)
+          message = "#{prefix}#{message}"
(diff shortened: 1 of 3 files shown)


lib/semian.rb
```diff
@@ -122,11 +122,16 @@ module AdapterError
...
+      message = super
       if @semian_identifier
-        "[#{@semian_identifier}] #{super}"
-      else
-        super
+        prefix = "[#{@semian_identifier}] "
+        # When errors are created from other errors messages they might
+        # already have a semian identifier in their message
+        unless message.start_with?(prefix)
+          message = "#{prefix}#{message}"
```
miry, line 131: Should prefix variable moved inside this block?

miry: my bad. pls ignore

casperisfine:
> Is there anywhere in the repo where we're currently constructing an error from another error's message where this was a problem?

Not in the repo, but `redis-rb` 5.0 does this a lot with `redis-client` errors: https://github.com/redis/redis-rb/blob/d4b24e1e4fac7b9e3fee01947bf400f52ea500cf/lib/redis/client.rb#L118-L126

miry:
I think it should be mentioned in Changelog as well.

PR #426, opened 2022-10-09
Remove unused variable in log_state_transition

miry:
The function returns the last message,
where the result is not required.


pftg (approved):
👍

PR #427, opened 2022-10-09
Remove query of ENV on each circuit breaker acquire.

miry:
Move logic of disabled? for Circuit breaker to Initalization part.
Circtuit breaker should have any knowledge of ENV or configurations.


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 * Avoid prepending the same prefix twice to errors messages. (#423)
   This mostly happens with the `redis-rb 5+` gem as it translate `redis-client` errors.
+* Environment variable `SEMIAN_DISABLED` now skips creating a Semian resource. (#427)
+  Introduce environment variable `SEMIAN_CIRCUIT_BREAKER_DISABLED` to disable only
+  `circuit_breaker` and `SEMIAN_BULKHEAD_DISABLED` only `bulkhead`.
```
casperisfine, line 7: So it's a bit nitpick territory, but common Ruby gems use an imperative stance for these variables:

  - `DISABLE_BOOTSNAP=1`
  - `DISABLE_BOOTSNAP_LOAD_PATH_CACHE=1`
  - `DISABLE_SPRING=1`
  - etc.

It might make sense to be consistent here.

lib/semian.rb
```diff
@@ -275,7 +277,9 @@ def thread_safe=(thread_safe)
   private
 
   def create_circuit_breaker(name, **options)
-    circuit_breaker = options.fetch(:circuit_breaker, true)
+    not_disabled_by_env = !ENV.key?("SEMIAN_CIRCUIT_BREAKER_DISABLED")
+
+    circuit_breaker = options.fetch(:circuit_breaker, not_disabled_by_env)
```
casperisfine, line 282: It's interesting that you chose this precedence. Could you expand on how you expect it to be used?

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
...
+  Explicit settings has higher priority over environment variables, except `SEMIAN_DISABLED`.
```
AbdulRahmanAlHamali, line 8: This needs some elaboration for users. Maybe an example?

CHANGELOG.md
AbdulRahmanAlHamali: This would make it a breaking change, since those environment variable names already existed

AbdulRahmanAlHamali (commented):
LGTM, but I see no documentation in the README about these environment variables. Might be worth including that?

lib/semian/platform.rb
```diff
@@ -13,6 +13,6 @@ def semaphores_enabled?
   end
 
   def disabled?
-    ENV["SEMIAN_SEMAPHORES_DISABLED"] || ENV["SEMIAN_DISABLED"]
+    ENV.key?("SEMIAN_SEMAPHORES_DISABLED") || ENV.key?("SEMIAN_DISABLED")
```
jgaskins, line 16: The PR description says that these checks were being moved to initialization. Should these env checks maybe be assigned to a constant to achieve that?

jgaskins (commented):
Feedback is mostly questions. I haven't looked too deeply into the APIs in this gem.

Also TIL `ENV[]` is slower than `ENV.key?`

```
                  []   511.958k i/100ms
               .key?   928.543k i/100ms
                  []      5.104M (± 0.6%) i/s -     25.598M in   5.015834s
               .key?      9.333M (± 0.4%) i/s -     47.356M in   5.073909s

Comparison:
               .key?:  9333316.7 i/s
                  []:  5103582.6 i/s - 1.83x  (± 0.00) slower
```

<details><summary>Benchmark code</summary>

```ruby
# frozen_string_literal: true
require "benchmark/ips"

Benchmark.ips do |x|
  x.report("[]") { ENV["HOME"] }
  x.report(".key?") { ENV.key?("HOME") }
  x.compare!
end
```

</details>

lib/semian.rb
miry: Currently we have 2 different mechanisms of disabling circuit_breaker.
1. Resource configuration: `circuit_breaker: false` - It does not initialize circuit_breaker field in ProtectedResrouce.
2. With variable SEMIAN_CIRCUIT_BREAKER_DISABLED, that checked on each `acquire` and do `yield` if it is exists.

I merged both solutions in single one. (kinda get rid from second solution).

I have doubts to have ENV variable, because there is no documentation exists of using those variables. 
My first impression to implement the similar flow, 
but honestly I would just remove all usage of ENVs and
allow Apps decide create or not the Semian resources base on ENVs.

@byroot let me know what do you think.

lib/semian/circuit_breaker.rb
```diff
@@ -36,8 +36,6 @@ def initialize(name, exceptions:, success_threshold:, error_threshold:,
     end
 
     def acquire(resource = nil, &block)
-      return yield if disabled?
```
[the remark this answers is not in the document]
miry: It is done in `protected_resource` exactly the same thing:

https://github.com/Shopify/semian/blob/master/lib/semian/protected_resource.rb#L41

CHANGELOG.md
casperisfine: I missed these already existed. It's not worth renaming.

lib/semian.rb
AbdulRahmanAlHamali: hmmmm good point

lib/semian/platform.rb
miry: @jgaskins I would left as it is for now. It is used in few places with few branchings, that would made PR even bigger. I am going to create a separate PR for this.

test/semian_test.rb
```diff
@@ -3,11 +3,10 @@
...
 class TestSemian < Minitest::Test
-  def setup
-    Semian.destroy(:testing)
-  rescue
-    nil
-  end
+  # def setup
+  #   Semian.unregister_all_resources
+  #   Semian.destroy_all_resources
+  # end
```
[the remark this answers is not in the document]
miry: fixed

lib/semian/platform.rb
miry: created a separate issue for it: https://github.com/Shopify/semian/issues/442

miry:
Hi folks. If you have time, pls review again.

test/circuit_breaker_test.rb
```diff
@@ -358,18 +358,50 @@ def test_error_threshold_timeout_is_skipped_when_not_using_error_threshold_and_n
...
+      :global_env_var_disables_circuit_breaker,
+      bulkhead: true,
+      tickets: 1,
+      circuit_breaker: true,
+      error_threshold: 2,
+      error_timeout: 5,
+      success_threshold: 1,
+    )
+
+    assert_nil(resource)
```
jgaskins, line 404: Looks like we always returned a non-nil value from `.register` before (or raised an exception). Is this method exposed to apps using Semian?

If so, and I'm reading it correctly, this may not be backwards-compatible.

jgaskins (commented):
Mostly clarity in docs but also a compatibility question

test/circuit_breaker_test.rb
miry: Mostly. The problem that before it would not be raise, but NOOP. Because it created always resource objects, and skip logic during on acquire.

miry: That is true. Need may be stress it in changelog and also in gemspec post install notifications.

README.md
```diff
@@ -419,9 +419,11 @@ response time. This is the problem Semian solves by failing fast.
...
 ## How does Semian work?
 
-Semian consists of two parts: **Circuit breaker** and **Bulkheading**. To understand
-Semian, and especially how to configure it, we must understand these patterns
-and their implementation.
+Semian consists of two parts: **Circuit Breaker** and **Bulkheading**.
+To understand Semian, and especially how to configure it,
+we must understand these patterns and their implementation.
+
+Disable Semian via environment variable `SEMIAN_DISABLED`.
```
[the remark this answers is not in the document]
miry: fixed

README.md
```diff
@@ -516,11 +528,15 @@ still experimenting with ways to figure out optimal ticket numbers. Generally
...
 queried frequently has worked well for us.
 
+* **bulkhead**. Enable or Disable Bulkhead. Defaults to `true` if not set.
 * **tickets**. Number of workers that can concurrently access a resource.
 * **timeout**. Time to wait in seconds to acquire a ticket if there are no tickets left.
   We recommend this to be `0` unless you have very few workers running (i.e.
   less than ~5).
 
+It is possible to disable Bulkhead with environment variable
+`SEMIAN_BULKHEAD_DISABLED`.
```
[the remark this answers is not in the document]
miry: fixed

README.md
```diff
@@ -452,18 +454,28 @@ all workers on a server.
...
+  the time window for evicting old exceptions. `error_timeout` is still used and
+  will reset the circuit. Defaults to `true` if not set.
 * **success_threshold**. The amount of successes on the circuit until closing it
   again, that is to start accepting all requests to the circuit.
-* **half_open_resource_timeout**. Timeout for the resource in seconds when the circuit is half-open (supported for MySQL, Net::HTTP and Redis).
+* **half_open_resource_timeout**. Timeout for the resource in seconds when
+  the circuit is half-open (supported for MySQL, Net::HTTP and Redis).
+
+It is possible to disable Circuit Breaker with environment variable
+`SEMIAN_CIRCUIT_BREAKER_DISABLED`.
```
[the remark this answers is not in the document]
miry: fixed

PR #435, opened 2022-10-31
[Proof of Concept] Make Semian work with `activerecord-trilogy-adapter`

adrianna-chang-shopify:
An alternative approach to https://github.com/Shopify/semian/pull/420, where we target resource acquisition at the Active Record adapter layer instead of patching the Trilogy client directly.

This is a simpler approach, and allows us to deal with native Active Record errors instead of the various errors Trilogy is currently raising. The tests are also a lot simpler because we don't have to handle re-instantiating the Trilogy client every time the connection is closed (either server or client side).

Caveats:
- We'll need to deal with SFR separately
- There is still work to be done in `activerecord-trilogy-adapter` to make some of the error classes better (specifically in `#new_client`)

### DIFF
--- a/lib/semian/trilogy_adapter.rb
+++ b/lib/semian/trilogy_adapter.rb
@@ -27,6 +27,9 @@ module Semian
   module TrilogyAdapter
     include Semian::Adapter
 
+    ResourceBusyError = ::ActiveRecord::ConnectionAdapters::TrilogyAdapter::ResourceBusyError
+    CircuitOpenError = ::ActiveRecord::ConnectionAdapters::TrilogyAdapter::CircuitOpenError
+
     attr_reader :raw_semian_options, :semian_identifier
 
     def initialize(*options)
@@ -58,6 +61,8 @@ module Semian
       acquire_semian_resource(adapter: :trilogy_adapter, scope: :ping) do
         super
       end
+    rescue ResourceBusyError, CircuitOpenError => error
+      false
     end
 
     def with_resource_timeout(temp_timeout)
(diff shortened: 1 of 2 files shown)


lib/semian/trilogy_adapter.rb
```diff
@@ -0,0 +1,117 @@
...
+        @config[:read_timeout] = temp_timeout # Create new client with config
+      else
+        prev_read_timeout = connection.read_timeout
+        connection.read_timeout = temp_timeout
+      end
+      yield
+    ensure
+      @config[:read_timeout] = prev_read_timeout
+      connection&.read_timeout = prev_read_timeout
+    end
```
adrianna-chang-shopify, line 68: Okay, so unfortunately making `half_open_resource_timeout` work is still a bit hacky, even working at the adapter layer. I'm racking my brain for another way to do this, but not sure we have many options.

Basically, the Trilogy adapter forwards configs to the client, one of which is the `read_timeout`. In the context of `with_resource_timeout`, we might be connecting to the db, in which case `connection` is nil. Consequently, the only way I see being able to set a new (temp) `read_timeout` is to mutate the `@config` so that the newly instantiated client gets created with the timeout we want. If we _have_ the connection already, we can reach directly to the connection (the `Trilogy::Client`) and modify the `read_timeout`.

In the ensure block, we can reset the timeout on the connection, and I'm assuming we should also reset the config.

This is additionally complex because if _no_ `read_timeout` is set in the config, we need to use the client's default, which I _know_ is 0 but could potentially change.

PR #453, opened 2022-12-05
Tune examples to cover some cases

miry:
Update examples to limit expected expectations and mark lines that should not be reached.
Replaced some rescue to single errors, to not cover to many problems.
Add lines to bulkhead examples to make sure script finished with unsuccess,
if thread or fork got exception.

Extract changes from https://github.com/Shopify/semian/pull/441/


Eagerod (approved):
👌🏽 Big fan of catching the specifically expected exception, rather than any.

PR #468, opened 2023-02-01
Semian + Trilogy: Introduce adapter for Trilogy AR Adapter

adrianna-chang-shopify:
We are planning to migrate Shopify/shopify (and our other Rails applications) to the Trilogy client instead of Mysql2. This PR introduces a Semian adapter for the Trilogy Active Record adapter.

### Why isn't this built at the Trilogy client layer?
Building the Semian adapter at the Trilogy lient layer is challenging because of the way Trilogy manages its connection to the server. Any time an error is encountered, Trilogy will close the connection, and automatic reconnects are not supported. This means that `TRILOGY_CONNECTION_CLOSED` errors can happen for various reasons: if the network is down, but also if some other error was encountered and the connection is no longer usable. We’d need to make non-trivial changes to Trilogy to differentiate these use cases. Targeting the AR adapter layer allows us to lean on Rails to reconnect as needed, and allows us to differentiate between true connection errors (e.g. the network is down), vs Trilogy having closed the connection itself.

### Why don't we write a Semian adapter that targets AbstractAdapter or AbstractMysqlAdapter?
I experimented with this [here](https://github.com/Shopify/semian/pull/435/commits/eca4b8734641d2173a593ff2f02da397a2779beb), but decided this wasn't a sensible option. Active Record translates certain errors we care about to `ActiveRecord::StatementInvalid`, and we need to look at the underlying error to see if it's something that should trigger the Semian circuit breaker (e.g. a timeout error). We'd need to bake in references to the various adapter classes (e.g. `Trilogy::TimeoutError`, `Mysql2::Error`), so this doesn't make sense.

---

I've tested this adapter out with Shopify/shopify and have a passing CI build: https://buildkite.com/shopify/shopify-selective-tests/builds/969555

### Feedback

Dealing with `half_open_resource_timeout` is unfortunately still a bit hacky, even at the adapter layer. If anyone has thoughts on how to do this better, I'm all ears.

Basically, the Trilogy adapter forwards configs to the client, one of which is the `read_timeout`. In the context of `with_resource_timeout`, we might be connecting to the db, in which case `connection` is nil. Consequently, the only way I see being able to set a new (temp) `read_timeout` is to mutate the `@config` so that the newly instantiated client gets created with the timeout we want. If we have the `connection` already, we can reach directly to the connection (the Trilogy client) and modify the `read_timeout`.

In the ensure block, we restore the original timeout on the `connection`, and I'm assuming we should also reset the `@config`.

This is even more hacky because if no `read_timeout` is set in the config, we need to use the client's default (0), which I've hardcoded into the adapter, but could potentially change on Trilogy's end down the road...

### DIFF
--- a/test/adapters/activerecord_trilogy_adapter_test.rb
+++ b/test/adapters/activerecord_trilogy_adapter_test.rb
@@ -373,6 +373,12 @@ module ActiveRecord
         assert_equal(2, raw_connection.write_timeout)
       end
 
+      def test_trilogy_default_read_timeout
+        client = Trilogy.new(@configuration.slice(:username, :host, :port))
+
+        assert_equal(0, client.read_timeout)
+      end
+
       def test_circuit_open_errors_do_not_trigger_the_circuit_breaker
         @proxy.down do
           ERROR_THRESHOLD.times do


gemfiles/trilogy.gemfile
```diff
@@ -0,0 +1,18 @@
...
+gem "rake"
+gem "rake-compiler"
+gem "minitest"
+gem "mocha"
+gem "toxiproxy"
+gem "webrick"
+
+gem "trilogy", github: "github/trilogy", branch: "main", glob: "contrib/ruby/*.gemspec"
+gem "activerecord-trilogy-adapter",
+  github: "adrianna-chang-shopify/activerecord-trilogy-adapter",
```
eileencodes, line 14: Should this point at github/activerecord-trilogy-adapter now that those changes are merged?

eileencodes (commented):
I left one comment where I think we need to change the repo.

Thank you for the great explanation on the route you chose and why, makes a lot of sense. I don't love how mysql2 and trilogy are so different here but I understand that there's not a better option right now. I know we're working on reconnects so we should revisit this implementation later on.

gemfiles/trilogy.gemfile
adrianna-chang-shopify: Yes, thanks forgot to update both Gemfiles 🤦‍♀️

paarthmadan (approved):
Nice, the tests read really well 🙌

lib/semian/trilogy_adapter.rb
```diff
@@ -0,0 +1,120 @@
...
+      end
+
+      ResourceBusyError = Class.new(SemianError)
+      CircuitOpenError = Class.new(SemianError)
+    end
+  end
+end
+
+module Semian
+  module TrilogyAdapter
```
casperisfine, line 26: I think we should call this `Semian::ActiveRecordTrilogyAdapter`.

casperisfine (approved):

lib/semian/trilogy_adapter.rb
miry: That a good point.

https://github.com/Shopify/semian/pull/468/files#diff-d09ea66f8227784ff4393d88a19836f321c915ae10031d16c93d67e6283ab55fR22

I wonder what will be in future name of adapter when we implement semian for pure trilogy.

PR #479, opened 2023-02-16
POC for per-request HTTP resource naming

pjambet:
POC for #480

---

A few notes about the POC, in no particular order:

- One of the big issues with the way the code works right now is that in the context of a single request (or a single connection opening), we execute the configuration block multiple times, which is unnecessary. Memoization was given us that "for free", now we'd have to move things around to reach what I'd consider the ideal goal: "a single request/connection => a single execution of the configuration block".
- The whole `deterministic: false` is how I ended up with it, but this is in no way trying to be prescriptive
- I used the `time` method to get a sense of the performance difference of memoization vs no memoization when running the acquire resource method. For my use case where I'm making an API call that will be in the best case scenario in the 50-100ms range, and often slower than 1s, the performance cost of calling a block per request seems to be marginal (~1us vs ~10us on my local machine)

---

TODO list:

- [x] Add tests to show the desired behavior

### DIFF
--- a/lib/semian/adapter.rb
+++ b/lib/semian/adapter.rb
@@ -12,7 +12,7 @@ module Semian
       return @semian_resource if @semian_resource
 
       options ||= semian_options
-      name = semian_identifier(options: options)
+      name = semian_identifier
 
       case options
       when false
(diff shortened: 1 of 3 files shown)


pjambet:
Closing in favor of #485

PR #485, opened 2023-04-24
Support non-deterministic resource configuration

pjambet:
Supersedes #479


lib/semian/adapter.rb
```diff
@@ -53,7 +59,11 @@ def semian_options
       return @semian_options if defined? @semian_options
 
       options = raw_semian_options
-      @semian_options = options && options.map { |k, v| [k.to_sym, v] }.to_h
+
+      symbolized_options = options && options.map { |k, v| [k.to_sym, v] }.to_h
+      symbolized_options.tap do
+        @semian_options = symbolized_options if (symbolized_options || {}).fetch(:deterministic, true)
```
miry, line 65: It looks a bit complex line for me.

I think `symbolized_options` could be `nil`, only when `options` is `nil`.
In other cases `symbolized_options` is a Hash.

Also I have doubts of using `tap` in this section of changes. 
Can you help me to understand, why `tap` is required?

Do you know if it is possible to avoid this block

lib/semian/adapter.rb
```diff
@@ -53,7 +59,11 @@ def semian_options
       return @semian_options if defined? @semian_options
 
       options = raw_semian_options
-      @semian_options = options && options.map { |k, v| [k.to_sym, v] }.to_h
+
+      symbolized_options = options && options.map { |k, v| [k.to_sym, v] }.to_h
```
miry, line 63: Would it better to symbolize Hash keys on initialization part for all adapters or HTTP one?
My goal here to have less operations.

Probably it is a good task for a separate PR.

miry:
There are no blockers from my side.
I would wait for @byroot review as well.

miry (approved):

lib/semian/net_http.rb
```diff
@@ -91,16 +91,38 @@ def disabled?
...
+      with_cleared_non_deterministic_options do
+        return super if disabled?
+
+        acquire_semian_resource(adapter: :http, scope: :query) do
+          handle_error_responses(super)
+        end
+      end
+    end
+
+    def with_cleared_non_deterministic_options
```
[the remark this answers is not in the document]
pjambet: Yes, I could have sworn I made it private. Will move it.

lib/semian/adapter.rb
pjambet: Agreed, it does look overly complex.

The reasoning behind `tap` was to get an easy way to always return `symbolized_options` (regardless of the value for `:deterministic`). The alternative version I started with (and ended up refactoring to the `tap` version) looked like this:

```ruby
symbolized_options = options && options.transform_keys(&:to_sym)
@semian_options = symbolized_options if (symbolized_options || {}).fetch(:deterministic, true)
symbolized_options
```

The `symbolized_options || {}` piece was for cases where `raw_semian_options` returns `false` or `nil`.

With all that said, would you prefer a `tap`-less, and without the `|| {}` thingy, like this: 

```
symbolized_options = options && options.transform_keys(&:to_sym)
@semian_options = symbolized_options if !symbolized_options || symbolized_options.fetch(:deterministic, true)
symbolized_options
```

pjambet:
@byroot I do like `:dynamic`, with a default to `false`. Will make that change.

@miry I will update the changelog and the examples as you suggested

lib/semian/adapter.rb
```diff
@@ -53,7 +59,11 @@ def semian_options
       return @semian_options if defined? @semian_options
 
       options = raw_semian_options
-      @semian_options = options && options.map { |k, v| [k.to_sym, v] }.to_h
+
+      symbolized_options = options && options.transform_keys(&:to_sym) # rubocop:disable Style/SafeNavigation
```
pjambet, line 63: Rubocop was incorrectly flaggin this line with:

> Style/SafeNavigation: Use safe navigation (&.) instead of checking if an object exists before calling the method.

But that wouldn't work when `options` is `false`.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+* New `dynamic` options, allowing a per-request resource configuration with the HTTP adapter (#485)
```
miry, line 3:
```suggestion
* New `dynamic` options, allowing a per-request resource configuration with the HTTP adapter. (#485)
```

pjambet: Updated 👍

pjambet:
Just rebased because of the conflicts in `CHANGELOG.md` which was recently updated on master

PR #487, opened 2023-04-27
Also test trilogy adapter through activerecord

etiennebarrie:
Follow-up to #486 

Since the adapter is in Active Record now, we can test Trilogy through Active Record too. So I just extracted the rails tests and used them both for the mysql2 and trilogy adapter.

Actually at first, I thought maybe the `semian/rails` code isn't too related to the specific adapter, so maybe there's no need to test it for each adapter/client semian supports, but then I saw the code from https://github.com/Shopify/semian/pull/395 which is specific to mysql2.

By adding a share test we could see the special case made for Mysql2 and reproduce it for Trilogy, but looking more into it, I don't think it's ever used:

```
>> ActiveRecord::ConnectionAdapters::Mysql2Adapter.method(:new_client)
=> #<Method: ActiveRecord::ConnectionAdapters::Mysql2Adapter.new_client(config) /Users/etienne/.gem/ruby/3.2.2/gems/activerecord-7.0.4.3/lib/active_record/connection_adapters/mysql2_adapter.rb:43>
>> ActiveRecord::ConnectionAdapters::Mysql2Adapter.method(:new_client).super_method
=> #<Method: Semian::Rails::ClassMethods#new_client(config) /Users/etienne/.gem/ruby/3.2.2/gems/semian-0.18.0/lib/semian/rails.rb:13>
```

and `ActiveRecord::ConnectionAdapters::Mysql2Adapter.new_client` doesn't call `super` so those exceptions are never rescued/re-raised.

Anyway, I can look into this later but this may still be worth it. Let me know what you think.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -117,7 +117,8 @@ jobs:
           - grpc
           - mysql2
           - net_http
-          - rails
+          - rails_mysql2
+          - rails_trilogy
           - redis_4
           - redis_5
           - redis_client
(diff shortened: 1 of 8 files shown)


miry (approved):

adrianna-chang-shopify:
❤️

PR #494, opened 2023-05-15
Active Record Trilogy Adapter needs to intercept `#raw_execute`

adrianna-chang-shopify:
For: https://github.com/Shopify/rails-infra/issues/437

Changes made to Rails (https://github.com/rails/rails/commit/f69bbcbc0752ca5d5af327d55922614a26f5c7e9) mean that most database methods no longer use the public `#execute` method and go through `#raw_execute` instead.

This PR also ensures that we're testing with the Trilogy adapter that lives in Rails (and no longer the GitHub Trilogy adapter), and the latest released version of the Trilogy client.

### DIFF
--- a/test/adapters/activerecord_trilogy_adapter_test.rb
+++ b/test/adapters/activerecord_trilogy_adapter_test.rb
@@ -228,7 +228,9 @@ module ActiveRecord
 
       def test_resource_timeout_on_connect
         @proxy.downstream(:latency, latency: 500).apply do
-          background { @adapter.connect! }
+          background do
+            assert_raises(TrilogyAdapter::CircuitOpenError) { @adapter.connect! }
+          end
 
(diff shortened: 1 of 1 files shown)


eileencodes (approved):
I verified that in 7.0 `execute` calls `raw_execute` so we don't risk calling the wrong method in applications not running on main / 7.1.x.

adrianna-chang-shopify:
>  verified that in 7.0 execute calls raw_execute so we don't risk calling the wrong method in applications not running on main / 7.1.x.

Oh yes thanks for verifying that @eileencodes 👍 I'm wondering if we want to test against multiple versions of Rails (e.g. 7 / 7.1 once release + edge) moving forward..? Might also make sense to drop the lockfiles for rails so that we're always testing Semian against the _latest_ edge of Rails (although to be fair, Shopify Core is a good safety net for issues with Rails edge + Semian).

PR #496, opened 2023-05-19
Use ActiveRecord::ConnectionNotEstablished as base error for TrilogyAdapter

casperisfine:
One part of the design of semian is that the exception it raises should ideally be caught by naive `rescue SomeClient::BaseConnectionError`.

The reason is that semian raise a fast error when it thinks it's likely using the resource right now would lead to a slow network error of some kind.

FYI: @adrianna-chang-shopify (for when you get back from 🌴), @eileencodes, @pegelston 

Note:

I see `acquire_semian_resource` does:

```ruby
    rescue ActiveRecord::StatementInvalid => error
      if error.cause.is_a?(Trilogy::TimeoutError)
```

Which may indicate that the upstream adapter is also not properly translating the client timeout into a proper network related error (`StatementInvalid` being the kitchen sink). I'll have a check upstream as well.

### DIFF
--- a/lib/semian/activerecord_trilogy_adapter.rb
+++ b/lib/semian/activerecord_trilogy_adapter.rb
@@ -9,7 +9,7 @@ module ActiveRecord
     class TrilogyAdapter
       ActiveRecord::ActiveRecordError.include(::Semian::AdapterError)
 
-      class SemianError < StatementInvalid
+      class SemianError < ConnectionNotEstablished
         def initialize(semian_identifier, *args)
           super(*args)
           @semian_identifier = semian_identifier


etiennebarrie:
> One part of the design of semian is that the exception it raises should ideally be caught by naive `rescue SomeClient::BaseConnectionError`.

Sorry I don't understand. What's the base connection error for ActiveRecord then? We have both `StatementInvalid < ActiveRecordError` and `ConnectionNotEstablished < ActiveRecordError`. `ConnectionNotEstablished` is raised when failing to connect, and `ConnectionFailed < QueryAborted < StatementInvalid` is raised when the connection was previously establish but now fails.

We can see a bunch of places using both these classes in Active Record, e.g. [`ActiveRecord::ConnectionAdapters::AbstractAdapter#retryable_connection_error?`](https://github.com/rails/rails/blob/d137b10f946ff78fac8fe203d5aeaf2bb4c3a1d9/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L1052-L1054) or [considering a transaction rolled back](https://github.com/rails/rails/blob/d137b10f946ff78fac8fe203d5aeaf2bb4c3a1d9/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb#L411-L412).

So it's unclear to me why this one would be better.

On another note but related, `Semian::ActiveRecordTrilogyAdapter` declares `ActiveRecord::ConnectionNotEstablished` as one of the ["resource exceptions"](https://github.com/Shopify/semian/blob/6871dd82dcd3e4bec4d09a4249b7dcc9f25bed17/lib/semian/activerecord_trilogy_adapter.rb#L95-L97). If we swap the base class, but maybe even if we don't, shouldn't we add `ConnectionFailed` to the resource exceptions, which are [used in `CircuitBreaker`](https://github.com/Shopify/semian/blob/6871dd82dcd3e4bec4d09a4249b7dcc9f25bed17/lib/semian/circuit_breaker.rb#L46-L48)?

--- 

And I think that's your last point, why do we need to redefine `acquire_semian_resource` to mark the resource as failed? With your upstream change, we could let `CircuitBreaker` mark the resource as failed and [Semian::Adapter store the semian identifier](https://github.com/Shopify/semian/blob/6871dd82dcd3e4bec4d09a4249b7dcc9f25bed17/lib/semian/adapter.rb#L53-L54) by making `ActiveRecord::AdapterTimeout` a "resource exception". Then we won't need to redefine `acquire_semian_resource` in `ActiveRecordTrilogyAdapter` at all.

adrianna-chang-shopify (approved):
LGTM 👍 Will be nice to get rid of the special handling in `acquire_semian_resource` by making `AdapterTimeout` a resource exception following the upstream changes.

We should also definitely add `ConnectionFailed` to that list, that was an oversight on my part that we weren't handling that class 😅

PR #497, opened 2023-05-31
AR Trilogy Adapter: Handle AdapterTimeout and ConnectionFailed errs

adrianna-chang-shopify:
See https://github.com/Shopify/semian/pull/497

This PR ensures that `ActiveRecord::AdapterTimeout` errors are handled for the AR Trilogy Adapter. As of https://github.com/rails/rails/pull/48272/commits/785f65680b55a33e02cd5614e709b99b1fe6218b, TimeoutErrors are translated to AdapterTimeout errors in Rails. As such, we no longer need special handling in `#acquire_semian_resource` for such errors.

This commit also ensures that ActiveRecord::ConnectionFailed errors are classified as resource exceptions. These are network-related exceptions and should be handled by Semian. I've added a corresponding test.

cc @casperisfine @etiennebarrie

### DIFF
--- a/lib/semian/activerecord_trilogy_adapter.rb
+++ b/lib/semian/activerecord_trilogy_adapter.rb
@@ -82,18 +82,12 @@ module Semian
 
     private
 
-    def acquire_semian_resource(**)
-      super
-    rescue ActiveRecord::StatementInvalid => error
-      if error.cause.is_a?(Trilogy::TimeoutError)
-        semian_resource.mark_failed(error)
-        error.semian_identifier = semian_identifier
-      end
-      raise
-    end
-
     def resource_exceptions
-      [ActiveRecord::ConnectionNotEstablished]
+      [
+        ActiveRecord::AdapterTimeout,
+        ActiveRecord::ConnectionFailed,
+        ActiveRecord::ConnectionNotEstablished,
+      ]
     end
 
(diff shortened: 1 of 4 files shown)


lib/semian/activerecord_trilogy_adapter.rb
```diff
@@ -82,18 +82,25 @@ def with_resource_timeout(temp_timeout)
...
     private
 
-    def acquire_semian_resource(**)
-      super
-    rescue ActiveRecord::StatementInvalid => error
-      if error.cause.is_a?(Trilogy::TimeoutError)
-        semian_resource.mark_failed(error)
-        error.semian_identifier = semian_identifier
+    # As of Rails 7.1, TimeoutErrors are translated to ActiveRecord::AdapterTimeout
+    if ActiveRecord.version < Gem::Version.new("7.1.0.alpha")
```
casperisfine, line 86: Do we really care about supporting 7.0?

casperisfine (approved):

lib/semian/activerecord_trilogy_adapter.rb
adrianna-chang-shopify: Yeah probably not, I forgot that Trilogy is only available as of Rails 7.1 anyways 😅

PR #514, opened 2023-12-11
Remove upper bound of 1 for quotas

catlee:


ext/semian/resource.c
```diff
@@ -289,7 +289,7 @@ check_quota_arg(VALUE quota)
     if (TYPE(quota) != T_FIXNUM && TYPE(quota) != T_FLOAT) {
       rb_raise(rb_eTypeError, "expected decimal type for quota");
     }
-    if (NUM2DBL(quota) <= 0 || NUM2DBL(quota) > 1) {
+    if (NUM2DBL(quota) <= 0) {
```
dalehamel, line 292: a quota ratio > 1 should have the same effect as a quota of 1.

This would mean that there as many (or more) available tickets as there are protected resources.

I don't see any benefit to removing this restriction, can you elaborate on your reasoning for removing it?

PR #560, opened 2024-10-16
Explicit support for acquire_semaphore/release_semaphore in semian resources

kovyrin:
We're working on using Semian in a project that uses Ruby fibers and we wanted to implement MySQL bulkheads in a way where we would share the same semian resource semaphore (ticket) between multiple fibers. To be able to do that, we need to be able to acquire a semaphore without passing a block into the call and then release the semaphore later at the time of our choosing.

This PR extracts the semaphore acquisition and release logic into separate methods making the described scenario possible.

### DIFF
--- a/ext/semian/resource.c
+++ b/ext/semian/resource.c
@@ -5,9 +5,6 @@ ID id_wait_time;
 ID id_timeout;
 int system_max_semaphore_count;
 
-static VALUE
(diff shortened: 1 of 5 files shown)


casperisfine (approved):

kovyrin:
Squashed all the commits into one, will merge when it is green on CI.

PR #567, opened 2025-01-03
Correct broken blog post link in README.md

rnewman:


iandelahorne:
@rnewman this got fixed in #568 I believe

rnewman:
@chastell and I — great minds think alike, and ours too :P

PR #606, opened 2025-05-20
Add lumping_interval Feature for Semian Circuit Breakers

abishanan-shopify:
# Changes: 

- Adjust the `push_time`, `create_circuit_breaker`, and `initialize` functions to incorporate lumping intervals
- Implement tests for respective changes

# Rationale:

When it comes to real-world applications, errors can sometimes occur in short-term bursts due to **insignificant** network disruptions, outages, or other small "blips" in the network that rapidly self-resolve.

Under the current implementation, each error, no matter how close together in time, counts individually. This means that a short burst of errors, no matter how far apart they are in time, will affect the circuit breaker by accumulating the error count. This has a few issues:

- This makes the circuit breaker more sensitive to high-frequency error accumulations
- It makes it harder to identify larger sustained problems beyond small instances of noise
- Small noise has the ability to break the circuit entirely
- The system is impacted at a larger scale, since the circuit breaker can open and close for insignificant issues - decreasing resiliency

# Implementation:

- Added a "lump interval" which groups multiple errors that occur within a set timeframe (e.g., 1 second)
- The "lump interval" is not based on fixed time windows but rather based on the timestamps of the latest error (see graphic below)
- This makes error counter less sensitive to error bursts and improves the stability and user experience of those using services involving circuit breakers

lumping__interval

### DIFF
--- a/README.md
+++ b/README.md
@@ -532,7 +532,7 @@ There are four configuration parameters for circuit breakers in Semian:
   again, that is to start accepting all requests to the circuit.
 - **half_open_resource_timeout**. Timeout for the resource in seconds when
   the circuit is half-open (supported for MySQL, Net::HTTP and Redis).
-- **lumping_interval**. Interval for timeframe of errors to be lumped and recorded as one.
+- **lumping_interval**. If provided, errors within this timeframe (in seconds) will be lumped and recorded as one.
 
 It is possible to disable Circuit Breaker with environment variable
 `SEMIAN_CIRCUIT_BREAKER_DISABLED=1`.


lib/semian/circuit_breaker.rb
```diff
@@ -26,6 +27,11 @@ def initialize(name, exceptions:, success_threshold:, error_threshold:,
       @error_timeout = error_timeout
       @exceptions = exceptions
       @half_open_resource_timeout = half_open_resource_timeout
+      @lumping_interval = lumping_interval
+
+      if @lumping_interval >= @error_threshold_timeout && @lumping_interval > 0
```
AbdulRahmanAlHamali, line 32: do we need the `&&` here? I think what would make more sense is to prevent negative values for all parameters instead

lib/semian/circuit_breaker.rb
```diff
@@ -137,11 +143,14 @@ def push_error(error)
 
     def push_time
       time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+
       if error_threshold_timeout_enabled
         @errors.reject! { |err_time| err_time + @error_threshold_timeout < time }
       end
 
-      @errors << time
+      if @lumping_interval == 0 || @errors.empty? || @errors.last < time - @lumping_interval
```
AbdulRahmanAlHamali, line 151: is the first condition necessary here? 🤔

AbdulRahmanAlHamali (approved):
LGTM!

README.md
```diff
@@ -513,22 +516,23 @@ all workers on a server.
...
+- **error_threshold_timeout_enabled**. If set to false it will disable
   the time window for evicting old exceptions. `error_timeout` is still used and
   will reset the circuit. Defaults to `true` if not set.
-* **success_threshold**. The amount of successes on the circuit until closing it
+- **success_threshold**. The amount of successes on the circuit until closing it
   again, that is to start accepting all requests to the circuit.
-* **half_open_resource_timeout**. Timeout for the resource in seconds when
+- **half_open_resource_timeout**. Timeout for the resource in seconds when
   the circuit is half-open (supported for MySQL, Net::HTTP and Redis).
+- **lumping_interval**. Interval for timeframe of errors to be lumped and recorded as one.
```
AbdulRahmanAlHamali, line 535:
```suggestion
- **lumping_interval**. If provided, errors within this interval will be lumped and recorded as one.
```

AbdulRahmanAlHamali (approved):
Approved, but suggested a clarification in the README

abishanan-shopify:
> Regarding the visual, I love it! But I also feel like it's sending the wrong message
> 
> image
> The feature is helping us not open the circuit too soon on a network blip. So I feel like the visuals should show the errors happening within the first second, followed by a period of calmness. And how the lumping_interval helped us not open the circuit in that situation

Updated with a more relevant visual! 👍

README.md
Aguasvivas22: would also be good to mention that it's in seconds.

Issue #633, opened 2025-06-03
Add configuration validation for Semian


AbdulRahmanAlHamali:
Currently, you can put whatever values you wish for a circuit breaker. You can set values that would cause the circuit to never open, etc.

We should add configuration validation to make sure configured semians are effective. This would mean crashing if they're not effective, and accordingly, this should be a major version bump

PR #646, opened 2025-06-16
Setup Semian Configuration Validation

abishanan-shopify:
Addressing this issue: https://github.com/Shopify/semian/issues/633

# Changes

Created `ConfigurationValidator` class that handles all validations and constraints for the following Semian config values:

- `name`: Must be string / symbol and cannot be a duplicate Semian name of one that exists in `Semian.resources`
- `:circuit_breaker` & `:bulkhead`: Cannot both be disabled
- `:tickets` & `:quota`: If bulkhead, can only have `:tickets` or `:quota`, cannot be both
   - `:tickets`: Must be between 0 and `Semian::MAX_TICKETS`, inclusive
   - `:quota`: Must be a ratio between 0 and 1, inclusive
- `:success_threshold`, `:error_threshold` & `:error_timeout`: If circuit breaker, validate that these `required` params exist.
- `:success_threshold` & `:error_threshold`: If circuit breaker, validate that they are positive numbers
- `:error_timeout`, `:error_threshold_timeout`, `:half_open_resource_timeout`, `:lumping_interval`: If circuit breaker and exists, validate that they are non-negative.

We have also added constraints to the timeout values to ensure that Semians don't end up in weird scenarios (never opens, closes, never half-opens, etc.):
- `lumping_interval * (error_threshold - 1) <= error_threshold_timeout`

We also added a new flag - `force_config_validation` to pass as an argument to Semian to force configuration validation at runtime if desired (otherwise it will be logged verbosely in I/O)

Finally, we added test cases to address each of the above constraints, and modified non-validation related test cases to conform to `ConfigurationValidator` class. 

# Rationale

We are looking to add validations to configuration options on Semians, to ensure that invalid or otherwise impossible Semian configuration are not created (result in an Exception raised) and do not persist in production. This will help alleviate any cases of incidents or investigations where Semians do not open due to impossible conditions.

# Why are some validations required and others are not?

This is to avoid regression in our previous validation logic -- we want to make sure that we don't unexpectedly crash developers' Semians, but we don't allow things that we have previously disallowed.

This includes:
- Requiring `success_threshold, error_threshold, and error_timeout` to be required in circuit breakers
- Requiring either `tickets` or `quota` to be required in bulkheads
- Making sure `tickets` are within bounds `(0, MAX_TICKETS]` (`tickets != MAX_TICKETS` was not previously enforced so we use flag instead)
- Making sure `quota` is within bounds `(0, 1]` (`quota != 1` was not previously enforced so we use the flag instead)

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,10 +164,6 @@ if it is set to `false`, it will log misconfigured parameters verbosely in outpu
 If not specified, it will use `Semian.default_force_config_validation` as
 the flag.
 
-**IMPORTANT**: Future releases will deprecate and eventually remove this flag, making explicit configuration
-validation the default. **Please make changes as soon as possible** to ensure that your Semian does not break
-in a future version.
-
 ##### Migration Strategy for Force Config Validation
 
 When migrating to use `force_config_validation: true`, follow these steps:


lib/semian/configuration_validator.rb
```diff
@@ -0,0 +1,133 @@
...
+
+    def validate_timeouts!
+      error_timeout = @configuration[:error_timeout]
+      error_threshold_timeout = @configuration[:error_threshold_timeout]
+      error_threshold = @configuration[:error_threshold]
+      lumping_interval = @configuration[:lumping_interval]
+      half_open_resource_timeout = @configuration[:half_open_resource_timeout]
+
+      unless error_timeout && error_timeout >= 0
+        raise ArgumentError, "error_timeout must be non-negative"
```
AbdulRahmanAlHamali: TODO: should we make sure it's `positive` instead?

lib/semian/configuration_validator.rb
```diff
@@ -0,0 +1,133 @@
...
+      end
+
+      unless error_threshold_timeout.nil? || error_threshold_timeout >= 0
```
AbdulRahmanAlHamali: same

lib/semian/configuration_validator.rb
```diff
@@ -0,0 +1,133 @@
...
+        raise ArgumentError, "error_threshold_timeout must be non-negative"
+      end
+
+      unless half_open_resource_timeout.nil? || half_open_resource_timeout >= 0
```
AbdulRahmanAlHamali: same

lib/semian/configuration_validator.rb
```diff
@@ -0,0 +1,133 @@
...
+        raise ArgumentError, "half_open_resource_timeout must be non-negative"
+      end
+
+      unless lumping_interval.nil? || lumping_interval >= 0
```
AbdulRahmanAlHamali: ok, so here 0 means "unset", which maybe makes sense

lib/semian/configuration_validator.rb
```diff
@@ -0,0 +1,133 @@
...
+        raise ArgumentError, "lumping_interval must be non-negative"
+      end
+
+      unless lumping_interval.nil? || error_threshold_timeout.nil? || lumping_interval * (error_threshold - 1) <= error_threshold_timeout
```
AbdulRahmanAlHamali: needs a comment to explain the -1

lib/semian/configuration_validator.rb
```diff
@@ -0,0 +1,133 @@
...
+        raise ArgumentError, "constraint violated: lumping_interval * (error_threshold - 1) <= error_threshold_timeout, got lumping_interval: #{lumping_interval}, error_threshold: #{error_threshold}, error_threshold_timeout: #{error_threshold_timeout}"
+      end
+
+      unless half_open_resource_timeout.nil? || half_open_resource_timeout <= error_timeout
```
AbdulRahmanAlHamali: Discussed in person: half_open does not care about thresholds. The half_open_resource_timeout should be less than the original resource timeout, but that's the only restriction

lib/semian/configuration_validator.rb
```diff
@@ -0,0 +1,133 @@
...
+        raise ArgumentError, "constraint violated: half_open_resource_timeout <= error_timeout, got half_open_resource_timeout: #{half_open_resource_timeout}, error_timeout: #{error_timeout}"
+      end
+
+      unless half_open_resource_timeout.nil? || error_threshold_timeout.nil? || half_open_resource_timeout <= error_threshold_timeout
```
AbdulRahmanAlHamali: same as above

lib/semian/configuration_validator.rb
```diff
@@ -0,0 +1,133 @@
...
+      end
+    end
+
+    def validate_tickets!(tickets)
+      unless tickets.is_a?(Integer) && tickets >= 0 && tickets <= Semian::MAX_TICKETS
+        raise ArgumentError, "ticket count must be a non-negative integer and less than #{Semian::MAX_TICKETS}"
+      end
+    end
+
+    def validate_resource_name!
```
AbdulRahmanAlHamali, line 212: good idea!

lib/semian/configuration_validator.rb
```diff
@@ -0,0 +1,133 @@
...
+
+    def validate_bulkhead_configuration!
+      return if ENV.key?("SEMIAN_BULKHEAD_DISABLED")
+      return unless @configuration.fetch(:bulkhead, true)
+
+      tickets = @configuration[:tickets]
+      quota = @configuration[:quota]
+
+      if tickets.nil? && quota.nil?
+        raise ArgumentError, "Semian configuration require either the :tickets or :quota parameter, you provided neither"
```
AbdulRahmanAlHamali:
```suggestion
        raise ArgumentError, "Bulkhead configuration require either the :tickets or :quota parameter, you provided neither"
```

lib/semian/configuration_validator.rb
```diff
@@ -0,0 +1,133 @@
...
+      end
+
+      if tickets && quota
+        raise ArgumentError, "Semian configuration require either the :tickets or :quota parameter, you provided both"
```
AbdulRahmanAlHamali:
```suggestion
        raise ArgumentError, "Bulkhead configuration require either the :tickets or :quota parameter, you provided both"
```

lib/semian/configuration_validator.rb
```diff
@@ -0,0 +1,133 @@
...
+        raise ArgumentError, "constraint violated: half_open_resource_timeout <= error_timeout, got half_open_resource_timeout: #{half_open_resource_timeout}, error_timeout: #{error_timeout}"
+      end
+
+      unless half_open_resource_timeout.nil? || error_threshold_timeout.nil? || half_open_resource_timeout <= error_threshold_timeout
+        raise ArgumentError, "constraint violated: half_open_resource_timeout <= error_threshold_timeout, got half_open_resource_timeout: #{half_open_resource_timeout}, error_threshold_timeout: #{error_threshold_timeout}"
+      end
+    end
+
+    def validate_quota!(quota)
+      unless quota.is_a?(Numeric) && quota > 0 && quota <= 1
```
AbdulRahmanAlHamali: should this be strictly less than 1?

lib/semian/configuration_validator.rb
```diff
@@ -0,0 +1,133 @@
...
+        raise ArgumentError, "quota must be a decimal between 0 and 1"
+      end
+    end
+
+    def validate_tickets!(tickets)
+      unless tickets.is_a?(Integer) && tickets >= 0 && tickets <= Semian::MAX_TICKETS
```
AbdulRahmanAlHamali: also strictly less than

AbdulRahmanAlHamali:
Note: CI on this PR (and others) have been failing because `example.com` has been acting funkily. We switched all usages to shopify.com instead. 

(Ideally we would put these changes in a separate PR, but they are sufficiently trivial)

iandelahorne:
edit: my previous comment was in error, but this should likely be a minor version bump when we release it

iandelahorne (approved):
Looks good, excited to see this in production later

adriangudas (approved):

PR #662, opened 2025-07-02
Fix docker compose tests

AbdulRahmanAlHamali:
Testing locally just doesn't work for two reasons:

# Mysql Update
docker compose does not have a pinned mysql version, and so when we tried to run it, it started pulling v9, which doesn't like mysql native password plugin.

So, we decided to fix forward and just fix our tests to run on v9, which requires a password by default. This meant a few changes:

1. Start using a password and SSL (but we set the mode to 3, meaning "no verify", to simplify our setup)
2. Stop using netcat to test if mysql is up, and use mysql-client instead

# HTTP Tests

The way HTTP tests work is by spinning a MockServer that is running from within the same semian container (convoluted, I know), but there is an expectation of the hostname to be `semian`, which just isn't the case when we're running tests. So I fix the hostname and also make it clearer

### DIFF
--- a/README.md
+++ b/README.md
@@ -917,7 +917,7 @@ If you make any changes to `.devcontainer/` you'd need to recreate the container
 Run tests in containers:
 
 ```shell
-$ docker-compose -f ./.devcontainer/docker-compose.yml run --rm test
+$ docker-compose -f ./.devcontainer/docker-compose.yml --profile test run --rm test
 ```
 
 Running Tests:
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -917,7 +917,7 @@ If you make any changes to `.devcontainer/` you'd need to recreate the container
 Run tests in containers:
 
 ```shell
-$ docker-compose -f ./.devcontainer/docker-compose.yml run --rm test
+$ docker-compose -f ./.devcontainer/docker-compose.yml --profile test run --rm test
```
AbdulRahmanAlHamali, line 920: you must specify the profile explicitly in podman

https://github.com/containers/podman-compose/issues/930

abishanan-shopify (approved):
LGTM!

PR #721, opened 2025-08-27
Fix unbound growth of consumers WeakRef Array

jhawthorn:
Previously we used an array of WeakRef to track the list of consumers of a give resource. Although the weakref would allow the referenced consumer to be collected by GC, the WeakRef itself consumes some memory and the array could grow unbounded.

@luke-gruber and I found this large array while profiling
@franciscojmc and I investigated and found the sources of the problem.

This PR applies two fixes for (essentially) the same problem

The first commit swaps the Array full of individual WeakRefs for a single WeakMap. This both avoids adding duplicate copies of the same adapter and leans on the RubyVM to prune stale references.

The second commit observes that previously, when using `dynamic: true`, [we never memoize `@semian_resource` on the adapter](https://github.com/Shopify/semian/blob/d8090bfc8473546ec1f633b6690c564dbb8292a2/lib/semian/adapter.rb#L28), and so there's no need to track it as a "consumer". Previously this meant we would continuously be registering duplicate copies of the same adapter as a consumer.

cc @tenderworks and @whatisinternet who have also looked in this area previously

### DIFF
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -199,7 +199,7 @@ module Semian
     # If consumer who retrieved / registered by a Semian::Adapter, keep track
     # of who the consumer was so that we can clear the resource reference if needed.
     consumer = args.delete(:consumer)
-    if consumer&.class&.include?(Semian::Adapter)
+    if consumer&.class&.include?(Semian::Adapter) && !args[:dynamic]
       consumer_set = (consumers[name] ||= ObjectSpace::WeakMap.new)
       consumer_set[consumer] = true
     end
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -200,8 +200,8 @@ module Semian
     # of who the consumer was so that we can clear the resource reference if needed.
     consumer = args.delete(:consumer)
     if consumer&.class&.include?(Semian::Adapter)
-      consumers[name] ||= []
-      consumers[name] << WeakRef.new(consumer)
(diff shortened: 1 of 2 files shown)


whatisinternet (approved):

nirmitparikh8:
@anagayan and I may have found a bug with this PR. While testing my PR which is rebased off of current sfr main, I found a possible bug with the most recent commit on the Semian branch ( this PR).  This was tested when the storefront gem file points semian to the main branch and not to release branch 0.24.0.

Testing procedure:
1. Navigate to `world/trees/root/src/areas/core/storefront`
2. Change [storefront gem file](https://github.com/shop/world/blob/main/areas/core/storefront/Gemfile#L27) to point semian library to main branch (`gem "semian", github: "Shopify/semian", branch: "main"` )
3. `bundle install`
4. `dev test test/unit/patches/semian_circuit_breaker_configuration_test.rb`

This gives this error for all semian tests in this [file](https://github.com/shop/world/blob/main/areas/core/storefront/test/unit/patches/semian_circuit_breaker_configuration_test.rb): `Minitest::UnexpectedError:Sonic::Db::TrilogyClient::ClientError: undefined method 'select!' for an instance of ObjectSpace::WeakMap `

@jhawthorn Just wanted to flag this issue to get your attention. How do we want to proceed?

Logs Screenshot:
Screenshot 2025-09-12 at 6 33 34 PM

PR #739, opened 2025-09-15
Add semian resource name to logs

nirmitparikh8:
- Add resource name to the semian configuration validator warning logs so that we can identify exactly what resource configuration needs to be fixed.

### DIFF
--- a/lib/semian/configuration_validator.rb
+++ b/lib/semian/configuration_validator.rb
@@ -32,7 +32,7 @@ module Semian
       if @force_config_validation
         raise ArgumentError, message
       else
-        Semian.logger.warn("[SEMIAN_CONFIG_WARNING]: #{message}")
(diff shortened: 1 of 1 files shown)


lib/semian/configuration_validator.rb
```diff
@@ -32,7 +32,7 @@ def raise_or_log_validation_required!(message)
       if @force_config_validation
         raise ArgumentError, message
       else
-        Semian.logger.warn("[SEMIAN_CONFIG_WARNING]: #{message}")
+        Semian.logger.warn("[SEMIAN_CONFIG_WARNING]: Semian Adapter #{@name}: #{message}")
```
AbdulRahmanAlHamali:
```suggestion
        Semian.logger.warn("[SEMIAN_CONFIG_WARNING]: Semian Resource #{@name}: #{message}")
```

AbdulRahmanAlHamali (approved):
approved with a suggestion

PR #740, opened 2025-09-15
Addressing variables that aren't thread safe

### LINKED ISSUE #566: Thread Safety
Hi,

Thanks!

kris-gaudel:
Fixes #566 

Replace instances of data structures that are not thread safe and migrates existing thread safe code to use the `concurrent-ruby` gem

Includes unit tests to ensure thread safe behaviour

### DIFF
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -241,7 +241,12 @@ module Semian
       consumers_for_resource = consumers[name]
       if consumers_for_resource
         consumers_for_resource.each_key(&:clear_semian_resource)
-        consumers[name] = nil
+        # Remove the entry by creating new WeakMap without this key
+        new_consumers = ObjectSpace::WeakMap.new
+        consumers.each_pair do |key, value|
+          new_consumers[key] = value unless key == name
+        end
+        @consumers = new_consumers
       end
     end
   end
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -238,8 +238,11 @@ module Semian
     resource = resources.delete(name)
     if resource
       resource.bulkhead&.unregister_worker
-      consumers_for_resource = consumers.delete(name) || ObjectSpace::WeakMap.new
-      consumers_for_resource.each_key(&:clear_semian_resource)
+      consumers_for_resource = consumers[name]
+      if consumers_for_resource
+        consumers_for_resource.each_key(&:clear_semian_resource)
+        consumers[name] = nil
+      end
     end
   end
 
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -4,6 +4,7 @@ require "forwardable"
 require "logger"
 require "weakref"
 require "thread"
+require "concurrent-ruby"
 
 require "semian/version"
 require "semian/instrumentable"
@@ -252,11 +253,11 @@ module Semian
 
   # Retrieves a hash of all registered resource consumers.
   def consumers
-    @consumers ||= {}
+    @consumers ||= Concurrent::Map.new
   end
 
   def reset!
-    @consumers = {}
+    @consumers = Concurrent::Map.new
     @resources = LRUHash.new
   end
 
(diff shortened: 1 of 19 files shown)


lib/semian.rb
```diff
@@ -233,7 +234,7 @@ def unregister(name)
     resource = resources.delete(name)
     if resource
       resource.bulkhead&.unregister_worker
-      consumers_for_resource = consumers.delete(name) || {}
+      consumers_for_resource = consumers.delete(name) || Concurrent::Map.new
```
anagayan:
```suggestion
      consumers_for_resource = consumers.delete(name) || ObjectSpace::WeakMap.new
```

anagayan: Additionally for the method `retrieve_or_register` you should modify this line:
```
consumer_set = (consumers[name] ||= ObjectSpace::WeakMap.new)
```
to use the thread safe method instead:
```
consumer_set = consumers.compute_if_absent(name) { ObjectSpace::WeakMap.new }
```

lib/semian/simple_integer.rb
```diff
@@ -26,14 +27,29 @@ def destroy
   end
 
   module ThreadSafe
-    class Integer < Simple::Integer
-      def initialize(*)
-        super
-        @lock = Mutex.new
+    class Integer
+      def initialize
+        @atom = Concurrent::Atom.new(0)
```
anagayan:
```suggestion
        @atom = Concurrent::AtomicFixnum.new(0)
```
We should use `Concurrent::AtomicFixnum` instead as it's optimized for numerical values

lib/semian/simple_sliding_window.rb
```diff
@@ -45,26 +46,50 @@ def resize_to(size)
...
+        @window_atom.value.empty?
+      end
 
-      def reject!(*)
-        @lock.synchronize { super }
+      def reject!(&block)
+        @window_atom.swap do |window|
+          window.reject(&block)
+        end
+      end
```
anagayan:
```suggestion
      def reject!(&block)
        @window_atom.swap do |window|
          window.reject(&block)
        end
        self
      end
```
Shouldn't this have the `self` here instead?

lib/semian/simple_sliding_window.rb
```diff
@@ -45,26 +46,50 @@ def resize_to(size)
   end
 
   module ThreadSafe
-    class SlidingWindow < Simple::SlidingWindow
-      def initialize(**)
-        super
-        @lock = Mutex.new
+    class SlidingWindow
```
anagayan, line 49: Do we not need the `resize_to` method? Looks like it's missing now

test/adapter_test.rb
```diff
@@ -122,6 +122,58 @@ def test_adapter_error_message
     assert_equal("[my_service] Same Prefix", error.message)
   end
 
+  def test_concurrent_consumer_registration
```
anagayan: Would like to see a couple more variations of tests:

- Consumers registration and unregistration at the same time
- A test for the subscribers change on Instrumentable

lib/semian/simple_sliding_window.rb
kris-gaudel: The resizing logic is implemented in the `push` function, but I will add a private `resize_to` method like the non-thread-safe version

lib/semian/simple_sliding_window.rb
kris-gaudel: The `self` would allow method chaining like the non-thread-safe version, I will add it

kris-gaudel:
`ObjectSpace::WeakMap` in 3.2 **does not** have a `delete` method: https://docs.ruby-lang.org/en/3.2/ObjectSpace/WeakMap.html

But in 3.3 it does https://docs.ruby-lang.org/en/3.3/ObjectSpace/WeakMap.html

PR #755, opened 2025-09-19
Make variables thread safe and introduce concurrent-ruby gem

### LINKED ISSUE #566: Thread Safety

kris-gaudel:
Fixes #566 

Replace instances of data structures that are not thread safe and migrates existing thread safe code to use the concurrent-ruby gem

Includes unit tests to ensure correctness, thread safety and triggering race conditions will be followed up in #756

### DIFF
--- a/gemfiles/activerecord_trilogy_adapter.gemfile
+++ b/gemfiles/activerecord_trilogy_adapter.gemfile
@@ -8,6 +8,7 @@ gem "minitest"
 gem "mocha"
 gem "toxiproxy"
 gem "webrick"
+gem "concurrent-ruby"
 
 gem "trilogy", "~> 2.4"
 gem "activerecord", github: "rails/rails"
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -201,7 +201,7 @@ module Semian
     # of who the consumer was so that we can clear the resource reference if needed.
     consumer = args.delete(:consumer)
     if consumer&.class&.include?(Semian::Adapter) && !args[:dynamic]
-      consumer_set = (consumers[name] ||= ObjectSpace::WeakMap.new)
+      consumer_set = consumers.compute_if_absent(name) { ObjectSpace::WeakMap.new }
       consumer_set[consumer] = true
     end
     self[name] || register(name, **args)
@@ -234,7 +234,7 @@ module Semian
     resource = resources.delete(name)
     if resource
       resource.bulkhead&.unregister_worker
-      consumers_for_resource = consumers.delete(name) || Concurrent::Map.new
+      consumers_for_resource = consumers.delete(name) || ObjectSpace::WeakMap.new
       consumers_for_resource.each_key(&:clear_semian_resource)
     end
   end
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -124,6 +124,8 @@ module Semian
   self.resources = LRUHash.new
   self.consumers = Concurrent::Map.new
 
+  @reset_mutex = Mutex.new
+
   def issue_disabled_semaphores_warning
     return if defined?(@warning_issued)
 
@@ -260,8 +262,10 @@ module Semian
   end
 
   def reset!
-    self.consumers = Concurrent::Map.new
-    self.resources = LRUHash.new
+    @reset_mutex.synchronize do
+      self.consumers = Concurrent::Map.new
+      self.resources = LRUHash.new
+    end
   end
 
   THREAD_BULKHEAD_DISABLED_VAR = :semian_bulkheads_disabled
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -4,6 +4,7 @@ require "forwardable"
 require "logger"
 require "weakref"
 require "thread"
+require "concurrent-ruby"
 
 require "semian/version"
 require "semian/instrumentable"
@@ -252,11 +253,11 @@ module Semian
 
   # Retrieves a hash of all registered resource consumers.
   def consumers
-    @consumers ||= {}
+    @consumers ||= Concurrent::Map.new
   end
 
   def reset!
-    @consumers = {}
+    @consumers = Concurrent::Map.new
     @resources = LRUHash.new
   end
 
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -111,6 +111,16 @@ module Semian
   self.default_permissions = 0660
   self.default_force_config_validation = false
 
+  def thread_safe?
+    return @thread_safe if defined?(@thread_safe)
+
+    @thread_safe = true
+  end
+
+  def thread_safe=(thread_safe)
+    @thread_safe = thread_safe
+  end
+
   self.resources = LRUHash.new
   self.consumers = Concurrent::Map.new
 
@@ -262,16 +272,6 @@ module Semian
   def reset!
     self.consumers = Concurrent::Map.new
     self.resources = LRUHash.new
-  end
-
-  def thread_safe?
-    return @thread_safe if defined?(@thread_safe)
-
-    @thread_safe = true
-  end
-
-  def thread_safe=(thread_safe)
-    @thread_safe = thread_safe
   end
 
   THREAD_BULKHEAD_DISABLED_VAR = :semian_bulkheads_disabled
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -234,7 +234,7 @@ module Semian
     resource = resources.delete(name)
     if resource
       resource.bulkhead&.unregister_worker
-      consumers_for_resource = consumers.delete(name) || {}
+      consumers_for_resource = consumers.delete(name) || Concurrent::Map.new
       consumers_for_resource.each_key(&:clear_semian_resource)
     end
   end
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -259,16 +259,6 @@ module Semian
     end
   end
 
-  # # Retrieves a hash of all registered resources.
-  # def resources
-  #   @resources ||= LRUHash.new
-  # end
-
-  # Retrieves a hash of all registered resource consumers.
-  # def consumers
-  #   @consumers ||= Concurrent::Map.new
-  # end
-
   def reset!
     self.consumers = Concurrent::Map.new
     self.resources = LRUHash.new
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -111,6 +111,8 @@ module Semian
   self.default_permissions = 0660
   self.default_force_config_validation = false
 
+  # We only allow disabling thread-safety for parts of the code that are on the hot path.
+  # Since locking there could have a significant impact. Everything else is enforced thread safety
   def thread_safe?
     return @thread_safe if defined?(@thread_safe)
 
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -104,12 +104,15 @@ module Semian
   OpenCircuitError = Class.new(BaseError)
   SemaphoreMissingError = Class.new(BaseError)
 
-  attr_accessor :maximum_lru_size, :minimum_lru_time, :default_permissions, :namespace, :default_force_config_validation
+  attr_accessor :maximum_lru_size, :minimum_lru_time, :default_permissions, :namespace, :default_force_config_validation, :resources, :consumers
 
   self.maximum_lru_size = 500
   self.minimum_lru_time = 300 # 300 seconds / 5 minutes
   self.default_permissions = 0660
   self.default_force_config_validation = false
+
+  self.resources = LRUHash.new
+  self.consumers = Concurrent::Map.new
 
   def issue_disabled_semaphores_warning
     return if defined?(@warning_issued)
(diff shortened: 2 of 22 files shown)


test/adapter_test.rb
```diff
@@ -122,6 +122,203 @@ def test_adapter_error_message
...
+          clients_created << {
+            client: client,
+            resource: resource,
+            identifier: client.semian_identifier,
+            thread_id: i,
+          }
+        else
+          # Second half of threads: unregister existing resources
+          resource_to_unregister = initial_resources[(i - thread_count / 2) % initial_resources.size]
+          if Semian.resources[resource_to_unregister]
```
anagayan: We can probably remove this check since the `unregister` method does a check on resource as well

test/adapter_test.rb
```diff
@@ -122,6 +122,203 @@ def test_adapter_error_message
...
+    # Verify unregistration results
+    unregistrations_performed.each do |unregistration_data|
+      identifier = unregistration_data[:identifier]
+
+      assert_nil(Semian.resources[identifier], "Resource #{identifier} should be unregistered")
+      refute(Semian.consumers.key?(identifier), "Consumer #{identifier} should be removed")
+    end
+  end
+
+  def test_instrumentable_subscribers_change
```
anagayan, line 223: This test is doing a bunch of different things all at once. It's better to split it out into multiple tests

test/adapter_test.rb
```diff
@@ -122,6 +122,203 @@ def test_adapter_error_message
...
+    subscriber3 = ->(event, resource, scope, adapter, *payload) do
+      notifications_received << {
+        subscriber: :subscriber3,
+        event: event,
+        resource: resource,
+        scope: scope,
+        adapter: adapter,
+        payload: payload,
+      }
+    end
```
anagayan: Feels like we can make some kind of helper or loop for this so it's not duplicated

test/simple_integer_test.rb
```diff
@@ -51,6 +51,55 @@ def test_reset
 
       assert_equal(0, @integer.value)
     end
+
+    def test_concurrent_increment_and_access
```
anagayan: Now that the ThreadSafe implementation is no longer inheriting the `Simple::Integer`, it makes more sense to have these thread-safe tests in a file with a corresponding name.

Same comment goes for the SlidingWindow test file below.

On that note, we _may want to consider_ adding tests that prove something isn't thread safe (use of `Simple::*`) separately?

lib/semian/simple_integer.rb
```diff
@@ -26,14 +27,29 @@ def destroy
   end
 
   module ThreadSafe
-    class Integer < Simple::Integer
-      def initialize(*)
-        super
-        @lock = Mutex.new
+    class Integer
```
anagayan, line 30: Not crucial, but I noticed there is no test checking that the correct type is used based on the `thread_safe` value. You can consider adding a test in `semian_test.rb` that asserts the type based on the thread_safe value or `thread_safety_disabled` config.

For example
```ruby
def test_implementation_selection_respects_thread_safe_setting
  # Test with thread_safe = true (default)
  Semian.thread_safe = true
  resource = Semian.register(:test_threadsafe, 
    success_threshold: 1, error_threshold: 2, error_timeout: 5)
  
  # Access private instance variables to check implementation
  errors = resource.circuit_breaker.instance_variable_get(:@errors)
  successes = resource.circuit_breaker.instance_variable_get(:@successes)
  
  assert_instance_of(Semian::ThreadSafe::SlidingWindow, errors)
  assert_instance_of(Semian::ThreadSafe::Integer, successes)
  
  # Test with thread_safe = false
  Semian.thread_safe = false
  resource2 = Semian.register(:test_simple,
    success_threshold: 1, error_threshold: 2, error_timeout: 5)
  
  errors2 = resource2.circuit_breaker.instance_variable_get(:@errors)
  successes2 = resource2.circuit_breaker.instance_variable_get(:@successes)
  
  assert_instance_of(Semian::Simple::SlidingWindow, errors2)
  assert_instance_of(Semian::Simple::Integer, successes2)
```

test/simple_sliding_window_test.rb
```diff
@@ -29,17 +29,112 @@ def test_sliding_window_edge_falloff
...
+    end
+
+    threads.each(&:join)
+
+    assert_equal(@sliding_window.max_size, @sliding_window.size)
+
+    final_window_data = if @sliding_window.instance_variable_defined?("@window_atom")
+      @sliding_window.instance_variable_get("@window_atom").value
+    else
+      @sliding_window.instance_variable_get("@window")
```
anagayan: Seems like an unnecessary check - can we just have different tests if it's Simple vs ThreadSafe?

anagayan (approved):
Actual implementation looks good. I mainly left some comments to cleanup tests but they aren't blocking

test/simple_integer_test.rb
kris-gaudel: I separated the tests for the `Simple::*` and ThreadSafe implementations, but writing tests that trigger race conditions has been quite challenging. Could that be made into a separate follow-up issue?

gemfiles/activerecord_trilogy_adapter.gemfile
```diff
@@ -8,6 +8,7 @@ gem "minitest"
 gem "mocha"
 gem "toxiproxy"
 gem "webrick"
+gem "concurrent-ruby"
```
AbdulRahmanAlHamali, line 11: interesting. I didn't know we had different gemfiles for every adapter

lib/semian.rb
```diff
@@ -252,11 +253,11 @@ def resources
...
   # Retrieves a hash of all registered resource consumers.
   def consumers
-    @consumers ||= {}
+    @consumers ||= Concurrent::Map.new
   end
 
   def reset!
-    @consumers = {}
+    @consumers = Concurrent::Map.new
     @resources = LRUHash.new
```
AbdulRahmanAlHamali: don't we also want to fix `@resources`? Or is it already thread-safe?

lib/semian.rb
```diff
@@ -252,11 +253,11 @@ def resources
 
   # Retrieves a hash of all registered resource consumers.
   def consumers
-    @consumers ||= {}
+    @consumers ||= Concurrent::Map.new
```
AbdulRahmanAlHamali: I think this access pattern is still unsafe. If two threads come here at the same time, the first could create, assign, and start using `@consumers`, and then ther other could overwrite it

lib/semian.rb
```diff
@@ -233,7 +234,7 @@ def unregister(name)
     resource = resources.delete(name)
     if resource
       resource.bulkhead&.unregister_worker
-      consumers_for_resource = consumers.delete(name) || {}
+      consumers_for_resource = consumers.delete(name) || ObjectSpace::WeakMap.new
```
AbdulRahmanAlHamali, line 254: technically this does not have to be a weak map since it's not long-lived, but not a bad idea to stay consistent

lib/semian.rb
```diff
@@ -252,11 +253,11 @@ def resources
 
   # Retrieves a hash of all registered resource consumers.
   def consumers
-    @consumers ||= {}
+    @consumers ||= Concurrent::Map.new
   end
 
   def reset!
```
AbdulRahmanAlHamali, line 266: I think this access is also not thread-safe

lib/semian/instrumentable.rb
```diff
@@ -24,7 +26,7 @@ def notify(*args)
     private
 
     def subscribers
```
AbdulRahmanAlHamali: same as above. I don't think this access pattern is thread-safe

lib/semian/simple_sliding_window.rb
```diff
@@ -45,25 +46,56 @@ def resize_to(size)
   end
 
   module ThreadSafe
-    class SlidingWindow < Simple::SlidingWindow
-      def initialize(**)
-        super
-        @lock = Mutex.new
+    class SlidingWindow
+      extend Forwardable
```
AbdulRahmanAlHamali: what does this do? 🤔

lib/semian/simple_sliding_window.rb
```diff
@@ -45,25 +46,56 @@ def resize_to(size)
...
+        @window_atom.swap do |window|
+          window.reject(&block)
+        end
+        self
       end
 
-      def push(*)
-        @lock.synchronize { super }
+      def push(value)
+        @window_atom.swap do |window|
```
AbdulRahmanAlHamali: sounds very expensive. We do that on every push?

AbdulRahmanAlHamali (commented):
Thanks for the work Kris!

A couple of notes:

* I think there is still room for thread unsafety to break in when recreating the consumers (and other) objects
* I feel like the sliding window implementation is significantly less efficient than the original implementation. Note that this sliding window is called on the hot path, so it needs to be as efficient as possible.

On a separate note: Do we run [Rubocop thread safety](https://docs.rubocop.org/rubocop-thread_safety/index.html) on semian? It could probably benefit from that

kris-gaudel:
> Rubocop thread safety

I think that's a good idea! I'll try it out

Update: This did not catch any of the thread safety issues you mentioned, but it could be worth adding as a surface level check

lib/semian.rb
kris-gaudel: `LRUHash` is a custom class, that uses a mutex for thread safety if `Semain.thread_safe?` is enabled

lib/semian/simple_sliding_window.rb
kris-gaudel: It was used in the simple implementation to provide method delegation (e.g., `size` calls `@window.size`). In the thread safe code it's not used, so I'll remove it.

lib/semian/simple_sliding_window.rb
kris-gaudel: The atomic swap is lock-free so this might yield a very flight performance gain, however the resizing logic (duplicating the window, then resizing it) inside is costly compared to the old implementation.

I decided in the end to get rid of the private resizing method because the new push value doesn't require any of this duplication logic

lib/semian.rb
kris-gaudel: [Atomic reference](https://ruby-concurrency.github.io/concurrent-ruby/master/Concurrent/AtomicReference.html) to the rescue 😄

lib/semian/simple_sliding_window.rb
```diff
@@ -45,26 +46,50 @@ def resize_to(size)
...
 
   module ThreadSafe
-    class SlidingWindow < Simple::SlidingWindow
-      def initialize(**)
-        super
-        @lock = Mutex.new
+    class SlidingWindow
+      attr_reader :max_size
+
+      def initialize(max_size:)
```
AbdulRahmanAlHamali, line 57: We should use the available Array construct

lib/semian.rb
kris-gaudel: I added a mutex to protect the reset method, I think this provides the protection we need, but we should discuss this next week

lib/semian.rb
```diff
@@ -103,13 +104,28 @@ module Semian
...
 
-  attr_accessor :maximum_lru_size, :minimum_lru_time, :default_permissions, :namespace, :default_force_config_validation
+  attr_accessor :maximum_lru_size, :minimum_lru_time, :default_permissions, :namespace, :default_force_config_validation, :resources, :consumers
 
   self.maximum_lru_size = 500
   self.minimum_lru_time = 300 # 300 seconds / 5 minutes
   self.default_permissions = 0660
   self.default_force_config_validation = false
 
+  def thread_safe?
```
AbdulRahmanAlHamali, line 116: can you add a comment here saying:

```
We only allow disabling thread-safety for parts of the code that are on the hot path. Since locking there could have a significant impact. Everything else is enforced thread safety
```

Because otherwise we would have to add thread safety enablement checks for everything in here, and tbh it's not worth it

lib/semian/simple_sliding_window.rb
```diff
@@ -1,69 +1,62 @@
 # frozen_string_literal: true
 
 require "thread"
+require "concurrent"
 
 module Semian
-  module Simple
-    class SlidingWindow # :nodoc:
-      extend Forwardable
+  module SlidingWindowBehavior
```
AbdulRahmanAlHamali, line 7: clean, I like it

test/adapter_test.rb
```diff
@@ -84,7 +84,7 @@ def test_consumer_registration_does_not_prevent_gc
     clients.clear
     2.times { GC.start }
 
-    assert_equal(0, Semian.consumers[identifier].size)
+    assert_operator(Semian.consumers[identifier].size, :<=, 1)
```
AbdulRahmanAlHamali, line 87: AI?

kris-gaudel: No, this was an issue with testing the garbage collector discussed with the Ruby team. I'll forward you my DMs with John on Slack
image

AbdulRahmanAlHamali (approved):

PR #760, opened 2025-09-23
Add adaptive circuit breaker

AbdulRahmanAlHamali:
Add a new adaptive circuit breaker that:

1. Works without need for configuration
2. Has the ability to open partially/fully depending on the severity of the incident

The circuit breaker has two main components:

# Ideal Error Rate Estimator

This estimator tries to find out the expected error rate from a healthy dependency. It does that through a method of simple exponential smoothing, which basically relies on calculating a weighted average, with a few extra domain-specific hints:

1. Ignore any observations that are obviously too high
2. Converge slowly towards higher values, and quickly towards lower values
3. Be more receptive to signals for the first 30 minutes after boot (because we start with a random guess for the value), and less receptive afterwards

# PID Controller (Process, Integral, Derivative) Controller

[See wikipedia link](https://en.wikipedia.org/wiki/Proportional%E2%80%93integral%E2%80%93derivative_controller)

This controller increases/decreases the the rejection rate of the circuit breaker based on the value of:

```
kI * Integral + kP * P + kD * derivative
```

Where P is:

```
ErrorDelta - (1 - ErrorDelta) * RejectionRate
```

The intuition of the value of P:
* Increases when error rate increases
* Decreases when rejection rate increases
* The rejection rate has a (1 - ErrorDelta) multiplier, that we call the "defensiveness" multiplier. This allows rejection to increase more aggressively if the error rate is high

Note: The derivate component is currently set to 0. The Integral component mostly contributes as a history to prevent the circuit breaker from fluctuating too much.

### DIFF
--- a/README.md
+++ b/README.md
@@ -616,21 +616,41 @@ monitors error rates and adjusts its behavior accordingly.
 
 ##### How It Works
 
-The adaptive circuit breaker uses the error function:
+The adaptive circuit breaker has two components:
+
+1. An ideal error rate estimator that determines when the service is starting to become unhealthy
+2. A PID controller that opens the circuit fully or partially based on how bad the situation is.
+
+The ideal error rate estimator uses a "simple exponential smoother", which means it simply takes the average error rate
+that it observes as the ideal. With the following caveat:
+
+1. It ignores any data that is too high from its calculations. For example, we know that 20% error rate is an anamolous
+observation so we ignore it.
+1. It starts with an educated guess about the ideal error rate,
+and then converges down quickly if it observes a lower error rate, and slowly if it observes a higher error rate.
+1. After 30 minutes, it becomes more confident of its guess, and thus converges even slower in either directions.
+
+The PID controller uses the following equation to determine whether to open or close the circuit:
+
 ```
 P = (error_rate - ideal_error_rate) - (1 - (error_rate - ideal_error_rate)) * rejection_rate
 ```
 
-This formula ensures that:
-- Rejection rate increases when the service is unhealthy
-- Rejection rate decreases when the service recovers
-- The system finds an equilibrium that protects against cascading failures while allowing recovery
+Or, more simply, if you define `delta_error = error_rate - ideal_error_rate` then:
+
+```
+P = delta_error - (1 - delta_error) * rejection_rate
+```
+
+In simple terms: This equation says: open more when the error rate is higher than the rejection rate,
+and less when the opposite. The multiplier of `(1 - delta_error)` is called the aggressiveness multiplier.
+It allows the circuit to open more aggressively depending on how bad the situation is.
+
+This P is fed into a typical PID equation, and is used to control the rejection rate of the circuit breaker.
 
 ##### Adaptive Circuit Breaker Configuration
 
-To enable the adaptive circuit breaker, simply set:
-
-- **adaptive_circuit_breaker**. Enable adaptive circuit breaker instead of traditional. Defaults to `false`.
+To enable the adaptive circuit breaker, simply set **adaptive_circuit_breaker** to true.
 
 Example configuration:
 ```ruby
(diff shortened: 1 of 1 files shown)


lib/semian/adaptive_circuit_breaker.rb
```diff
@@ -0,0 +1,136 @@
...
+
+require_relative "pid_controller"
+
+module Semian
+  # Adaptive Circuit Breaker that uses PID controller for dynamic rejection
+  class AdaptiveCircuitBreaker
+    attr_reader :name, :pid_controller, :ping_thread
+
+    def initialize(name:, kp: 1.0, ki: 0.1, kd: 0.0,
+      window_size: 10, history_duration: 3600,
```
kris-gaudel: Regarding the [calculation of online mean](https://docs.google.com/document/d/1gCNuHEDF8rHQLa-URckZ7eo5XZmU64xI-dLYOM_VMfQ/edit?disco=AAABoMer1rY), Welford's algorithm was discussed as a solution to use constant space

kris-gaudel: Also, do we want to allow developers to configure the interval of integration? (`history_duration`)?

.github/workflows/test.yml
```diff
@@ -6,7 +6,7 @@ on:
   push:
     branches: [ main ]
   pull_request:
-    branches: [ main ]
+    branches: [ "**" ]
```
adriangudas: is this something we want to commit to main or were we just testing here?

AbdulRahmanAlHamali: we needed this to make sure all PRs into this branch run CI, but I don't see a problem in keeping it

kris-gaudel:
Hey guys, hope you had a good start to the new year! Are there any issue I could work on in an open-source capacity on the ACB?

lib/semian/adaptive_circuit_breaker.rb
```diff
@@ -0,0 +1,149 @@
...
+
+      @update_thread = if Fiber.scheduler
+        Fiber.schedule(&update_proc)
+      else
+        Thread.new(&update_proc)
+      end
+    end
+
+    def wait_for_window
+      Kernel.sleep(@sliding_interval)
```
samuel-williams-shopify:
```suggestion
      Kernel.sleep(@sliding_interval * rand(0.9...1.1)
```

samuel-williams-shopify:
Here are the notes from our tuple:

```
(A) 1 PIDController
(B) 1 Faraday request

(A) -> sleep(sliding_window_timeout)
(B) -> timeout(t, IO#write -> IO#gets)
(A) -> wakes up, update
(B) -> timeout(t, IO#write -> IO#gets -> response line) -> good

(An) -> sleep(sliding_window_timeout + random_jitter)
(A0) -> wakes up, update -> some CPU time, -> A(N+1) -> time has elapsed
(B) -> timeout(t, IO#write -> IO#gets)
(AN+1) -> wakes up, update # process all of these O(An)
(B) -> timeout(t, IO#write -> IO#gets -> response line) -> times out


Assume sliding window of 1 second
(A0) t+0.1
(A1) t+0.2
(A2) t+0.3
(A0) -> wakes up, update -> some CPU time, -> A(N+1) -> time has elapsed
(A0) t+0.1 -> ready at t+1.0 -> sleep(sliding_window_timeout)
(A1) t+0.2 -> ready at t+1.0 -> sleep(sliding_window_timeout)
(A2) t+0.3 -> ready at t+1.0 -> sleep(sliding_window_timeout)

(1) +random jitter -> always late
(2) +/- random jitter -> gaussian distribution of lateness
(3) + robust loop which computes offset, but you may need to drop sample windows if running late
	(real time stuff)

duration = 1.0
def now
	Process.clock_gettime(Process::CLOCK_MONOTONIC)
end


while true
	current = now
	yield

	next_time = current + duration
	duration = next_time - now
	if duration > 0
		sleep(duration)
	else
		# droppped window
	end
end

while true
	# uniform distribution:
	duration = duration * rand(0.9..1.1)
	sleep(duration * rand(0.9..1.1))

	yield(duration) -> pid_controller.update(duration)
end
```

AbdulRahmanAlHamali:
Note, we are going to ignore CLA: It is complaining about commits made by people who were at Shopify and left

PR #775, opened 2025-10-03
Automate releases with GitHub Actions

### LINKED ISSUE #765: Run releases through Github Actions
Currently, we allow admins to push into `main`. This is because our release process requires that. We should move the releasing into a github action, and then protect the branch from direct pushes

1spyral:
Closes #765 

Creates a GitHub Actions workflow that builds and publishes a release upon a new version tag being added.

To create a release:
Your `git tag` must start with `v` or `test-` (for testing this workflow)

For testing, [this release](https://github.com/Shopify/semian/releases/tag/test-gh-action-release) was generated:

image

## **Post review steps**:
 - [ ] Delete the [test release](https://github.com/Shopify/semian/releases/tag/test-gh-action-release)
 - [ ] Consider removing the `test-` prefix from generating new releases entirely (exists as proof of concept)


.github/workflows/test.yml
```diff
@@ -7,6 +7,7 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+  workflow_call:
```
AbdulRahmanAlHamali, line 10: what is that? 🤔

AbdulRahmanAlHamali (commented):
So in this flow, we still need to first manually update the lib/semian/version.rb, as well as update the changelog, and create a tag. Correct?

I'm wondering if, instead, we can have a workflow that automatically does all of that. It asks you for the version number you want to update to (or maybe it asks you: do you want bugfix, minor or major), and then updates the file, the changelog, the tags, etc.

Worth calling out also: Ideally, we don't want to show most/all dependabot dependency update in the release or in the changelog. Pretty much all semian dependencies use the version of the application

.github/workflows/test.yml
1spyral: This allows for the workflow to be called from other workflows (release workflow)

1spyral:
Closing this PR to rework the behaviour.

New behaviour will ask for major, minor, or bugfix release version, and automate all release-related actions.

Issue #756, opened 2025-09-20
Testing `Simple::*` data structures to show non-thread-safe behaviour


kris-gaudel:
Following up on #755, it would be nice to show that the `Simple::*` data structures indeed are not thread safe by triggering race conditions. An ideal solution would showcase this in the form of unit tests

PR #778, opened 2025-10-04
Thread safety and concurrent-ruby gem migration

### LINKED ISSUE #566: Thread Safety

kris-gaudel:
Fixes #566 

Replace instances of data structures that are not thread safe and migrates existing thread safe code to use the concurrent-ruby gem

Includes unit tests to ensure correctness, thread safety and triggering race conditions will be followed up in #756 

Previous PR had timing issues in the tests, causing flakiness 

Old PR: https://github.com/Shopify/semian/pull/755

### DIFF
--- a/test/thread_safe_integer_test.rb
+++ b/test/thread_safe_integer_test.rb
@@ -1,0 +1,107 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class TestThreadSafeInteger < Minitest::Test
+  def setup
+    @integer = ::Semian::ThreadSafe::Integer.new
+  end
+
+  def teardown
+    @integer.destroy
+  end
+
+  module IntegerTestCases
+    def test_access_value
+      assert_equal(0, @integer.value)
+      @integer.value = 99
+
+      assert_equal(99, @integer.value)
+      time_now = Time.now.to_i
+      @integer.value = time_now
+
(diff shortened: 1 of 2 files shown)


lib/semian.rb
```diff
@@ -103,13 +104,30 @@ module Semian
...
 
-  attr_accessor :maximum_lru_size, :minimum_lru_time, :default_permissions, :namespace, :default_force_config_validation
+  attr_accessor :maximum_lru_size, :minimum_lru_time, :default_permissions, :namespace, :default_force_config_validation, :resources, :consumers
 
   self.maximum_lru_size = 500
   self.minimum_lru_time = 300 # 300 seconds / 5 minutes
   self.default_permissions = 0660
   self.default_force_config_validation = false
 
+  # We only allow disabling thread-safety for parts of the code that are on the hot path.
```
macournoyer, line 114: Is that still true? I only see it called in tests.

macournoyer (approved):
Looks good! But `thread_safe?` looks to be for testing, and comment says otherwise

lib/semian.rb
kris-gaudel: https://github.com/Shopify/semian/pull/755#discussion_r2394427766

PR #779, opened 2025-10-06
Create GitHub Action to publish new release

### LINKED ISSUE #765: Run releases through Github Actions

1spyral:
Closes #765

Creates a GitHub Actions workflow that builds and publishes a release when triggered manually.
The workflow:
- Updates `lib/semian/version.rb`, `Gemfile.lock`, `CHANGELOG.md` and commits them
  - `CHANGELOG.md` contains a list of commits with merge, dependabot, and version bumping filtered out 
- Tags the commit with the next version, determined by workflow input parameter (major, minor, patch)
- Creates and publishes a release, filtering dependabot PRs

[How to trigger a manual workflow](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow)

## Testing:

- `workflow_dispatch` workflows must be triggered from the `main` branch of a repository. Thus, for testing, it is necessary to test on a fork. An example (my fork) can be found at https://github.com/1spyral/semian/

Reviewers: you can leave a comment on this PR if you wish to be added as a Contributor to my fork, where you can test the workflow with minimal setup of your own


.github/workflows/release.yml
```diff
@@ -0,0 +1,167 @@
...
+
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      release_type:
+        description: Type of release
+        required: true
+        default: patch
```
AbdulRahmanAlHamali, line 11: let's remove the default to prompt the user to give this more thought

.github/workflows/release.yml
```diff
@@ -0,0 +1,167 @@
...
+      -
+        name: Read current version
+        id: current_version
+        run: |
+          current_version=$(ruby -e "require './lib/semian/version'; puts Semian::VERSION")
+          echo "version=$current_version" >> $GITHUB_OUTPUT
+          echo "Current version: $current_version"
+      -
+        name: Calculate new version
+        id: new_version
```
AbdulRahmanAlHamali, line 48: At the end of this step, could the action show the version and ask the user to confirm it is correct?

.github/workflows/release.yml
```diff
@@ -0,0 +1,167 @@
...
+          # Update Gemfile.lock with new version
+          bundle install
+      -
+        name: Update CHANGELOG
+        run: |
+          chmod +x .github/scripts/update_changelog.sh
+          .github/scripts/update_changelog.sh \
+            "${{ needs.read-version.outputs.new_version }}" \
+            "${{ needs.read-version.outputs.current_version }}"
+      -
```
AbdulRahmanAlHamali, line 167: Could we display the git diff here for the user to confirm before we commit?

.github/workflows/release.yml
```diff
@@ -0,0 +1,167 @@
...
+        with:
+          # Uses main branch to get the updated version after the commit
+          ref: main
+      -
+        name: Set up Ruby
+        uses: ruby/setup-ruby@v1.263.0
+        with:
+          ruby-version: "3.4"
+          bundler-cache: true
+      -
```
AbdulRahmanAlHamali, line 198: I _think_ these probably need to be done before we create the tag. The tag is what triggers the upload to the rubygems (in our internal tooling), so these steps should _probably_ happen before?

.github/workflows/release.yml
1spyral: I tried that, but it just defaults to the first one (major) instead

.github/workflows/release.yml
1spyral: I'll look into this

PR #795, opened 2025-10-13
Making `consumers` and `resources` private in Semian

kris-gaudel:
As discussed with @brendo, we want to have these fields private to prevent cases like `Semian.resources = "foo"`. We should only have interaction with these fields through public methods

### DIFF
--- a/lib/semian.rb
+++ b/lib/semian.rb
@@ -298,6 +298,8 @@ module Semian
     end
   end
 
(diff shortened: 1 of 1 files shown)


kris-gaudel:
https://shipit.shopify.io/shopify/storefront-renderer/staging-central/tasks/3258722 SFR Verifier passed

PR #797, opened 2025-10-14
Implement P2 Quantile Estimator in PID controller

kris-gaudel:
This PR implements the [P2 quantile estimator](https://dl.acm.org/doi/10.1145/4372.4378). The purpose of this quantile estimator is to help make the calculation of `p90 error_rate` more efficient. This online algorithm requires `O(1)` space and computes each observation in `O(1)` time. This is a substantial improvement from the previous method (storing observations, sorting entire window upon each calculation).

### DIFF
--- a/experiments/experimental_resource.rb
+++ b/experiments/experimental_resource.rb
@@ -28,7 +28,7 @@ module Semian
       # @param timeout [Float, nil] Maximum time to wait for a request (in seconds). If nil, no timeout is enforced.
       # @param error_rate [Float] Baseline error rate (0.0 to 1.0). Probability that any request will fail.
       # @param options [Hash] Additional Semian options
-      def initialize(name:, endpoints_count:, min_latency:, max_latency:, distribution:, timeout: nil, error_rate: 0.0, **options)
+      def initialize(name:, endpoints_count:, min_latency:, max_latency:, distribution:, timeout: nil, error_rate: 0.0, deterministic_errors: false, **options)
         @name = name
         @endpoints_count = endpoints_count
         @min_latency = min_latency
@@ -36,11 +36,16 @@ module Semian
         @distribution = validate_distribution(distribution)
         @timeout = timeout
         @base_error_rate = validate_error_rate(error_rate)
+        @deterministic_errors = deterministic_errors
         @raw_semian_options = options[:semian]
 
         # Initialize service degradation state
         @latency_degradation = { amount: 0.0, target: 0.0, ramp_start: nil, ramp_duration: 0 }
         @error_rate_degradation = { rate: @base_error_rate, target: @base_error_rate, ramp_start: nil, ramp_duration: 0 }
+
+        # Phase-synchronized tracking for precise error rates
+        @current_phase_requests = 0
+        @current_phase_failures = 0
 
         # Assign fixed latencies to each endpoint
         @endpoint_latencies = generate_endpoint_latencies
@@ -130,12 +135,22 @@ module Semian
 
         # If no ramp time, apply immediately
         @error_rate_degradation[:rate] = rate if ramp_time == 0
+
+        # Reset deterministic request counter when error rate changes
+        if @deterministic_errors
+          # Reset phase tracking for new error rate (perfect synchronization)
+          @current_phase_requests = 0
+          @current_phase_failures = 0
+        end
       end
 
       # Reset service to baseline (remove all degradation)
       def reset_degradation
         @latency_degradation = { amount: 0.0, target: 0.0, ramp_start: nil, ramp_duration: 0 }
         @error_rate_degradation = { rate: @base_error_rate, target: @base_error_rate, ramp_start: nil, ramp_duration: 0 }
+        # Reset phase tracking
+        @current_phase_requests = 0
+        @current_phase_failures = 0
       end
 
       # Get current latency degradation (accounting for ramp-up)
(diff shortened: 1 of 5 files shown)


Gemfile.lock
```diff
@@ -171,6 +171,7 @@ GEM
 
 PLATFORMS
   aarch64-linux
+  arm64-darwin-20
```
AbdulRahmanAlHamali: is this change needed?

lib/semian/p2_estimator.rb
```diff
@@ -0,0 +1,129 @@
...
+# At a high level, the algorithm works as follows:
+# Start with 5 marker heights and 5 positions
+# The marker heights are the values of the quantiles we are estimating
+# The positions are the positions of the marker heights in the sorted list of observations
+# As we get new observations, we update the marker heights and positions
+# We use a piecewise parabolic function to estimate the quantile (depending on our condition, we use either a parabolic or linear function)
+
+module Semian
+  # P^2 quantile estimator - used for P90 error-rate estimation
+  class P2QuantileEstimator
```
AbdulRahmanAlHamali, line 24: I'll probably need to read the paper to actually understand what's happening here, but maybe it would be a little bit easier if we also use words for variable names instead of letters?

Gemfile.lock
kris-gaudel: This was an automatically added change for when I ran `bundle exec`. If we don't need it I can remove it

lib/semian/p2_estimator.rb
kris-gaudel: The code is basically [this](https://aakinshin.net/posts/p2-quantile-estimator-rounding-issue/) but translated into Ruby (it's a much easier read and fixes a mistake in the OG paper).
`n` = Actual marker positions
`ns` = Desired marker positions
`q` = Marker heights

lib/semian/p2_estimator.rb
```diff
@@ -0,0 +1,129 @@
...
+    end
+
+    def estimate
+      raise "Sequence contains no elements" if @count == 0
+
+      if @count <= 5
+        sorted_marker_heights = @marker_heights[0...@count].sort
+        index = ((@count - 1) * @quantile).round
+        return sorted_marker_heights[index]
+      end
```
nirmitparikh8, line 52: Why not interpolate? You are rounding the index, but why can't we just interpolate between the indexes? Wouldn't it be more accurate?

lib/semian/p2_estimator.rb
```diff
@@ -0,0 +1,129 @@
...
+        @marker_interval_index = 0
+      elsif value < @marker_heights[1]
+        @marker_interval_index = 1
+      elsif value < @marker_heights[2]
+        @marker_interval_index = 2
+      elsif value < @marker_heights[3]
+        @marker_interval_index = 3
+      else
+        @marker_heights[4] = value
+        @marker_interval_index = 3
```
nirmitparikh8, line 104: Should this not be 4?

lib/semian/p2_estimator.rb
```diff
@@ -0,0 +1,129 @@
...
+      end
+
+      @marker_heights[2]
+    end
+
+    def reset
+      @count = 0
+      @actual_marker_positions = (0...5).to_a
+      @desired_marker_positions = (0...5).map(&:to_f)
+      @marker_heights = (0...5).map(&:to_f)
```
nirmitparikh8, line 61: What about race conditions? Is there ever going to be a time when two functions will be overwriting the same variable at the same time? Can we use maybe a lock or something to prevent this? Just thinking if we call the add observation method and the reset method right after. If there is a latency in the add observation method then reset method would conflict with the add observation method since they both overwrite the same instance variables.

Or is that too nitty gritty? It's just a thought lol

If thats an overkill I definitely see value in removing instance variables for local calculations in the functions. Some instance variables not required are `@marker_interval_index` and`@position_difference`.

lib/semian/p2_estimator.rb
```diff
@@ -0,0 +1,129 @@
...
+        @marker_interval_index = 2
+      elsif value < @marker_heights[3]
+        @marker_interval_index = 3
+      else
+        @marker_heights[4] = value
+        @marker_interval_index = 3
+      end
+
+      ((@marker_interval_index + 1)...5).each { |i| @actual_marker_positions[i] += 1 }
+
```
nirmitparikh8, line 108: Should we not increment count here so we can use the new count value instead of incrementing count at the end on L113?

lib/semian/p2_estimator.rb
kris-gaudel: iirc, Semian instances are isolated per service, so the PID controller is isolated to its respective Semian instance. Since our PID controller only takes in 1 Semian per input, I don't foresee race conditions

cc @AbdulRahmanAlHamali

lib/semian/p2_estimator.rb
kris-gaudel: image

(The paper uses 1-based indexing)

lib/semian/p2_estimator.rb
kris-gaudel: https://www.cse.wustl.edu/~jain/papers/ftp/psqr.pdf
https://aakinshin.net/posts/p2-quantile-estimator-initialization/ 

Line 93 pertains to step 2 in box 1 of the paper (1st link) and this part of the code in this C# implementation (2nd link):

`for (int i = k + 1; i < 5; i++)
            n[i]++;`

lib/semian/p2_estimator.rb
kris-gaudel: Good idea!

Gemfile.lock
AbdulRahmanAlHamali: yeah can you remove it please?

with_p2.out
```diff
@@ -0,0 +1,205 @@
+============================================================
```
AbdulRahmanAlHamali, line 1: Sorry, I had asked you to add these to your PR, but two points:
1. I meant these can probably be put as PR comments or in the PR description
2. Also, I don't fully understand what they're showing

lib/semian/p2_estimator.rb
AbdulRahmanAlHamali: The function that ends up calling this in the PID controller (the store_errors function, which is called from the update function) has thread safety protection

AbdulRahmanAlHamali (approved):

with_p2.out
kris-gaudel: They're showing that the example `examples/adaptive_with_net_http.rb` still works, nothing about performance

AbdulRahmanAlHamali: I still don't think they should be in the PR body, but we can remove them in pid-take-2 later

PR #800, opened 2025-10-15
Set error rates deterministically

nirmitparikh8:
fixes [#6553](https://github.com/Shopify/resiliency/issues/6553)
### Changes
- **Added `deterministic_errors` parameter** (default: `false`)
- **Phase-synchronized error injection** that precisely targets error rates
- **Smart failure logic**: calculates error rate for fail/pass options, chooses whichever gets closer to target
- **Counter tracking**: `@current_phase_requests` and `@current_phase_failures` reset on error rate changes
- **PID error rate history prefilling**: Initialize the entire `error_rate_history` array with the `seed_error_rate value`

### How It Works

- **Deterministic errors**: Instead of `rand < error_rate`, calculates which decision (fail/pass) produces an error rate closest to target. Resets counters between test phases for perfect synchronization.

- **PID prefilling**: Fills error rate history with baseline values so p90 calculations work immediately instead of waiting to accumulate real data.

- **Result**: Reproducible experiments with exact error rate targeting and immediate PID controller responsiveness.

### Future Work
- I will also work on adding a config map to store error rate history from all web workers to prefill new instances with real world data instead of a arbitrary value.

### DIFF
--- a/lib/semian/pid_controller.rb
+++ b/lib/semian/pid_controller.rb
@@ -12,7 +12,7 @@ module Semian
     attr_reader :name, :rejection_rate
 
     def initialize(name:, kp: 1.0, ki: 0.1, kd: 0.0, target_error_rate: nil,
-      window_size: 10, history_duration: 3600)
+      window_size: 10, history_duration: 3600, seed_error_rate: 0.01)
       @name = name
 
       # PID coefficients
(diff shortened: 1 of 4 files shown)


lib/semian/pid_controller.rb
```diff
@@ -159,6 +164,15 @@ def metrics
 
     private
 
+    def prefill_error_rate_history
+      # Fill the history with baseline error rate values lower than expected test values
+      # This allows us to see p90 increase as higher error rates are observed
+      baseline_error_rate = 0.01
```
Aguasvivas22: how about parameterizing this in the initializer so it's more flexible?

something like:

```
def initialize(... seed_error_rate: nil)
...

if seed_error_rate
    windows_to_seed = history_duration / window_size
    @error_rate_history = Array.new(windows_to_seed, seed_error_rate)
end
```

nirmitparikh8: sure I like that better

Aguasvivas22 (approved):
LGTM.

PR #803, opened 2025-10-16
Add `concurrent-ruby` runtime dependency

kris-gaudel:
It was brought to my attention that some repositories using Semian don't have `concurrent-ruby`. This PR adds the gem as a runtime dependency so that these repositories don't have to install it themselves


kkwan3 (approved):

kris-gaudel:
Passed SFR Verifier

PR #808, opened 2025-10-21
dual circuit breaker implementation for switching between adaptive and classic modes

adriangudas:
A basic implementation of a `DualCircuitBreaker` interface that allows us to check a lambda every time we enter `acquire`, and hand off the request to either the classic circuit breaker, or the adaptive circuit breaker, for processing (both are initialized at application start). The lambda has to return `true` (use adaptive) or `false` (use legacy) and provides a pluggable interface to other applications to define the lambda outside of the semian codebase itself, and set it at runtime.


lib/semian/dual_circuit_breaker.rb
```diff
@@ -0,0 +1,163 @@
+# frozen_string_literal: true
+
+module Semian
+  # DualCircuitBreaker wraps both legacy and adaptive circuit breakers,
+  # allowing runtime switching between them via a callable that determines which to use.
+  class DualCircuitBreaker
```
AbdulRahmanAlHamali, line 6: I think this should implement the "CircuitBreakerBehaviour"

lib/semian/dual_circuit_breaker.rb
```diff
@@ -0,0 +1,163 @@
...
+    attr_reader :name, :legacy_circuit_breaker, :adaptive_circuit_breaker
+
+    # use_adaptive should be a callable (Proc/lambda) that returns true/false
+    # to determine which circuit breaker to use. If it returns true, use adaptive.
+    def initialize(name:, legacy_circuit_breaker:, adaptive_circuit_breaker:, use_adaptive:)
+      @name = name
+      @legacy_circuit_breaker = legacy_circuit_breaker
+      @adaptive_circuit_breaker = adaptive_circuit_breaker
+      @use_adaptive = use_adaptive || ->() { false } # Default to legacy
```
AbdulRahmanAlHamali: this defaulting behaviour already exists in `semian.rb`, we should only have it in one location

lib/semian/dual_circuit_breaker.rb
```diff
@@ -0,0 +1,163 @@
...
+    def metrics
+      {
+        active: use_adaptive? ? :adaptive : :legacy,
+        legacy: legacy_metrics,
+        adaptive: adaptive_metrics,
+      }
+    end
+
+    # Get state from the active circuit breaker
+    def state
```
AbdulRahmanAlHamali: do we need the dual circuit breaker to implement this, and the other functions below?

adriangudas:
OK. I think I made some changes based on that pattern that will allow us to avoid using the full adapter pattern. Basically:

1. Create a class method, [here](https://github.com/Shopify/semian/blob/384078794f50937af55cea02d197bd09fa5cc5c3/lib/semian/dual_circuit_breaker.rb#L18-L20), that can be used to set the selector using `Semian::DualCircuitBreaker.adaptive_circuit_breaker_selector(proc)`
2. Write a function that looks up the appropriate value from our dynamic config and returns true or false, and assign it to that proc in the application code

lib/semian/dual_circuit_breaker.rb
adriangudas: no, I think we can remove them?

lib/semian/dual_circuit_breaker.rb
adriangudas: removed it, so there's no default. Instead, I made it so that `def use_adaptive?` will return `false` if the variable is not set. Technically, I think that case still needs to be handled. There could be a short delay between `require semian` executing in app.rb, and the initializer executing in `998_semian.rb`.

adriangudas:
Also wondering if we need to call `reset` to clear the circuit breaker state (i.e. close all circuits) when we switch back and forth 🤔

lib/semian/dual_circuit_breaker.rb
```diff
@@ -0,0 +1,145 @@
...
+      @active_circuit_breaker = nil
+    end
+
+    def DualCircuitBreaker.adaptive_circuit_breaker_selector=(selector)
+      @@adaptive_circuit_breaker_selector = selector
+    end
+
+    # Main acquire method - delegates to the active circuit breaker
+    def acquire(resource = nil, &block)
+      @active_circuit_breaker = get_active_circuit_breaker(resource)
```
adriangudas, line 67: I realized before that we were calling our `use_adaptive?` function (and therefore the proc call to look up our latest configuration) in every method. Instead, we want to only do this on acquire, and then store the active circuit breaker for re-use when we want to call methods on it like open, closed, etc.

Also, this allows us to pass in the `resource` to our proc call so that we can make a decision based on semian resource.

lib/semian/circuit_breaker.rb
```diff
@@ -97,6 +98,16 @@ def in_use?
...
+    def maybe_with_half_open_resource_timeout(resource, &block)
+      if half_open? && @half_open_resource_timeout && resource.respond_to?(:with_resource_timeout)
+        resource.with_resource_timeout(@half_open_resource_timeout) do
+          block.call
+        end
+      else
+        block.call
+      end
+    end
+
```
adriangudas: moved out of `private` so we can access it from `dual_circuit_breaker.rb`

lib/semian/dual_circuit_breaker.rb
```diff
@@ -0,0 +1,188 @@
...
+    attr_reader :legacy_circuit_breaker, :adaptive_circuit_breaker
+
+    # use_adaptive should be a callable (Proc/lambda) that returns true/false
+    # to determine which circuit breaker to use. If it returns true, use adaptive.
+    def initialize(name:, legacy_circuit_breaker:, adaptive_circuit_breaker:)
+      initialize_behaviour(name: name)
+
+      @legacy_circuit_breaker = legacy_circuit_breaker
+      @adaptive_circuit_breaker = adaptive_circuit_breaker
+      @active_circuit_breaker = @adaptive_circuit_breaker
```
AbdulRahmanAlHamali: we should default to classic

lib/semian/dual_circuit_breaker.rb
AbdulRahmanAlHamali: This is not thread-safe/atomic, but might be ok, we should think about it and write a comment

lib/semian/dual_circuit_breaker.rb
```diff
@@ -0,0 +1,188 @@
...
+        handle_legacy_acquire(resource, &block)
+      end
+    end
+
+    def handle_adaptive_acquire(&block)
+      begin
+        result = block.call
+        mark_success
+        result
+      rescue => error
```
AbdulRahmanAlHamali: this needs to be fixed in both adaptive and here. The logic should be the same as the classic

lib/semian/circuit_breaker.rb
```diff
@@ -9,6 +9,7 @@ class CircuitBreaker # :nodoc:
 
     def_delegators :@state, :closed?, :open?, :half_open?
 
+    attr_accessor :exceptions
```
AbdulRahmanAlHamali: exceptions should be exposed from `CircuitBreakerBehaviour` instead

lib/semian/dual_circuit_breaker.rb
```diff
@@ -0,0 +1,188 @@
...
+        mark_failed(error)
```
AbdulRahmanAlHamali: we need an if statement here: both here and in the adaptive file, to match the classic

AbdulRahmanAlHamali: generally, let's have our adaptive code look like the classic one

lib/semian/dual_circuit_breaker.rb
```diff
@@ -0,0 +1,188 @@
...
+    end
+
+    def mark_success
+      @legacy_circuit_breaker&.mark_success
+      @adaptive_circuit_breaker&.mark_success
+    end
+
+    # Stop both circuit breakers
+    def stop
+      @legacy_circuit_breaker&.stop if @legacy_circuit_breaker&.respond_to?(:stop)
```
AbdulRahmanAlHamali: let's just remove this line

lib/semian/dual_circuit_breaker.rb
```diff
@@ -0,0 +1,188 @@
...
+
+    # Destroy both circuit breakers
+    def destroy
+      @legacy_circuit_breaker&.destroy
+      @adaptive_circuit_breaker&.destroy
+    end
+
+    # Check if either circuit breaker is in use
+    def in_use?
+      (@legacy_circuit_breaker&.in_use? || false) ||
```
AbdulRahmanAlHamali: let's simplify, this will never be nil, so we can remove the || false

lib/semian/dual_circuit_breaker.rb
```diff
@@ -0,0 +1,188 @@
...
+
+    def self.adaptive_circuit_breaker_selector(selector)
+      @@adaptive_circuit_breaker_selector = selector
+    end
+
+    # Main acquire method
+    # Logic from both acquire methods is implemented here so that we can fan out mark_success and mark_failed
+    # to both circuit breakers.
+    def acquire(resource = nil, &block)
+      @active_circuit_breaker = get_active_circuit_breaker(resource)
```
AbdulRahmanAlHamali, line 67:
```suggestion
      old_type = active_circuit_breaker_type
      @active_circuit_breaker = get_active_circuit_breaker(resource)
      if old_type != active_circuit_breaker_type
        Semian.Notify(blabla)
      end
```

lib/semian/dual_circuit_breaker.rb
```diff
@@ -0,0 +1,188 @@
...
+    # Get metrics from both circuit breakers for comparison
+    def metrics
+      {
+        active: active_breaker_type,
+        legacy: legacy_metrics,
+        adaptive: adaptive_metrics,
+      }
+    end
+
+    def active_breaker_type
```
AbdulRahmanAlHamali: should probably be `private`

lib/semian/dual_circuit_breaker.rb
```diff
@@ -0,0 +1,188 @@
...
+        (@adaptive_circuit_breaker&.in_use? || false)
+    end
+
+    # Get the last error from the active circuit breaker
+    def last_error
+      @active_circuit_breaker.last_error
+    end
+
+    # Get metrics from both circuit breakers for comparison
+    def metrics
```
AbdulRahmanAlHamali, line 121: we'll remove AI fluff here

test/dual_circuit_breaker_test.rb
```diff
@@ -0,0 +1,242 @@
...
+
+  def teardown
+    Semian.destroy_all_resources
+  end
+
+  def test_creates_both_circuit_breakers
+    resource = create_dual_resource
+
+    assert_instance_of(Semian::DualCircuitBreaker, resource.circuit_breaker)
+    assert(resource.circuit_breaker.legacy_circuit_breaker)
```
AbdulRahmanAlHamali: assert_instance_of here as well

test/dual_circuit_breaker_test.rb
```diff
@@ -0,0 +1,242 @@
...
+    resource = create_dual_resource
+    @use_adaptive_flag = false
+
+    # Legacy circuit breaker should handle the request
+    success_count = 0
+    3.times do
+      resource.acquire { success_count += 1 }
+    end
+
+    assert_equal(3, success_count)
```
AbdulRahmanAlHamali: shouldn't be implicitly testing unrelated stuff in the test

test/dual_circuit_breaker_test.rb
```diff
@@ -0,0 +1,242 @@
...
+    resource = create_dual_resource
+    @use_adaptive_flag = true
+
+    # Adaptive circuit breaker should handle the request
+    success_count = 0
+    3.times do
+      resource.acquire { success_count += 1 }
+    end
+
+    assert_equal(3, success_count)
```
AbdulRahmanAlHamali: same comment as above

test/dual_circuit_breaker_test.rb
```diff
@@ -0,0 +1,242 @@
...
+  end
+
+  def test_destroy_destroys_both_breakers
+    resource = create_dual_resource
+
+    # Destroy
+    resource.circuit_breaker.destroy
+
+    # Verify cleanup (threads should be stopped for adaptive)
+    assert(resource.circuit_breaker.adaptive_circuit_breaker.instance_variable_get(:@stopped))
```
AbdulRahmanAlHamali, line 97: instead use `expects` and do for both

test/dual_circuit_breaker_test.rb
```diff
@@ -0,0 +1,242 @@
...
+  end
+
+  def test_metrics_includes_both_breakers
```
AbdulRahmanAlHamali: AI fluff

test/dual_circuit_breaker_test.rb
```diff
@@ -0,0 +1,242 @@
...
+      error_timeout: 5,
+      tickets: 2,
+    )
+
+    assert_nil(resource.circuit_breaker)
+  ensure
+    ENV.delete("SEMIAN_CIRCUIT_BREAKER_DISABLED")
+  end
+
+  def test_returns_nil_if_either_breaker_cannot_be_created
```
AbdulRahmanAlHamali: we should remove this test

test/dual_circuit_breaker_test.rb
```diff
@@ -0,0 +1,242 @@
...
+  def test_both_breakers_track_last_error
+    resource = create_dual_resource
+
+    begin
+      resource.acquire { raise TestError, "test error" }
+    rescue TestError
+      # Expected
+    end
+
+    assert(resource.circuit_breaker.last_error)
```
AbdulRahmanAlHamali: should this one also be `assert_equal`?

lib/semian/dual_circuit_breaker.rb
```diff
@@ -0,0 +1,188 @@
...
+  # DualCircuitBreaker wraps both legacy and adaptive circuit breakers,
+  # allowing runtime switching between them via a callable that determines which to use.
+  class DualCircuitBreaker
+    include CircuitBreakerBehaviour
+
+    attr_reader :legacy_circuit_breaker, :adaptive_circuit_breaker
+
+    # use_adaptive should be a callable (Proc/lambda) that returns true/false
+    # to determine which circuit breaker to use. If it returns true, use adaptive.
+    def initialize(name:, legacy_circuit_breaker:, adaptive_circuit_breaker:)
```
anagayan: FWIW, this is mainly a comment about naming and my two 🪙 . 

Referring to the existing circuit breaker as `legacy` implies it will be replaced by the other. The bigger point is that it's not obvious what it's referring to (take the example that we somehow come up with multiple versions of circuit breaker). Therefore, I am proposing to name it according to the class defined, so it would be `circuit_breaker`.

Example:
```suggestion
    attr_reader :circuit_breaker, :adaptive_circuit_breaker

    # use_adaptive should be a callable (Proc/lambda) that returns true/false
    # to determine which circuit breaker to use. If it returns true, use adaptive.
    def initialize(name:, circuit_breaker:, adaptive_circuit_breaker:)
```

test/dual_circuit_breaker_test.rb
```diff
@@ -0,0 +1,222 @@
...
+  end
+
+  def test_can_switch_between_breakers_at_runtime
+    resource = create_dual_resource
+
+    # Start with legacy
+    @use_adaptive_flag = false
+    resource.acquire { "success" }
+
+    assert_instance_of(Semian::CircuitBreaker, resource.circuit_breaker.legacy_circuit_breaker)
```
adriangudas: I need to fix this :(
expects is not working for some reason.

lib/semian/circuit_breaker.rb
```diff
@@ -97,6 +97,16 @@ def in_use?
       !error_timeout_expired? && !@errors.empty?
     end
 
+    def maybe_with_half_open_resource_timeout(resource, &block)
```
AbdulRahmanAlHamali: need to make this private again

PR #809, opened 2025-10-21
Fix cold start issue

nirmitparikh8:
Fixes [#6545](https://github.com/Shopify/resiliency/issues/6545)

- Add initial_history_duration and initial_error_rate param in the PID controller.
- Added a prefill p90 estimator function that is called during the PID controllers initialization. This function just adds initial_error_rate as an observation n times. n = initial_history_duration/window_size

### DIFF
--- a/lib/semian/pid_controller.rb
+++ b/lib/semian/pid_controller.rb
@@ -13,7 +13,7 @@ module Semian
     attr_reader :name, :rejection_rate
 
     def initialize(name:, kp: 1.0, ki: 0.1, kd: 0.0, target_error_rate: nil,
-      window_size: 10, initial_history_duration: 3600, initial_error_rate: 0.01)
+      window_size: 10, initial_history_duration: 900, initial_error_rate: 0.01)
       @name = name
 
       # PID coefficients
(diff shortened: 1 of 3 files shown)


lib/semian/pid_controller.rb
```diff
@@ -140,6 +147,9 @@ def reset
       @last_error_rate = 0.0
       @last_ping_failure_rate = 0.0
       @p90_estimator.reset
+
+      # Refill P90 estimator after reset
+      prefill_p90_estimator
```
nirmitparikh8, line 152: After a reset should we prefill? I would argue we shouldn't but want to get your thoughts?

kris-gaudel: Hmmm what scenario would we be calling reset? If we really do want to clear and "reset" the observations I agree we shouldn't prefill. 

If we're treating the reset as more of a "restart", then we should prefill

kris-gaudel: For now let's just leave it I think

kris-gaudel (approved):

PR #812, opened 2025-10-22
Inherit from parent `CircuitBreaker` class

kris-gaudel:
`AdaptiveCircuitBreaker` now inherits from the `CircuitBreaker` class

### DIFF
--- a/lib/semian/adaptive_circuit_breaker.rb
+++ b/lib/semian/adaptive_circuit_breaker.rb
@@ -12,8 +12,7 @@ module Semian
 
     def initialize(name:, kp: 1.0, ki: 0.1, kd: 0.01,
       window_size: 10, initial_history_duration: 900, initial_error_rate: 0.01,
-      thread_safe: true, exceptions: [StandardError], implementation: Semian,
-      error_timeout: 10, error_threshold: 3, success_threshold: 1)
+      thread_safe: true)
       initialize_behaviour(name: name)
 
       @window_size = window_size


lib/semian/adaptive_circuit_breaker.rb
```diff
@@ -147,21 +155,5 @@ def start_update_thread
         Semian.logger&.warn("[#{@name}] Background update thread error: #{e.message}")
       end
     end
-
-    def start_update_thread
```
AbdulRahmanAlHamali, line 151: ummm?

kris-gaudel: This was a duplicated method

AbdulRahmanAlHamali: gotcha

adriangudas:
What I think we need most is a common interface to both, and have the adaptive and regular circuit breakers share common methods so the interface applies to both simultaneously. Then the interface can be used to e.g. check the feature flag and gate access to one or the other, for example - this is what `DualCircuitBreaker` was supposed to do in [this PR](https://github.com/Shopify/semian/pull/808).

kris-gaudel:
For now I commented out the broken tests related to unit testing the ACB and PID controller, implementing the behaviour does not cause any old tests (related to the normal CB) to fail

image

image

lib/semian/adaptive_circuit_breaker.rb
```diff
@@ -1,20 +1,24 @@
...
+    include CircuitBreakerBehaviour
+
+    attr_reader :pid_controller, :update_thread
 
     def initialize(name:, kp: 1.0, ki: 0.1, kd: 0.01,
       window_size: 10, initial_history_duration: 900, initial_error_rate: 0.01,
-      thread_safe: true)
-      @name = name
+      thread_safe: true, exceptions: [StandardError], implementation: Semian,
+      error_timeout: 10, error_threshold: 3, success_threshold: 1)
```
AbdulRahmanAlHamali: I guess we don't need these params anymore

PR #824, opened 2025-10-24
add experiments and results for sudden error spikes

adriangudas:
Add experiments and tests for adaptive and classic circuit breakers, for a total of 6 (3 each):

- Two spikes (20%, then 60%)
- One spike of 20%
- One spike of 100%

As expected, the classic circuit breaker is opening constantly since the error rate detection is deterministic on a very low count of errors.


experiments/test_sudden_error_spikes_adaptive.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+$LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
+
+require "semian"
+require_relative "mock_service"
+require_relative "experimental_resource"
+require_relative "test_helpers"
+
+# Error spikes test: 0.1% -> 20% -> 0.1% -> 60% -> 0.1% with 20 second spikes
```
Aguasvivas22: .1% -> 1% (same for graph title below)

experiments/test_sudden_error_spikes_adaptive.rb
```diff
@@ -0,0 +1,25 @@
...
+runner = Semian::Experiments::CircuitBreakerTestRunner.new(
+  test_name: "Sudden Error Spikes Test - adaptive",
+  resource_name: "protected_service_sudden_error_spikes_adaptive",
+  error_phases: [0.01] * 3 + [0.20] + [0.01] * 3 + [0.60] + [0.01] * 3,
```
Aguasvivas22: this test seems to be encompassing the 20% scenario that's in test_error_spike_20_adaptive.rb also, any reason not to include the 100% one here as well, in which case it'd be 1 test for all?

adriangudas: I was mistakenly not being consistent. I'll add the 100% scenario.

PR #828, opened 2025-10-24
fix adaptive circuit breaker tests

AbdulRahmanAlHamali:
Another part of https://github.com/Shopify/resiliency/issues/6576. In this PR:

* Some clean up for adaptive circuit breaker, removing fluff comments, etc.
* Full rewrite of its tests

### DIFF
--- a/test/adaptive_circuit_breaker_test.rb
+++ b/test/adaptive_circuit_breaker_test.rb
@@ -16,9 +16,7 @@ class TestAdaptiveCircuitBreaker < Minitest::Test
     end
 
     def sleep(duration)
-      until @should_start
-        Kernel.sleep(0.01)
-      end
+      Kernel.sleep(0.01) until @should_start
 
       # Only count window_size sleeps, this helps us detect if the wrong value is being passed
       if duration == @window_size
@@ -53,7 +51,7 @@ class TestAdaptiveCircuitBreaker < Minitest::Test
     # Execute the block and verify it returns the expected value
     result = @breaker.acquire { "successful_result" }
 
-    assert_equal "successful_result", result
+    assert_equal("successful_result", result)
   end
 
   def test_acquire_with_error_raises_and_records_request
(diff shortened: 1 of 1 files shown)


lib/semian/adaptive_circuit_breaker.rb
```diff
@@ -4,20 +4,26 @@
 require_relative "circuit_breaker_behaviour"
 
 module Semian
+  # Default clock implementation using real time
```
kris-gaudel, line 7: What is the goal/purpose for being able to pass in a clock? Is it only for testing?

AbdulRahmanAlHamali: yupp

AbdulRahmanAlHamali: I'm not sure if there are smarter ways in Ruby that allow us to get around dependency injection.

In general, we could, you could outright mock anything out. However, in this case I needed a proper class in the test that would hold variables that I could use for synchronizing the test, so I resorted to a classical dependency injection.

If you know of better ways I'm up to change this

kris-gaudel: Hmmm tbh I don't have any better ideas, but I think it's fine. Creating a separate class just for testing seems counter-productive to showing the correctness of the ACB

kris-gaudel:
`bundle exec rubocop -A` failed on my local branch but it auto-corrected the failing lines

AbdulRahmanAlHamali:
thanks!

lib/semian/adaptive_circuit_breaker.rb
AbdulRahmanAlHamali: It's standard practice in a lot of languages: See [dependency injection](https://en.wikipedia.org/wiki/Dependency_injection).

You often don't see it in Ruby because it allows you to mock more flexibly, but sometimes you still need it

PR #829, opened 2025-10-27
Add thread timing utilization metrics for experiments

adriangudas:
Add a summary which captures the amount of time spent making requests per thread, which looks like this:

```
=== Thread Timing Statistics ===
Total threads: 60
Test wall clock duration: 180.00s

Time spent making requests per thread:
  Min:     45.23s
  Max:     52.17s
  Average: 48.50s
  Total (all threads): 2910.00s

Thread utilization:
  Average: 26.94% (time in requests / wall clock time)

Requests per thread:
  Average: 8000 requests
  Average time per request: 0.0061s
```

Additionally, a second graph is created to visualize the total time spent making requests, with `duration-` as a filename prefix (the rest of the filename is re-used). Duration graph:
 
duration-sudden_error_spike_20_adaptive

### DIFF
--- a/experiments/test_helpers.rb
+++ b/experiments/test_helpers.rb
@@ -163,7 +163,6 @@ module Semian
               if semian_resource&.circuit_breaker
                 metrics = semian_resource.circuit_breaker.pid_controller.metrics
 
-                # Calculate total time spent making requests across all threads
                 total_request_time = @thread_timings.values.sum { |t| t[:samples].sum { |s| s[:duration] } }
 
                 @pid_mutex.synchronize do
@@ -347,7 +346,6 @@ module Semian
           bucket_end = bucket_start + small_bucket_size
           bucket_data = @outcomes.select { |time, _| time >= bucket_start && time < bucket_end }
 
-          # Calculate sum of request durations from all samples in this bucket
           bucket_samples = []
           @thread_timings.each_value do |thread_data|
             bucket_samples.concat(thread_data[:samples].select { |s| s[:timestamp] >= bucket_start && s[:timestamp] < bucket_end })


experiments/test_helpers.rb
```diff
@@ -95,18 +97,32 @@ def start_request_threads
...
+
             sleep_interval = 1.0 / @requests_per_second_per_thread
             until @done
               sleep(sleep_interval)
 
+              # Measure time spent making the request
+              request_start = Time.now
               begin
                 thread_resource.request(rand(@service.endpoints_count))
+                request_duration = Time.now - request_start
```
AbdulRahmanAlHamali: you can move this (and the two repetitions below), to an `ensure` block

experiments/test_helpers.rb
```diff
@@ -95,18 +97,32 @@ def start_request_threads
...
         @num_threads.times do
           @request_threads << Thread.new do
+            thread_id = Thread.current.object_id
             thread_resource = ExperimentalResource.new(
               name: @resource_name,
               service: @service,
               semian: @semian_config,
             )
 
+            # Initialize timing for this thread
```
AbdulRahmanAlHamali: I don't think you need mutex synchronization for this at all. Since each key is unique per thread, it does not face the risk of parallel access

adriangudas:
> Looks good overall. Added a couple of comments, but I also think you should add time measurements to monitoring threads, so that we can draw a graph

Added. Just thinking about what the graph should look like.

AbdulRahmanAlHamali:
> Just thinking about what the graph should look like.

Let's display a time series graph that shows both request count and total processing time.

An experiment where we inject timeout-level latency on one endpoint would look very interesting

PR #839, opened 2025-10-29
Fix "error P" reporting in summary output

adriangudas:
Previously, "Error P" and "PrevError" were showing 0 in the summary report.

This PR fixes the logic so that:
- The previous P value is properly captured (copied from "current P value" at the beginning of the update function, not at the end).
- The proper metrics variables are now referenced in `pid_snapshots`.

Now the "Error P" and "PrevError" columns have real values:

```
Window   Current Err %   Ideal Err %     Error P      Reject %        Integral     PrevError    Derivative   Total Req Time
------------------------------------------------------------------------------------------------------------------------
1        0.0%            1.0%            0.0          0.0%            0.0          0.0          0.0          48.08s
2        0.97%           1.0%            -0.0003      0.0%            -0.0029      0.0          -0.0         563.09s
3        0.68%           1.0%            -0.0032      0.0%            -0.0349      -0.0003      -0.0003      1079.81s
4        0.77%           1.0%            -0.0023      0.0%            -0.0582      -0.0032      0.0001       1594.9s
5        0.99%           1.0%            -0.0001      0.0%            -0.059       -0.0023      0.0002       2112.02s
6        0.76%           1.0%            -0.0024      0.0%            -0.0832      -0.0001      -0.0002      2628.08s
7        0.82%           1.0%            -0.0018      0.0%            -0.101       -0.0024      0.0001       3143.89s
8        18.5%           1.0%            0.175        7.51%           1.6485       -0.0018      0.0177       3651.27s
9        19.31%          1.27%           0.1053       10.78%          2.7017       0.175        -0.007       4153.14s
10       1.43%           1.68%           -0.1103      5.8%            1.5985       0.1053       -0.0216      4661.18s
11       1.14%           1.68%           -0.0633      4.66%           0.9651       -0.1103      0.0047       5173.26s
12       0.77%           1.68%           -0.0557      3.19%           0.4085       -0.0633      0.0008       5686.28s
13       0.88%           1.68%           -0.0398      2.15%           0.0104       -0.0557      0.0016       6200.82s
```

### DIFF
--- a/test/pid_controller_test.rb
+++ b/test/pid_controller_test.rb
@@ -32,6 +32,7 @@ class TestPIDController < Minitest::Test
         error_rate: 0.0,
         ideal_error_rate: 0.01, # Default when no history
         p_value: 0.0,
+        previous_p_value: 0.0,
         integral: 0.0,
         derivative: 0.0,
         current_window_requests: { success: 0, error: 0, rejected: 0 },
@@ -184,6 +185,7 @@ class TestPIDController < Minitest::Test
         error_rate: 0.0,
         ideal_error_rate: 0.01, # Default when no history
         p_value: 0.0,
+        previous_p_value: 0.0,
         integral: 0.0,
         derivative: 0.0,
         current_window_requests: { success: 0, error: 0, rejected: 0 },


experiments/test_helpers.rb
```diff
@@ -171,11 +171,11 @@ def start_monitoring_thread
                     window: @pid_snapshots.length + 1,
                     current_error_rate: metrics[:error_rate],
                     ideal_error_rate: metrics[:ideal_error_rate],
-                    error_metric: metrics[:error_metric],
+                    p_value: metrics[:p_value],
```
Aguasvivas22, line 174: the name for this variable was changed previously since one might confuse this for the value of the P term. the change might have gotten lost in a refactor but I think we want to keep it as error_metric or something related to errors as it's for e(t) = (current - ideal) - rejections

adriangudas: OK. Makes sense. I'll keep it as `error_metric`. If anyone has a good suggestion on another name, we can use that.

Aguasvivas22: actually, now that I think about it, I'm ok with leaving it as p_value for simplicity.
e(t) = (current - ideal) - rejections but P is proportional to that so P is essentially P=e(t).

adriangudas: Sure. That sounds good to me.

adriangudas: Done

lib/semian/pid_controller.rb
```diff
@@ -120,7 +120,6 @@ def metrics
         p_value: @last_p_value,
         integral: @integral,
         derivative: @derivative,
-        previous_p_value: @previous_p_value,
```
adriangudas, line 123: This isn't really needed, since you can just look at the line above it if you want the previous P value. It was broken, so instead of fixing it, I just removed it

AbdulRahmanAlHamali (approved):
Chatted offline: we do want to report `previous_p_value`, and Adrian was right to move it up the code so that its value is not overwritten.

Approved

PR #840, opened 2025-10-29
Add an experiment with multiple services and latency degradation of one of them

AbdulRahmanAlHamali:
This PR:

1. Allows the ability to create several services instead of just one.
2. Allows the ability to provide error or latency degradation
3. Introduces a test where we have 10 services, and one of them experiences a latency degradation

Fixes https://github.com/Shopify/resiliency/issues/6599

### DIFF
--- a/experiments/test_helpers.rb
+++ b/experiments/test_helpers.rb
@@ -125,6 +125,9 @@ module Semian
             until @done
               sleep(sleep_interval)
               service = @services.sample
+              # technically, we are creating a new resource instance on every request.
+              # But the resource class is pretty much only a wrapper around things that are longer-living.
+              # So this works just fine.
               thread_resource = ExperimentalResource.new(
                 name: "#{@resource_name}_#{service.object_id}",
                 service: service,


experiments/test_helpers.rb
```diff
@@ -97,21 +119,21 @@ def start_request_threads
...
-            )
-
             @thread_timings[thread_id] = { samples: [] }
 
             sleep_interval = 1.0 / @requests_per_second_per_thread
             until @done
               sleep(sleep_interval)
+              service = @services.sample
+              thread_resource = ExperimentalResource.new(
+                name: "#{@resource_name}_#{service.object_id}",
```
AbdulRahmanAlHamali, line 132: make sure that each service has its own resource. Later one, we should have each thread have its own resource as well

lib/semian.rb
```diff
@@ -316,9 +316,9 @@ def create_adaptive_circuit_breaker(name, **options)
     # Fixed parameters based on design document recommendations
     AdaptiveCircuitBreaker.new(
       name: name,
-      kp: 0.3, # Standard proportional gain
-      ki: 0.003, # Moderate integral gain
-      kd: 1.0, # Small derivative gain (as per design doc)
+      kp: 0.75, # Standard proportional gain
```
AbdulRahmanAlHamali, line 319: @Aguasvivas22 I put the new K values. Are you ok with me merging them as part of this?

experiments/test_helpers.rb
Aguasvivas22: if I'm reading this right, we're now creating a new resource for every request? Is that the intent?

AbdulRahmanAlHamali: hmmmm I think you're right. This is wrong. I wanted to retrieve the existing resources, but I think I messed up. Let me fix that

AbdulRahmanAlHamali: ok, I looked into `ExperimentalResource`, and it's not really holding that much state by itself. It only contains references to other things. So I could basically introduce a "retrieve" function and call that instead, but tbh it doesn't seem worth it.

I added an in-code comment about that as well

PR #845, opened 2025-10-30
Remove dependency injection

kris-gaudel:
Remove dependency injection so that we don't add additional parameters

### DIFF
--- a/test/adaptive_circuit_breaker_test.rb
+++ b/test/adaptive_circuit_breaker_test.rb
@@ -10,16 +10,11 @@ class TestAdaptiveCircuitBreaker < Minitest::Test
       kp: 1.0,
       ki: 0.1,
       kd: 0.01,
-      window_size: 10,
+      window_size: 0.05,
       initial_history_duration: 100,
       initial_error_rate: 0.01,
       thread_safe: true,
-      clock: clock,
     )
-  end
-
-  def setup
-    @breaker = create_test_breaker(name: "test_breaker")
   end
 
   def teardown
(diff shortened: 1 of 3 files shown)


test/adaptive_circuit_breaker_test.rb
```diff
@@ -88,37 +66,34 @@ def test_acquire_with_rejection_raises_and_records_request
...
-
-    done = false
-    mock_clock = MockClock.new(max_sleeps: 3) do |_|
-      done = true
-      # NOTE: breaker.stop kills the thread. Any line after it will not be executed.
-      breaker.stop
-    end
-
+    sleep_count = 0
     breaker = Semian::AdaptiveCircuitBreaker.new(
```
AbdulRahmanAlHamali: we could just use the one from the setup, since this one is no different anymore

test/adaptive_circuit_breaker_test.rb
```diff
@@ -88,37 +66,34 @@ def test_acquire_with_rejection_raises_and_records_request
...
-
-    Kernel.sleep(0.01) until done
+    # Stub wait_for_window to avoid actual sleeping and control when to stop
+    breaker.stub(:wait_for_window, lambda {
+      sleep_count += 1
+      breaker.stop if sleep_count >= 3
+      Kernel.sleep(0.01) # Small sleep to prevent tight loop
+    }) do
+      # We call update after waiting. Since we stop on the third wait, we only expect 2 updates.
+      breaker.pid_controller.expects(:update).times(2)
```
AbdulRahmanAlHamali: I'm hesitant about this. You are currently hinging on the assumption that the block will actually run before those 3 sleeps. But you have no guarantee about this.

Even if one of them runs before the block, your block is going to fail. You can simulate this by adding a sleep of a few seconds right before line 92, and you'll see your test failing.

You could prevent this by preventing the sleep from progressing by adding a loop in the beginning that waits for you to do your `expects` before it actually starts. See how I was using `should_start`

AbdulRahmanAlHamali (approved):

PR #846, opened 2025-10-30
Different Semians between different threads

kris-gaudel:
Also added a line to the `Gemfile` as I was running into errors when trying to run experiments

Edit: Did not end up increasing error threshold

### DIFF
--- a/experiments/test_helpers.rb
+++ b/experiments/test_helpers.rb
@@ -110,7 +110,6 @@ module Semian
         @outcomes = {}
         @done = false
         @outcomes_mutex = Mutex.new
-        @pid_snapshots = []
         @pid_mutex = Mutex.new
         @thread_timings = {}
         @state_transitions = []
@@ -190,67 +189,27 @@ module Semian
 
       def subscribe_to_metrics
         target_resource_prefix = "#{@resource_name}_#{@target_service.object_id}_thread_"
-        # Track metrics per window to aggregate across threads
-        @window_metrics = {}
+        # Collect raw metrics from all threads
+        @raw_metrics = []
 
         @subscription = Semian.subscribe do |event, resource, _scope, _adapter, payload|
           # Only capture adaptive_update events for our target service thread resources
           next unless event == :adaptive_update && resource.name.to_s.start_with?(target_resource_prefix)
 
-          total_request_time = @thread_timings.values.sum { |t| t[:samples].sum { |s| s[:duration] } }
-
           @pid_mutex.synchronize do
             timestamp = Time.now.to_i
-            window_key = timestamp
 
-            # Initialize window metrics if this is the first thread reporting for this window
-            @window_metrics[window_key] ||= {
+            @raw_metrics << {
               timestamp: timestamp,
-              window: nil,
-              error_rates: [],
+              resource_name: resource.name,
+              error_rate: payload[:error_rate],
               ideal_error_rate: payload[:ideal_error_rate],
-              p_values: [],
-              previous_p_values: [],
-              rejection_rates: [],
-              integrals: [],
-              derivatives: [],
-              total_request_time: total_request_time,
+              p_value: payload[:p_value],
+              previous_p_value: payload[:previous_p_value],
+              rejection_rate: payload[:rejection_rate],
+              integral: payload[:integral],
+              derivative: payload[:derivative],
             }
-
-            # Accumulate metrics from each thread
-            @window_metrics[window_key][:error_rates] << payload[:error_rate]
-            @window_metrics[window_key][:p_values] << payload[:p_value]
-            @window_metrics[window_key][:previous_p_values] << payload[:previous_p_value]
-            @window_metrics[window_key][:rejection_rates] << payload[:rejection_rate]
-            @window_metrics[window_key][:integrals] << payload[:integral]
-            @window_metrics[window_key][:derivatives] << payload[:derivative]
-
-            # When we've collected metrics from all threads, create the aggregated snapshot
-            if @window_metrics[window_key][:error_rates].length == @num_threads
-              metrics = @window_metrics[window_key]
-
-              @pid_snapshots << {
-                timestamp: metrics[:timestamp],
-                window: @pid_snapshots.length + 1,
-                error_rate_avg: metrics[:error_rates].sum / metrics[:error_rates].length.to_f,
-                error_rate_min: metrics[:error_rates].min,
-                error_rate_max: metrics[:error_rates].max,
-                ideal_error_rate: metrics[:ideal_error_rate],
-                rejection_rate_avg: metrics[:rejection_rates].sum / metrics[:rejection_rates].length.to_f,
-                rejection_rate_min: metrics[:rejection_rates].min,
-                rejection_rate_max: metrics[:rejection_rates].max,
-                integral_avg: metrics[:integrals].sum / metrics[:integrals].length.to_f,
-                integral_min: metrics[:integrals].min,
-                integral_max: metrics[:integrals].max,
-                derivative_avg: metrics[:derivatives].sum / metrics[:derivatives].length.to_f,
-                derivative_min: metrics[:derivatives].min,
-                derivative_max: metrics[:derivatives].max,
-                total_request_time: metrics[:total_request_time],
-              }
-
-              # Clean up processed window
-              @window_metrics.delete(window_key)
-            end
           end
         end
       end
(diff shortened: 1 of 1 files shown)


experiments/Gemfile
```diff
@@ -2,6 +2,9 @@
 
 source "https://rubygems.org"
 
+# Load the local semian gem
```
AbdulRahmanAlHamali: good one. This would install all semian dependencies in here

experiments/test_helpers.rb
```diff
@@ -94,11 +94,11 @@ def start_request_threads
         puts "Each thread will have its own adapter instance connected to the shared service...\n"
 
         @request_threads = []
-        @num_threads.times do
+        @num_threads.times do |thread_num|
           @request_threads << Thread.new do
             thread_id = Thread.current.object_id
             thread_resource = ExperimentalResource.new(
-              name: @resource_name,
+              name: "#{@resource_name}_thread_#{thread_num}",
```
AbdulRahmanAlHamali: you could also use thread id

AbdulRahmanAlHamali: but I guess that would make it hard to reference this outside of the thread. So nvm

experiments/test_helpers.rb
```diff
@@ -159,29 +159,43 @@ def start_monitoring_thread
...
                     timestamp: Time.now.to_i,
                     window: @pid_snapshots.length + 1,
-                    current_error_rate: metrics[:error_rate],
-                    ideal_error_rate: metrics[:ideal_error_rate],
-                    error_metric: metrics[:error_metric],
-                    rejection_rate: metrics[:rejection_rate],
-                    integral: metrics[:integral],
-                    derivative: metrics[:derivative],
-                    previous_error: metrics[:previous_error],
+                    ideal_error_rate: all_metrics.first[:ideal_error_rate], # Same for all threads
```
AbdulRahmanAlHamali: how about you still give it the same treatment as the others though? To simplify the code

experiments/test_helpers.rb
```diff
@@ -214,6 +228,12 @@ def wait_for_completion
         @request_threads.each(&:join)
         @monitor_thread.join if @is_adaptive
         @end_time = Time.now
+
+        # Clean up per-thread resources
```
AbdulRahmanAlHamali, line 274: not saying it's bad, but I'm interested to know whether you saw something that made you want to add a cleanup

kris-gaudel: Preventative measure I guess, to ensure system resources (SysV semaphores) are cleaned up

kris-gaudel:
> What's the reason you chose to increase the error_threshold?

From the [issue](https://github.com/orgs/Shopify/projects/13332/views/1?pane=issue&itemId=135288169&issue=Shopify%7Cresiliency%7C6594): "Increase error thresholds for the classic circuit breaker, so that it's not constantly open"

> Also, you seem to have committed some PNGs, is that intentional?

Wanted to show some updated results after running the experiments, can remove if we don't need

experiments/test_helpers.rb
```diff
@@ -116,9 +116,15 @@ def start_request_threads
         puts "Each thread will have its own adapter instance connected to the shared service...\n"
 
         @request_threads = []
-        @num_threads.times do
+        @num_threads.times do |thread_num|
           @request_threads << Thread.new do
             thread_id = Thread.current.object_id
+            thread_resource = ExperimentalResource.new(
```
AbdulRahmanAlHamali: I think this needs to be removed

AbdulRahmanAlHamali: and you should modify the one a few lines below instead

experiments/test_helpers.rb
```diff
@@ -184,30 +190,42 @@ def start_monitoring_thread
...
           until @done
             begin
-              # We monitor the PID controller of the target service
-              semian_resource = Semian["#{@resource_name}_#{@target_service.object_id}".to_sym]
-              if semian_resource&.circuit_breaker
-                metrics = semian_resource.circuit_breaker.pid_controller.metrics
+              # Collect metrics from all thread resources
+              all_metrics = []
+              @num_threads.times do |thread_num|
+                resource_name = "#{@resource_name}_thread_#{thread_num}".to_sym
```
AbdulRahmanAlHamali: you're removing the service from there. We want the service in the name as well

AbdulRahmanAlHamali (changes requested):
I'm not sure you are fully understanding the effect of the changes before making them.

You are removing the service from the resource name and replacing it by the thread ID. We want both to be in there.

Also, regarding increasing the error thresholds for the classical semian, the goal is to find an error threshold that would make it not open when the service is healthy, but open when it's degraded. The graph you shared does not show that effect

experiments/test_helpers.rb
AbdulRahmanAlHamali: fair enough, we don't currently use the sysv semaphores, but we might in the future, so it makes sense to do a cleanup

experiments/sudden_error_spike_20.png
AbdulRahmanAlHamali, line 1: it seems you deleted this file. You can bring it back with

```
git checkout pid-take-2 <file_name>
```

I think that should work

Gemfile.lock
```diff
@@ -28,6 +28,7 @@ PATH
   specs:
     semian (0.26.6)
       concurrent-ruby
+      logger
```
AbdulRahmanAlHamali: accidental changes?

experiments/Gemfile.lock
```diff
@@ -3,6 +3,7 @@ PATH
   specs:
     semian (0.26.6)
       concurrent-ruby
+      logger
```
AbdulRahmanAlHamali: accidental changes?

experiments/test_helpers.rb
```diff
@@ -189,42 +189,84 @@ def start_request_threads
...
 
           total_request_time = @thread_timings.values.sum { |t| t[:samples].sum { |s| s[:duration] } }
 
           @pid_mutex.synchronize do
-            @pid_snapshots << {
-              timestamp: Time.now.to_i,
-              window: @pid_snapshots.length + 1,
-              current_error_rate: payload[:error_rate],
+            timestamp = Time.now.to_i
+            window_key = payload[:window_number] || timestamp
```
AbdulRahmanAlHamali: what is window_number? 🤔

semian.gemspec
```diff
@@ -28,6 +28,7 @@ Gem::Specification.new do |s|
   }
 
   s.add_runtime_dependency("concurrent-ruby")
+  s.add_runtime_dependency("logger")
```
AbdulRahmanAlHamali: Let's not add extra dependencies to semian.

Not sure why you needed that, maybe you used it while troubleshooting, but technically you could assign a logger to semian externally.

Also, for troubleshooting while iterating, you could use `puts` which is the easiest option imo

experiments/test_helpers.rb
```diff
@@ -189,42 +189,84 @@ def start_request_threads
...
+            # Accumulate metrics from each thread
+            @window_metrics[window_key][:error_rates] << payload[:error_rate]
+            @window_metrics[window_key][:p_values] << payload[:p_value]
+            @window_metrics[window_key][:previous_p_values] << payload[:previous_p_value]
+            @window_metrics[window_key][:rejection_rates] << payload[:rejection_rate]
+            @window_metrics[window_key][:integrals] << payload[:integral]
+            @window_metrics[window_key][:derivatives] << payload[:derivative]
+
+            # When we've collected metrics from all threads, create the aggregated snapshot
+            if @window_metrics[window_key][:error_rates].length == @num_threads
```
AbdulRahmanAlHamali: so here, there is the implicit assumption that all the semians will report their metric within the same second. 

If one of them doesn't (i.e. is a bit late), you'll end up not adding this entry.

I think you had this on your mind when you thought of the window_number concept. But that would be a much more complex implementation.

So I think a better implementation here would be to not do any aggregation at this level. Just let everyone add their entries, and later, in the analysis phase, you can aggregate everything that has the same timestamp

Gemfile.lock
kris-gaudel:
```bash
/Users/kristofergaudel/src/github.com/Shopify/semian/lib/semian.rb:4: warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add logger to your Gemfile or gemspec to silence this warning.
Creating mock service...
```

This was the warning I was referring to. I added the dep to silence it

kris-gaudel: Turns out I didn't need to add it to the main Semina :)

experiments/test_helpers.rb
kris-gaudel: Blip

AbdulRahmanAlHamali (approved):

PR #847, opened 2025-10-30
Fix integral windup

kris-gaudel:
Changed integral such that we only accumulate when rejection rate is not saturated, and added a line to the `Gemfile` as I was running into errors when trying to run experiments.

Will update with result from `test_lower_bound_windup_adaptive.rb`

### DIFF
--- a/test/pid_controller_test.rb
+++ b/test/pid_controller_test.rb
@@ -105,8 +105,9 @@ class TestPIDController < Minitest::Test
     @controller.update
     # control signal = Ki * integral + Kp * p_value + Kd * (p_value - previous_p_value) / dt
     # p_value = (error_rate - ideal_error_rate) - rejection_rate = (0.11 - 0.01) - 0 = 0.1
-    # With anti-windup, integral behavior changes when output is saturated
-    # The exact value depends on whether previous windows caused saturation
+    # With back-calculation anti-windup: when rejection_rate saturates (hits 0 or 1),
+    # the integral accumulation is reversed: integral -= p_value * dt
+    # This prevents integral windup but means the exact response depends on saturation history
     assert_in_delta(@controller.metrics[:rejection_rate], 0.20, 0.02)
     assert_equal(1004, @controller.metrics[:p90_estimator_state][:observations])
     # ----------------------------------------------------------------


lib/semian/pid_controller.rb
```diff
@@ -82,8 +82,17 @@ def update
...
 
+      # Calculate what the new rejection rate would be
+      new_rejection_rate = @rejection_rate + control_signal
+
       # Update rejection rate (clamped between 0 and 1)
-      @rejection_rate = (@rejection_rate + control_signal).clamp(0.0, 1.0)
+      @rejection_rate = new_rejection_rate.clamp(0.0, 1.0)
+
+      # Anti-windup: back out the integral accumulation if output was saturated
+      if new_rejection_rate != @rejection_rate
```
AbdulRahmanAlHamali, line 95: interesting approach, did you read about this somewhere?

kris-gaudel: https://www.youtube.com/watch?v=H4YlL3rZaNw&t=251s this video gave me the idea to only accumulate if the output won't be saturated

AbdulRahmanAlHamali: I like that, it's simple enough, and better than setting arbitrary limits

AbdulRahmanAlHamali: let me know when this is ready for a full review

test/pid_controller_test.rb
```diff
@@ -105,9 +105,9 @@ def test_update_flow
     @controller.update
     # control signal = Ki * integral + Kp * p_value + Kd * (p_value - previous_p_value) / dt
     # p_value = (error_rate - ideal_error_rate) - rejection_rate = (0.11 - 0.01) - 0 = 0.1
-    # = 0.1 * (-0.01+ 0.1) * 10 + 1 * 0.1 + 0.01 * (0.1 - (-0.01)) / 10 = 0.19011
-    # rejection rate = 0 + 0.19011 = 0.19011
-    assert_in_delta(@controller.metrics[:rejection_rate], 0.19, 0.01)
+    # With anti-windup, integral behavior changes when output is saturated
```
AbdulRahmanAlHamali: could you add a comment with what the new calculation (including the integral anti-windup is?) That would help people looking at the test in the future

test/pid_controller_test.rb
```diff
@@ -211,6 +211,75 @@ def test_ideal_error_rate_never_returns_more_than_10
 
     assert_equal(0.1, @controller.metrics[:ideal_error_rate])
   end
+
+  def test_integral_anti_windup
```
AbdulRahmanAlHamali, line 216: good test!

test/pid_controller_test.rb
```diff
@@ -211,6 +211,75 @@ def test_ideal_error_rate_never_returns_more_than_10
...
+
+    # Now introduce an error spike - controller should respond immediately
+    80.times { @controller.record_request(:success) }
+    20.times { @controller.record_request(:error) }
+    @controller.update
+
+    # Controller should respond to the spike with proportional and derivative terms
+    # With anti-windup, integral is 0 (not negative), so response is stronger
+    # p_value = (0.20 - 0.01) - 0 = 0.19
+    # control_signal = Kp*0.19 + Ki*0*dt + Kd*(0.19 - (-0.01))/dt
```
AbdulRahmanAlHamali, line 278: the integral also has the current P, which is probably why you did not get the exact value you were thinking of

PR #849, opened 2025-10-30
notify on state transitions and on controller updates

Aguasvivas22:
notify on state transitions and on controller updates

### DIFF
--- a/lib/semian/adaptive_circuit_breaker.rb
+++ b/lib/semian/adaptive_circuit_breaker.rb
@@ -23,7 +23,6 @@ module Semian
       @window_size = window_size
       @stopped = false
       @clock = clock || RealClock.new
-      @last_rejection_rate = 0.0
 
       @pid_controller = if thread_safe
         ThreadSafePIDController.new(
@@ -128,19 +127,14 @@ module Semian
 
           @clock.sleep(@window_size)
 
-          old_rejection_rate = @last_rejection_rate
-
-          # Capture metrics BEFORE update resets the window
+          old_rejection_rate = @pid_controller.rejection_rate
           pre_update_metrics = @pid_controller.metrics
 
           @pid_controller.update
           new_rejection_rate = @pid_controller.rejection_rate
 
           check_and_notify_state_transition(old_rejection_rate, new_rejection_rate, pre_update_metrics)
-
           notify_metrics_update
-
-          @last_rejection_rate = new_rejection_rate
         end
       rescue => e
         Semian.logger&.warn("[#{@name}] PID controller update thread error: #{e.message}")
@@ -192,6 +186,7 @@ module Semian
         p_value: metrics[:p_value],
         integral: metrics[:integral],
         derivative: metrics[:derivative],
+        previous_p_value: metrics[:previous_p_value],
       )
     end
   end
--- a/experiments/test_helpers.rb
+++ b/experiments/test_helpers.rb
@@ -107,7 +107,7 @@ module Semian
         @thread_timings = {}
 
         start_request_threads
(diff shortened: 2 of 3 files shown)


lib/semian/adaptive_circuit_breaker.rb
```diff
@@ -127,11 +128,71 @@ def start_pid_controller_update_thread
 
           @clock.sleep(@window_size)
 
+          old_rejection_rate = @last_rejection_rate
```
AbdulRahmanAlHamali: I don't think you need an instance variable for this. You could just put:

```
old_rejection_rate = @pid_controller.rejection_rate
```

lib/semian/adaptive_circuit_breaker.rb
```diff
@@ -127,11 +128,71 @@ def start_pid_controller_update_thread
...
       end
     end
+
+    def check_and_notify_state_transition(old_rate, new_rate, pre_update_metrics)
+      old_state = old_rate == 0.0 ? :closed : :open
+      new_state = new_rate == 0.0 ? :closed : :open
+
+      if old_state != new_state
+        notify_state_transition(new_state)
+        log_state_transition(old_state, new_state, new_rate, pre_update_metrics)
```
AbdulRahmanAlHamali, line 150: hmmm:

1. I think I would want to record PID values every cycle regardless of whether we went through a state transition. Like, if we were already rejecting some requests, but now we're rejecting more, I would want that info to be propagated.
2. I don't think we should be logging here. We should use `Semian.notify`, and then let the user choose what to do: do they want to log? Maybe emit metrics, etxc.

lib/semian/adaptive_circuit_breaker.rb
```diff
@@ -127,11 +128,71 @@ def start_pid_controller_update_thread
...
+
+      str = "[#{self.class.name}] State transition from #{old_state} to #{new_state}."
+      str += " success_count=#{requests[:success]}"
+      str += " error_count=#{requests[:error]}"
+      str += " rejected_count=#{requests[:rejected]}"
+      str += " rejection_rate=#{(rejection_rate * 100).round(2)}%"
+      str += " error_rate=#{(pre_update_metrics[:error_rate] * 100).round(2)}%"
+      str += " ideal_error_rate=#{(pre_update_metrics[:ideal_error_rate] * 100).round(2)}%"
+      str += " integral=#{pre_update_metrics[:integral].round(4)}"
+      str += " name=\"#{@name}\""
```
AbdulRahmanAlHamali, line 170: Instead of `self.class.name`, we should probably just put this `@name` in the beginning. That would be more consistent with the rest of our logging

lib/semian/adaptive_circuit_breaker.rb
AbdulRahmanAlHamali: Oh, I see now that you have a separate call for notify_metric_update. I wonder why we would want to add our own extra log though

Aguasvivas22: I kept notify_state_transition the same as [classic](https://github.com/Shopify/semian/blob/pid-take-2/lib/semian/circuit_breaker.rb#L166) for compatibility sake (in case something is expecting a specific amount of fields) and separated it from the continuous updates since it's not as relevant in ACB. But they can be merged if that's not an issue.

lib/semian/adaptive_circuit_breaker.rb
AbdulRahmanAlHamali: chatted offline: we're keeping this log (and its shape) to keep a similar behaviour with the classic circuit breaker

AbdulRahmanAlHamali (approved):

PR #853, opened 2025-11-03
add vertical reference lines on classic CB state changes and more helper flex

Aguasvivas22:
add vertical reference lines on classic CB state changes and more helper flex


experiments/test_helpers.rb
```diff
@@ -448,6 +476,28 @@ def generate_visualization
...
+        
+        @state_transitions.each_with_index do |transition, idx|
+          # Calculate which bucket this transition falls into
+          elapsed = transition[:timestamp] - test_start
+          bucket_idx = (elapsed / bucket_size).to_i
+          
+          next if bucket_idx < 0 || bucket_idx >= num_buckets
+          
+          # Add vertical reference line at this bucket index
+          color = transition[:state] == :open ? 'red' : 'green'
```
anagayan: Since the classic circuit breaker has a `half_open` state, maybe we should include that point visually as well

Aguasvivas22: done

anagayan: Thanks - does the image on PR change with this implementation?

Aguasvivas22: we were already including half_open on the graph, it's just that it was bundled in the same color as "closed". You have to really zoom in to see it, but it's there.

PR #854, opened 2025-11-03
clamp our rejection rate at 99%

Aguasvivas22:
Clamp rejection rate at 99%.


Aguasvivas22:
abandoning PR in preference to tuning a few PID parameters instead.

PR #855, opened 2025-11-03
Add automated workflow for pid

nirmitparikh8:
fixes [#6619](https://github.com/Shopify/resiliency/issues/6619)

### **Why This Is Important**
This PR establishes automated validation for the adaptive PID controller circuit breaker implementation, ensuring performance regression detection and providing continuous benchmarking against classic circuit breakers across multiple failure scenarios.

### **What We're Accomplishing**
- **Continuous Integration**: Testing of PID controller and classic circuit breaker performance on every PR
- **Performance Validation**: Detection of regressions in circuit breaker behavior
- **Visual Documentation**: Ensuring performance graphs are updated with every commit

#### **1. CI Workflow Setup**
- Created `.github/workflows/automated-experiment-result-checkeryml` 
- Check if latest commit updated all the graphs that indicate true performance of teh circuit breaker. If any images are outdated CI fails and tells the user what command to run.

#### **2. Experiment Infrastructure Refactor**
- **Reorganized experiment structure**: Moved test files to `experiments/experiments/` directory and results into `experiments/results/` directory
- **Parallel test execution**: Created `run_all_experiments.rb` with threaded execution (14 concurrent tests)
- Automated graph generation (requests, duration, throughput)

#### **3. Test Coverage** 
**14 test scenarios** comparing adaptive vs classic circuit breakers:
- Sustained load (20% error rate)
- Error spikes (20% and 100% rates)  
- Gradual error increases
- Oscillating error patterns
- Multiple service degradation scenarios

**NOTE:** We skip the lower_bound_adaptive test for the PID controller because it takes an hour. We will create a ticket to make it shorter or more optimized and then include it.

## **Future Steps**
- Block merging to Semian main branch if all CI checks don't pass.
- Set something up where we can comment on the PR to generate updated graphs and commit back to the PR

This new check will automatically help catch any regressions introduced by code changes.


experiments/tests/test_oscillating_errors_adaptive.rb
```diff
@@ -3,9 +3,9 @@
 $LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
```
nirmitparikh8: ../

experiments/results/protected_service/throughput-one_of_many_services_latency_degradation.png
nirmitparikh8, line 1: Are these throughput graphs needed for all tests? Like do they really tell us anything significant?

experiments/results/protected_service/duration-one_of_many_services_latency_degradation.png
nirmitparikh8, line 1: Are these duration graphs needed for all tests? Like do they really tell us anything significant?

experiments/experiments/experiment_lower_bound_windup_adaptive.rb
nirmitparikh8, line 1: Why does it say it is a new file? I moved it and made some changes. Those changes are evident in my last commit.

.github/workflows/pid_controller_automated_test.yml
```diff
@@ -0,0 +1,57 @@
+---
+name: "PID Controller Automated Test"
```
AbdulRahmanAlHamali: let's call this "Automated Experiment Runner" maybe, and drop mention of the PID controller. It could be used for more later

.github/workflows/pid_controller_automated_test.yml
```diff
@@ -0,0 +1,57 @@
...
+
+# yamllint disable-line rule:truthy
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+concurrency:
+  group: ${{ github.ref }}-pid-test
```
AbdulRahmanAlHamali: what does this mean? 🤔

.github/workflows/pid_controller_automated_test.yml
```diff
@@ -0,0 +1,57 @@
...
+  cancel-in-progress: true
+
+permissions:
+  contents: write
+  pull-requests: write
+jobs:
+  pid-controller-automated-test:
```
AbdulRahmanAlHamali: same notes as above

.github/workflows/automated-experiment-runner.yml
```diff
@@ -0,0 +1,57 @@
...
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+      - uses: ruby/setup-ruby@0481980f17b760ef6bca5e8c55809102a0af1e5a # v1.263.0
+        with:
+          bundler-cache: true
+
+      - name: Setup Environment
+        run: |
+          git add .
```
AbdulRahmanAlHamali: what's the purpose of those two lines? 🤔

.github/workflows/pid_controller_automated_test.yml
```diff
@@ -0,0 +1,57 @@
...
+          git reset --hard
+          bundle install
+          bundle exec rake compile
+          cd experiments
+          bundle install
+
+      - name: Run Experiments
+        run: |
+          cd experiments
+          bundle exec ruby run_all_tests.rb
```
AbdulRahmanAlHamali: since we're already refactoring. I've been meaning to call these experiments instead of tests. So could you name this `run_all_experiments`?

.github/workflows/automated-experiment-runner.yml
```diff
@@ -0,0 +1,57 @@
...
+          bundle install
+
+      - name: Run Experiments
+        run: |
+          cd experiments
+          bundle exec ruby run_all_tests.rb
+
+      - name: Commit Graphs
+        run: |
+          git checkout ${{ github.head_ref }}
```
AbdulRahmanAlHamali: I thought you were already checked out 🤔

experiments/example/example_with_circuit_breaker.rb
```diff
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
-$LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
+$LOAD_PATH.unshift(File.expand_path("../../lib", __dir__))
```
AbdulRahmanAlHamali: let's just delete the example. It's not up-to-date with how we're doing most things now, and we do have a ton of real tests instead of the example

experiments/run_all_experiments.rb
AbdulRahmanAlHamali, line 1: as noted above, let's replace all usages of test with experiment

experiments/run_all_tests.rb
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require "thread"
+
+# Get all test files (excluding the windup test)
+test_files = Dir.glob("tests/*.rb").reject { |file| file.include?("test_lower_bound_windup_adaptive.rb") }.sort
```
AbdulRahmanAlHamali: why are you rejecting this specific test?

experiments/run_all_experiments.rb
```diff
@@ -0,0 +1,90 @@
...
+    rescue => e
+      error = e.message
+      exit_status = 1
+    end
+
+    end_time = Time.now
+    duration = end_time - start_time
+
+    # Thread-safe result storage
+    mutex.synchronize do
```
AbdulRahmanAlHamali, line 42: you don't need the mutex. Every thread is writing to a different key

experiments/run_all_experiments.rb
```diff
@@ -0,0 +1,90 @@
...
+# Summary
+total_duration = results.values.map { |r| r[:duration] }.sum
+successful = results.select { |_, r| r[:exit_status] == 0 }
+failed = results.select { |_, r| r[:exit_status] != 0 }
+
+puts "\nSUMMARY:"
+puts "  Total files: #{results.length}"
+puts "  Successful: #{successful.length}"
+puts "  Failed: #{failed.length}"
+puts "  Total execution time: #{total_duration.round(2)}s"
```
AbdulRahmanAlHamali, line 72: this would appear as hours even if the execution time was 10 minutes. Just because we're running things in parallel, and you're summing up the times.

Is that what you wanted to report?

.github/workflows/pid_controller_automated_test.yml
nirmitparikh8: Creates a concurrency group for the branch with the suffix of the group id being `pid-test`. This allows us to cancel this check if there is a newer commit and we have to run the check on that. if we don't have this, the check will continue running for outdated commits when it doesn't need to.

.github/workflows/automated-experiment-runner.yml
nirmitparikh8: This was to clean the environment from previous failed checks. It just ensures the environment is completely clean. We need this to be able to successfully checkout new branch.

.github/workflows/automated-experiment-runner.yml
nirmitparikh8: We checkout a specific commit to run the experiments on once graphs are generated we need to checkout to the branch to push the graphs to it.

experiments/run_all_tests.rb
nirmitparikh8: This test on its own takes > 1hr. That would take too long for CI to run especially for every commit. it will also block developers when they are trying to merge. I don't think we should block on this test especially when we have 14 other tests to compare against.

experiments/run_all_experiments.rb
nirmitparikh8: Generally it is safe to just write to different keys in a hash without a mutex. However to be safer because different ruby compilers behave differently, mutex is recommended. This avoids corrupting the hash at all costs.

experiments/run_all_experiments.rb
nirmitparikh8: Yeah I wanted to report how long all of our tests would've taken if run sequentially. I can remove the metric if we would like.

PR #857, opened 2025-11-04
Introduce a slow query experiment

AbdulRahmanAlHamali:
Fixes https://github.com/Shopify/resiliency/issues/6620

This PR does a bunch of things:

1. Introduce the concept of `max_threads` into the mock service, which basically causes it to reject requests if it's already process `max_threads` requests
2. Allow specific endpoint degradation: this is useful to allow turning one endpoint into a slow query.
3. Refactor a bit how requests are issued: Currently, the client sleeps specific amounts between making requests, which makes it hard to predict how much traffic the client is going to generate. Instead, I added a thread that adds requests to a queue, and then the clients pull from that queue. That makes it much closer to a real web server implementation
4. Finally, I introduced a slow query test: Where one endpoint on a single service becomes very slow.

Regarding the experiment: The results are still a bit confusing. But the classic circuit breaker is clearly winning. I still need to analyze the behaviour to understand why


experiments/mock_service.rb
```diff
@@ -182,31 +211,8 @@ def base_latency(endpoint_index)
         @endpoint_latencies[endpoint_index]
       end
 
-      # Check if a specific endpoint would timeout with current settings
```
AbdulRahmanAlHamali, line 185: these three function calls were not called anywhere. Just AI fluff. Removed them

experiments/experiment_helpers.rb
```diff
@@ -69,18 +71,23 @@ def run
...
-            max_latency: 0.3,
+            endpoints_count: 10,
+            min_latency: 0.1,
+            max_latency: 10,
             distribution: {
               type: :log_normal,
-              mean: 0.15,
-              std_dev: 0.05,
+              mean: mean_latency,
+              std_dev: 3,
```
AbdulRahmanAlHamali, line 93: the standard deviation was too low, and all endpoints were ending up pretty much having the same latency

experiments/experiment_helpers.rb
```diff
@@ -116,25 +123,33 @@ def start_threads
...
+        start_processing_threads
         subscribe_to_metrics if @is_adaptive
         subscribe_to_state_changes unless @is_adaptive
       end
 
-      def start_request_threads
-        total_rps = @num_threads * @requests_per_second_per_thread
-        puts "Starting #{@num_threads} concurrent request threads (#{@requests_per_second_per_thread} requests/second each = #{total_rps} rps total)..."
-        puts "Each thread will have its own adapter instance connected to the shared service...\n"
+      def start_request_arrival
```
AbdulRahmanAlHamali, line 139: In a real service, requests are received externally, and processed by the workers. This implementation is much closer to this reality, and it allows us better to control how many requests we are seeing

experiments/experiment_helpers.rb
```diff
@@ -76,18 +78,23 @@ def run
...
         @services = []
+        mean_latency = 0.15
+        # Provide barely enough threads for us to handle the expected load (using Little's law)
+        max_threads_per_service = ((@requests_per_second / @service_count) * (0.75 * mean_latency)).to_i
         @service_count.times do
           @services << MockService.new(
-            endpoints_count: 50,
-            min_latency: 0.01,
-            max_latency: 0.3,
+            endpoints_count: 10,
```
AbdulRahmanAlHamali, line 87: I changed the endpoint count to make it more likely for a degraded endpoint to actually be hit

AbdulRahmanAlHamali:
Note: the behaviour of a lot of the classical circuit breaker experiments looks different, mainly because we haven't run them in a while, especially since @kris-gaudel 's refactor that broke them down per thread

Aguasvivas22 (approved):
LGTM

experiments/experiment_helpers.rb
Aguasvivas22: Would it make sense to parameterize these instead?

Issue #834, opened 2025-10-27
Implement exponential backoff to replace error_timeout


AbdulRahmanAlHamali:
Currently, `error_timeout` decides how long the circuit should open before it transfers to half-open. It also serves as the `error_threshold_timeout` in case the latter was not defined.

There are two problems here:

1. It is far from great that `error_timeout` is acting as a fallback for lack of `error_threshold_timeout`. They serve extremely different purposes...
2. It is pretty arbitrary to decide to open the circuit for 10 seconds vs. 60 seconds.

This issue focuses on part 2, but I also believe that more adoption of this feature will help teams migrate towards getting rid of 1 faster.

-----

One replacement of `error_timeout` is just to have a default behaviour that implements an exponential backoff. The old behaviour is as follows:

```
Circuit receives `error_threshold` number of errors --->  transfer to open ---> wait for error_timeout ---> transfer to half open ---> still see errors? ---> transfer to open ---> wait for error_timeout ---> transfer to half open ---> no more errors? ----> transfer to closed
``` 

The new exponential backoff implementation would be:

```
Circuit receives `error_threshold` number of errors --->  transfer to open ---> wait for 500ms ---> transfer to half open ---> still see errors? ---> transfer to open ---> wait for 1s ---> transfer to half open ---> still see errors? --> transfer to open ---> wait for 2s ---> transfer to half open... 
```

This would allow us to detect recovery much faster, and would not require guess work around what this value should be.

Note: there should probably be a maximum for the backoff, 60 seconds or so

PR #861, opened 2025-11-05
Add exponential_backoff_error_timeout feature

AbdulRahmanAlHamali:
fixes https://github.com/Shopify/semian/issues/834, but doesn't fully replace the concept of error_timeout (as the issue suggests).

Instead, the user still provides `error_timeout`, and the feature is just to provide a boolean that allows us to creep towards it. Once we reach it, we go back to the original behaviour.

I also allow the user to specify the starting point, and the exponential factors. Tbh I doubt people would change those values that often, but wanted to keep this option available since this is an open source library

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -16,9 +16,8 @@ jobs:
     runs-on: ubuntu-latest
     if: |
       (github.event.issue.pull_request
-      && !github.event.issue.pull_request.merged_at
-      && contains(github.event.comment.body, 'signed')
-      )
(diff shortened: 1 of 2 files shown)


Aguasvivas22 (approved):
LGTM! Though it would've been great to have the testing setup from pid-take-2 to help visualize the differences with and without the backoff feature in difference scenarios.

AbdulRahmanAlHamali:
Agreed. Maybe we can port the classic circuit breaker experiments that we already have into main

PR #867, opened 2025-11-12
Replace p90 and P2Estimator with Simple Exponential Smoother (SES)

kris-gaudel:
Using the p90 as our ideal error rate is not resilient to 30min incidents. This PR replaces the p90 as calculated by the P2 estimator with the forecast from a SES. 

A small `alpha` is chosen to make reduce the weight on recent events. As our number of observations increases, we become more "confident" in our observations so we reduce `alpha` by half after 15mins, and half again at 30mins (this idea was proposed by @AbdulRahmanAlHamali, exact numbers and reduction facts are to be discussed).

Some things that are still being considered:
- What should is a good cap value for our forecast?
- What is a good initial value?
- It would be nice to verify these changes on SFR, perhaps our p90 was too sensitive and that's why our utilization spiked [here in Adaptive - Test 3](https://docs.google.com/document/d/1Un7SBZegLNgomhkP-MRkJpTWVAOnt_sP5dqbG-pAYRs/edit?tab=t.h91b0q49g33q)


lib/semian/pid_controller.rb
```diff
@@ -155,25 +155,15 @@ def calculate_window_error_rate
...
-      [p90_value, 0.1].min
-    end
-
-    # Prefill the P2 estimator with observations using initial_error_rate
-    def prefill_p90_estimator
-      initial_history_size = Integer(@initial_history_duration / @window_size)
-      initial_history_size.times do
-        @p90_estimator.add_observation(@initial_error_rate)
-      end
+      [smoothed_value, 0.1].min
```
AbdulRahmanAlHamali: the smoother has a built-in cap. Maybe we don't need this?

lib/semian/simple_exponential_smoother.rb
```diff
@@ -0,0 +1,131 @@
...
+    end
+
+    private
+
+    # Validate that alpha is in the valid range (0, 1]
+    #
+    # @param alpha [Float] The alpha value to validate
+    # @raise [ArgumentError] If alpha is not in valid range
+    def validate_alpha!(alpha)
+      if alpha <= 0 || alpha > 1
```
AbdulRahmanAlHamali: if alpha is >= 0.5 this would be bad, because you would then switch to giving the most recent observations more bias

lib/semian/simple_exponential_smoother.rb
```diff
@@ -0,0 +1,131 @@
...
+    #   smoothed = alpha * capped_value + (1 - alpha) * previous_smoothed
+    #
+    # @param value [Float] The new observation to incorporate
+    # @return [Float] The updated smoothed value
+    def add_observation(value)
+      # We want to lower alpha as our number of observations (confidence) increases
+      @observation_count += 1
+
+      # On startup, we have a higher alpha to place more weight on first influx of observations
+      if @observation_count >= 90 # 90 observations = 15 minutes
```
AbdulRahmanAlHamali: don't you mean `==`? we don't want to decrease the alpha continuously

lib/semian/simple_exponential_smoother.rb
```diff
@@ -0,0 +1,131 @@
...
+  # - Gradually adapts to sustained changes in error rate
+  # - Clips extreme values to prevent outliers from distorting the forecast
+  #
+  # Example usage:
+  #   smoother = SimpleExponentialSmoother.new
+  #   smoother.add_observation(0.01)
+  #   baseline = smoother.forecast  # Returns smoothed baseline error rate
+  #
+  class SimpleExponentialSmoother
+    # Default smoothing factor (default is 1%, but as our "confidence" increases we decrease the alpha)
```
AbdulRahmanAlHamali: if I understand the code correctly, your plan is to eventually decrease it to 0.0001, so the comment is incorrect

lib/semian/simple_exponential_smoother.rb
```diff
@@ -0,0 +1,131 @@
...
+    #
+    # @return [Float] The current smoothed value
+    def value
+      @smoothed_value
+    end
+
+    # Get the current state for monitoring/debugging
+    #
+    # @return [Hash] Current state including smoothed value and configuration
+    def state
```
AbdulRahmanAlHamali, line 79: let's also report the observation count

test/pid_controller_test.rb
```diff
@@ -70,8 +66,8 @@ def test_update_flow
     @controller.update
 
     assert_equal(0, @controller.metrics[:rejection_rate])
-    # update should add a new observation to the p90 estimator
-    assert_equal(1001, @controller.metrics[:p90_estimator_state][:observations])
+    # update should update the smoother with new observation
```
AbdulRahmanAlHamali: I think you changed those because you initially did not have an observation count. But now you do, so maybe they should keep looking at that?

They're basically asserting that we did record an observation

test/pid_controller_test.rb
```diff
@@ -26,27 +26,23 @@ def teardown
...
-          observations: 1000,
-          markers: [0.01] * 5,
-          # The positions are of P0, P45, P90, P95, P100 of 1000 observations
-          positions: [0, 449, 899, 949, 999],
-          quantile: 0.9,
-        },
-      },
-      metrics,
-    )
+    assert_equal(0.0, metrics[:rejection_rate])
```
AbdulRahmanAlHamali: what's the reason we're switching to field-by-field comparison?

I liked the original strict equality because it would help tell us if we're forgetting to test some fields

test/pid_controller_test.rb
```diff
@@ -180,37 +178,51 @@ def test_reset_clears_all_state
...
-          observations: 1000,
-          markers: [0.01] * 5,
-          # The positions are of P0, P45, P90, P95, P100 of 1000 observations
-          positions: [0, 449, 899, 949, 999],
-          quantile: 0.9,
-        },
-      },
-      @controller.metrics,
-    )
+    metrics = @controller.metrics
```
AbdulRahmanAlHamali: same question: why are we changing this?

test/simple_exponential_smoother_test.rb
```diff
@@ -0,0 +1,170 @@
...
+    assert_operator(smoother.forecast, :>, 0.075)
+  end
+
+  # Critical test: 30-minute incident resilience
+
+  def test_effect_of_30_min_incident_on_ideal_error_rate
+    smoother = Semian::SimpleExponentialSmoother.new
+    baseline = smoother.forecast
+
+    # Simulate 30-minute incident (180 observations at 1 obs/min = 30 min)
```
AbdulRahmanAlHamali:
```suggestion
    # Simulate 30-minute incident (180 observations at 1 obs/10 seconds = 30 min)
```

test/simple_exponential_smoother_test.rb
```diff
@@ -0,0 +1,170 @@
...
+    # High error rate during incident
+    180.times { smoother.add_observation(0.2) }
+
+    # We should be resilient to the incident and return close to baseline
```
AbdulRahmanAlHamali:
```suggestion
    # We should be resilient to the incident and our value should not change by much
```

test/simple_exponential_smoother_test.rb
```diff
@@ -0,0 +1,170 @@
...
+  def test_handles_edge_case_values
+    smoother = Semian::SimpleExponentialSmoother.new
+
+    # Zero observation
+    smoother.add_observation(0.0)
+
+    assert_operator(smoother.forecast, :<, 0.01)
+
+    # Negative observation (shouldn't crash)
+    smoother.add_observation(-0.1)
```
AbdulRahmanAlHamali: we should reject negative observations

test/simple_exponential_smoother_test.rb
AbdulRahmanAlHamali, line 1: We should add a test that confirms that the value of alpha increases after 90 observations and then after 30

kris-gaudel: Do you mean decrease?

lib/semian/simple_exponential_smoother.rb
```diff
@@ -0,0 +1,85 @@
...
+
+      @observation_count += 1
+
+      if @observation_count == 90
+        @alpha *= 0.5
+      elsif @observation_count == 180
+        @alpha *= 0.5
+      end
+
+      return @smoothed_value if value > cap_value
```
AbdulRahmanAlHamali, line 54:
```suggestion
     # ignore any observation higher than the cap_value. Anomalous error rates should not affect the ideal error rate calculation
      return @smoothed_value if value > cap_value
```

lib/semian/simple_exponential_smoother.rb
```diff
@@ -0,0 +1,85 @@
...
+      @smoothed_value = initial_value
+      @observation_count = 0
+      @alpha = initial_alpha
+      self
+    end
+
+    private
+
+    def validate_alpha!(alpha)
+      if alpha <= 0 || alpha >= 0.5
```
AbdulRahmanAlHamali:
```suggestion
      # alpha should always be less than 0.5. The goal is to provide less weight for the more recent observations
      if alpha <= 0 || alpha >= 0.5
```

lib/semian/simple_exponential_smoother.rb
```diff
@@ -0,0 +1,127 @@
...
+      @alpha = initial_alpha
+      @cap_value = cap_value
+      @initial_value = initial_value
+      @smoothed_value = initial_value
+      @observation_count = 0
+      @consecutive_deviations = 0
+      @last_deviation_direction = nil
+    end
+
+    def add_observation(value)
```
AbdulRahmanAlHamali, line 51: this method needs a lot of comments. I have no clue what is happening here 😓

test/pid_controller_test.rb
```diff
@@ -92,53 +91,55 @@ def test_update_flow
     @controller.update
 
     assert_equal(0, @controller.metrics[:rejection_rate])
-    assert_equal(1003, @controller.metrics[:p90_estimator_state][:observations])
+    assert_operator(@controller.metrics[:smoother_state][:smoothed_value], :<, 0.01)
+    assert_equal(3, @controller.metrics[:smoother_state][:observation_count])
     # ------------------------------------------------------------
     # Error rate now goes above the ideal error rate. Rejection rate should increase.
-    (1..89).each do |_|
+    (1..92).each do |_|
```
AbdulRahmanAlHamali, line 98: I think the goal here was to get round numbers in the P equation, but eh, I don't mind changing it if there is a good reason

test/pid_controller_test.rb
```diff
@@ -184,33 +185,48 @@ def test_reset_clears_all_state
...
       @controller.metrics,
     )
   end
 
-  def test_ideal_error_rate_never_returns_more_than_10
-    estimator = @controller.instance_variable_get(:@p90_estimator)
-    estimator.reset
-    (1..1000).each do |_|
-      estimator.add_observation(0.5)
+  def test_ideal_error_rate_converges_to_cap
```
AbdulRahmanAlHamali: the goal of the test was to confirm we never return more than 10%, not that we converge towards it. I think you need to add a bunch of very high observations, and confirm that we only go as high as 10%

AbdulRahmanAlHamali: actually, as a matter of fact, we won't even increase with these observations, since we drop them.

So maybe we just delete this test? I don't know what is the value of testing that we do converge towards 10%

test/simple_exponential_smoother_test.rb
```diff
@@ -0,0 +1,298 @@
...
+    assert_raises(ArgumentError) { Semian::SimpleExponentialSmoother.new(initial_alpha: -0.1) }
+    assert_raises(ArgumentError) { Semian::SimpleExponentialSmoother.new(initial_alpha: 0.5) }
+    assert_raises(ArgumentError) { Semian::SimpleExponentialSmoother.new(initial_alpha: 1.5) }
+
+    smoother = Semian::SimpleExponentialSmoother.new(initial_alpha: 0.49)
+
+    assert_equal(0.49, smoother.alpha)
+  end
+
+  def test_smoothing_formula
```
AbdulRahmanAlHamali, line 42: this test could use comments explaining the formula, or a better name (or both). I'm not sure what expectations it is testing

test/simple_exponential_smoother_test.rb
```diff
@@ -0,0 +1,298 @@
...
+
+  def test_add_observation_updates_smoothed_value
+    initial_forecast = @smoother.forecast
+    @smoother.add_observation(0.05)
+
+    assert_operator(@smoother.forecast, :>, initial_forecast)
+    assert_operator(@smoother.forecast, :<, 0.05)
+  end
+
+  def test_forecast_is_stable_and_value_is_alias
```
AbdulRahmanAlHamali: let's remove this test. We'll focus on the success criteria tests instead

test/forecast_success_criteria_test.rb
```diff
@@ -0,0 +1,168 @@
...
+
+    assert_in_delta(
+      initial_value,
+      smoother.forecast,
+      DELTA_TOLERANCE,
+      "IER should stay at initial value during low confidence incident above cap",
+    )
+  end
+
+  def test_low_confidence_ier_lower_than_actual_converges_in_30_minutes
```
AbdulRahmanAlHamali:
```suggestion
  def test_low_confidence_ier_lower_than_actual_converges_in_30_minutes_for_2x_error_rate
```

test/forecast_success_criteria_test.rb
```diff
@@ -0,0 +1,168 @@
...
+    smoother = Semian::SimpleExponentialSmoother.new
+    current_ier = smoother.forecast
+    target_error_rate = current_ier * 2
+
+    observations = minutes_to_observations(30)
```
AbdulRahmanAlHamali: could you split this? It would be good to confirm that we don't converge too quickly. So maybe split it into 5-mins interval and confirm that we only converge eventually

test/forecast_success_criteria_test.rb
```diff
@@ -0,0 +1,168 @@
...
+  def test_high_confidence_ier_lower_than_actual_converges_in_1_hour
+    smoother = Semian::SimpleExponentialSmoother.new
+    baseline_error_rate = 0.05
+
+    establish_high_confidence(smoother, baseline_error_rate)
+
+    current_ier = smoother.forecast
+    target_error_rate = current_ier * 2
+
+    observations = minutes_to_observations(60)
```
AbdulRahmanAlHamali: same comment as I had on the low_confidence test

test/simple_exponential_smoother_test.rb
```diff
@@ -0,0 +1,149 @@
...
+    20.times { smoother.add_observation(1.0) }
+
+    assert_equal(initial_forecast, smoother.forecast)
+
+    20.times { smoother.add_observation(0.05) }
+
+    assert_operator(smoother.forecast, :>, initial_forecast)
+  end
+
+  def test_effect_of_30_min_incident_on_ideal_error_rate
```
AbdulRahmanAlHamali: let's remove this. Replaced by the success criteria tests

experiments/Gemfile
```diff
@@ -7,3 +7,4 @@ gemspec path: ".."
 # Graphing library for visualization
 gem "gruff", "~> 0.23"
 gem "logger"
+gem "csv"
```
AbdulRahmanAlHamali, line 10: what's this?

lib/semian/simple_exponential_smoother.rb
```diff
@@ -0,0 +1,97 @@
...
+  #   smoothed = alpha * value + (1 - alpha) * previous_smoothed
+  #
+  # Key characteristics:
+  # - Adaptive alpha adjusts based on confidence period and convergence direction
+  # - Converges faster when error rate is decreasing (conservative)
+  # - Converges slower when error rate is increasing (cautious)
+  # - Drops extreme values above cap to prevent outliers from distorting the forecast
+  #
+  class SimpleExponentialSmoother
+    DEFAULT_ALPHA = 0.001
```
AbdulRahmanAlHamali: we never really use the default alpha or the inital alpha. We are always setting it to one of the other values. So let's remove it?

test/forecast_success_criteria_test.rb
```diff
@@ -0,0 +1,192 @@
...
+      "IER should stay at initial value during low confidence incident above cap",
+    )
+  end
+
+  def test_low_confidence_ier_lower_than_actual_converges_in_30_minutes_for_2x_error_rate
+    smoother = Semian::SimpleExponentialSmoother.new
+    current_ier = smoother.forecast
+    target_error_rate = current_ier * 2
+
+    [5, 10].each do |minutes|
```
AbdulRahmanAlHamali:
```suggestion
    [5, 10, 15, 20, 25].each do |minutes|
```

test/forecast_success_criteria_test.rb
```diff
@@ -0,0 +1,192 @@
...
+
+      refute_in_delta(
+        target_error_rate,
+        smoother.forecast,
+        DELTA_TOLERANCE,
+        "IER should not have converged to target after #{minutes} minutes during low confidence",
+      )
+    end
+
+    observations = minutes_to_observations(20)
```
AbdulRahmanAlHamali:
```suggestion
    observations = minutes_to_observations(5)
```

test/forecast_success_criteria_test.rb
```diff
@@ -0,0 +1,192 @@
...
+  def test_high_confidence_ier_lower_than_actual_converges_in_1_hour
+    smoother = Semian::SimpleExponentialSmoother.new
+    baseline_error_rate = 0.05
+
+    establish_high_confidence(smoother, baseline_error_rate)
+
+    current_ier = smoother.forecast
+    target_error_rate = current_ier * 2
+
+    [10, 20].each do |minutes|
```
AbdulRahmanAlHamali:
```suggestion
    [10, 20, 30, 40, 50].each do |minutes|
```

test/forecast_success_criteria_test.rb
```diff
@@ -0,0 +1,192 @@
...
+
+      refute_in_delta(
+        target_error_rate,
+        smoother.forecast,
+        DELTA_TOLERANCE,
+        "IER should not have converged to target after #{minutes} minutes during high confidence",
+      )
+    end
+
+    observations = minutes_to_observations(40)
```
AbdulRahmanAlHamali:
```suggestion
    observations = minutes_to_observations(10)
```

test/pid_controller_test.rb
```diff
@@ -92,53 +92,55 @@ def test_update_flow
...
     end
     @controller.update
-    # control signal = Ki * integral + Kp * p_value + Kd * (p_value - previous_p_value) / dt
-    # p_value = (error_rate - ideal_error_rate) - rejection_rate = (0.11 - 0.01) - 0.19011 = -0.09011
-    # = 0.1 * (-0.01+ 0.1 -0.09011) * 10 + 1 * -0.09011 + 0.01 * (-0.09011 - (0.19011)) / 10 = -0.09050022
-    # rejection rate = 0.19011 - 0.09050022 = 0.09960978
-    assert_in_delta(@controller.metrics[:rejection_rate], 0.11, 0.02)
-    assert_equal(1005, @controller.metrics[:p90_estimator_state][:observations])
+    # Rejection rate should stabilize around the 8% error rate
+    assert_in_delta(@controller.metrics[:rejection_rate], 0.08, 0.04)
```
AbdulRahmanAlHamali, line 125: there is a reason we are putting the equations in the comments. They help us know whether the algorithm is actually acting as expected.

We should not remove them. But I'm OK with this for now because I'm removing the I and D in a later PR

test/pid_controller_test.rb
```diff
@@ -184,33 +186,38 @@ def test_reset_clears_all_state
...
       @controller.metrics,
     )
   end
 
-  def test_ideal_error_rate_never_returns_more_than_10
-    estimator = @controller.instance_variable_get(:@p90_estimator)
-    estimator.reset
-    (1..1000).each do |_|
-      estimator.add_observation(0.5)
+  def test_ideal_error_rate_adapts_to_sustained_changes
```
AbdulRahmanAlHamali: let's just remove this test from here. It isn't really testing anything that is not in the other files

PR #869, opened 2025-11-13
Accept `QueryIntent` from Active Record 8.2 in `raw_execute`

nvasilevski:
### Why

Rails [is changing](https://github.com/rails/rails/commit/6901769a7e7332ca44bc9c05a26dd8431a9b890b) the signature of the private `raw_execute` method that we patch. Now it's going to accept only one argument and this argument is an instance of `ActiveRecord::ConnectionAdapters::QueryIntent` so we need to adjust the patch accordingly

### Branch conditions

I decided to branch and define two different versions of the method instead of handling `sql_or_intent` argument as I find it cleaner. The question is what to use as a branching condition. I'm using `ActiveRecord::VERSION::STRING` to get current AR version and compare it with "8.2" which is the version the change is supposed to be released on. Let me know if it's not the conventional way of checking Rails version or you have an entirely different check to branch on.

### How got get `sql` from an intent
First of all we need to call `compile_arel_in_intent` to make sure we get SQL from an arel-only query intent. Then we grab either `raw_sql` or `processed_sql` depending on availability. Calling `compile_arel_in_intent` in a patch doesn't mean the complication will happen twice since `compile_arel_in_intent` [has a check to return early if complication isn't needed.](https://github.com/rails/rails/blob/a9b4f5f638fbfc908dc4e657915278cc2451810c/activerecord/lib/active_record/connection_adapters/abstract/database_statements.rb#L56)

### DIFF
--- a/lib/semian/activerecord_trilogy_adapter.rb
+++ b/lib/semian/activerecord_trilogy_adapter.rb
@@ -80,16 +80,25 @@ module Semian
       super
     end
 
-    def raw_execute(sql, *)
-      if Semian::ActiveRecordTrilogyAdapter.query_allowlisted?(sql)
-        super
-      else
+    if ActiveRecord.version >= Gem::Version.new("8.2.a")
+      def raw_execute(intent)
+        compile_arel_in_intent(intent)
+        intent.processed_sql ||= preprocess_query(intent.raw_sql) if intent.raw_sql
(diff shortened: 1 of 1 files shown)


gmcgibbon:
Looks like I missed this. This change was merged in https://github.com/Shopify/semian/pull/904.

PR #870, opened 2025-11-13
Increase ki

AbdulRahmanAlHamali:


experiments/experiments/experiment_one_of_many_services_degradation.rb
```diff
@@ -9,7 +9,7 @@
 
 runner = Semian::Experiments::CircuitBreakerExperimentRunner.new(
   experiment_name: "One of Many Services Degradation Experiment",
```
AbdulRahmanAlHamali, line 11: several tests were sharing the resource name, which I think leads to conflicting results when running them in parallel.

I fixed them here, but we should probably just have the experiment helper automatically set the resource name

PR #872, opened 2025-11-13
Commit experiment result tables and fix bug in experiment helper

AbdulRahmanAlHamali:
This PR does two things:

1. Write tabular results of experiments, and commit those as well. They contain additional data that is sometimes very hard to see in the graphs
2. There was a bug in the experiment helper that basically caused the processing threads to keep running even when no requests have been added by the arrival thread. This PR fixes that as well, and as a result you will see a change in the graphs where we're currently rejecting much less


experiments/experiment_helpers.rb
```diff
@@ -430,17 +480,24 @@ def display_pid_controller_state
         end
 
         puts "\n=== PID Controller State Per Second (Aggregated across threads) ==="
-        puts format("%-8s %-10s %-20s %-15s %-20s %-20s %-20s %-15s", "Window", "# Threads", "Err % (min-max)", "Ideal Err %", "Reject % (min-max)", "Integral (min-max)", "Derivative (min-max)", "Total Req Time")
-        puts "-" * 150
+        @table_outputs << "=== PID Controller State Per Second (Aggregated across threads) ==="
+
+        header = format("%-8s %-10s %-22s %-15s %-22s %-25s %-25s %-15s", "Window", "# Threads", "Err % (min-max)", "Ideal Err %", "Reject % (min-max)", "Integral (min-max)", "Derivative (min-max)", "Total Req Time")
```
adriangudas, line 472: (nit) I also realized we could probably just use CSV format here - I think it looks fairly reasonable as a github preview, and might be easier to work with.

adriangudas (approved):

Aguasvivas22 (approved):
lgtm

experiments/experiment_helpers.rb
AbdulRahmanAlHamali: good call. I'll switch to csv format

PR #874, opened 2025-11-15
Sliding window implementation for PID controller

AbdulRahmanAlHamali:
Our tests have shown the discrete window implementation to be too janky, with rejection rate dropping too aggressively. Sliding window provides better smoothness.

fixes https://github.com/Shopify/resiliency/issues/6615


lib/semian/timestamped_sliding_window.rb
```diff
@@ -0,0 +1,97 @@
+# frozen_string_literal: true
```
kris-gaudel: Is there a reason `simple_sliding_window.rb` wouldn't work here?

kris-gaudel: Or maybe this could inherit from it

AbdulRahmanAlHamali: hmmm, I'll look at its implementation, but we'll have to use it in a slightly different way: the classic circuit breaker only slides the window when it receives a new response, but we want to slide it continuously every second. So it will probably need some tweaking

lib/semian.rb
```diff
@@ -320,9 +320,10 @@ def create_adaptive_circuit_breaker(name, **options)
       ki: 0.01, # Moderate integral gain
       kd: 0.5, # Small derivative gain (as per design doc)
       window_size: 10, # 10-second window for rate calculation and update interval
+      sliding_interval: 1, # 1-second interval for background health checks
       initial_history_duration: 900, # 15 minutes of initial history for p90 calculation
       initial_error_rate: options[:initial_error_rate] || 0.01, # 1% error rate for initial p90 calculation
-      thread_safe: Semian.thread_safe?,
+      implementation: implementation(**options),
```
AbdulRahmanAlHamali, line 326: Not exactly part of the PR, but just making the implementation more consistent with the rest of the repo, where they define an implementation and it's either "Simple" or "ThreadSafe"

lib/semian/pid_controller.rb
```diff
@@ -4,190 +4,203 @@
...
-
-      # Current window counters
-      @current_window_requests = { success: 0, error: 0, rejected: 0 }
-
-      # Last completed window metrics (used between updates)
-      @last_error_rate = 0.0
-
-      @last_p_value = 0.0
-    end
+  module Simple
```
AbdulRahmanAlHamali, line 7: Tip: set "Hide Whitespaces" to true while reviewing this file

lib/semian/pid_controller.rb
```diff
@@ -4,190 +4,203 @@
...
+        @sliding_interval = sliding_interval
+
+        # Ideal error rate estimation using Simple Exponential Smoother
+        @smoother = SimpleExponentialSmoother.new(
+          initial_alpha: smoother_alpha,
+          cap_value: smoother_cap_value,
+          initial_value: initial_error_rate,
+        )
+
+        @errors = implementation::SlidingWindow.new
```
AbdulRahmanAlHamali: I could've technically put these three into 1 array with entries looking like:

```
{type:, timestamp}
```
But this is more efficient since we later have to scan the array to decide how many errors vs. successes, while this way we could simply look at sizes

lib/semian/pid_controller.rb
```diff
@@ -4,190 +4,203 @@
...
-      # P = (error_rate - ideal_error_rate) - rejection_rate
-      # P increases when: error_rate > ideal
-      # P decreases when: rejection_rate > 0 (feedback mechanism)
-      (current_error_rate - ideal_error_rate) - @rejection_rate
-    end
+      def calculate_error_rate
+        # Clean up old observations
+        current_timestamp = current_time
+        cutoff_time = current_timestamp - @window_size
+        @errors.reject! { |timestamp| timestamp < cutoff_time }
```
AbdulRahmanAlHamali, line 153: this could be implemented more efficiently, since we know entries are ordered by time. So if found to be too slow, we can optimize

lib/semian/simple_sliding_window.rb
```diff
@@ -20,7 +20,7 @@ def reject!(&block)
     end
 
     def push(value)
-      resize_to(@max_size - 1) # make room
+      resize_to(@max_size - 1) unless @max_size.nil? # make room
```
AbdulRahmanAlHamali: I think max_size makes sense for the classic, because we know exactly how many requests we want to store before opening the circuit. For adaptive, we don't know that, so I elected to make them unlimited.

This _should_ be ok, since we don't expect to see too many requests to a single dependency. But we could add an optional max_size later if we see issues

test/pid_controller_test.rb
```diff
@@ -58,101 +59,101 @@ def test_record_requests
...
-    # So the new P value dominates, bringing down the rejection rate quickly.
-    (1..99).each do |_|
-      @controller.record_request(:success)
-    end
-    @controller.record_request(:error)
-    @controller.update
-
-    assert_equal(0, @controller.metrics[:rejection_rate])
-  end
+  # def test_update_flow
```
AbdulRahmanAlHamali, line 61: These tests have to be fixed. However, in the next PR, I'm going to simplify the controller, from a PID controller to a simple P controller as discussed. So I'm not going to invest in this now

lib/semian/adaptive_circuit_breaker.rb
```diff
@@ -134,7 +126,7 @@ def start_pid_controller_update_thread
     end
 
     def wait_for_window
-      Kernel.sleep(@window_size)
+      Kernel.sleep(@sliding_interval)
```
Aguasvivas22, line 129: Shouldn't this still be the window size?

AbdulRahmanAlHamali: I don't think so, if we're sliding every second, we would want to wake the thread up every second, but then look back at the entire window size

lib/semian/simple_sliding_window.rb
anagayan: I agree it might not be as concrete as the classic, but just thinking out loud it would be good to have some limit to prevent unbound growth

AbdulRahmanAlHamali: agreed. Do you have a number in mind? Or should we just put some large value, e.g. 1000 or something?

AbdulRahmanAlHamali: Ideally we should choose a setting that increases/decreases with window size. I'm guessing 100 x window_size. So for window_size = 10, that would be a 1000.

lib/semian/pid_controller.rb
```diff
@@ -4,188 +4,203 @@
...
+        @sliding_interval = sliding_interval
+
+        # Ideal error rate estimation using Simple Exponential Smoother
+        @smoother = SimpleExponentialSmoother.new(
+          cap_value: smoother_cap_value,
+          initial_value: initial_error_rate,
+          observations_per_minute: 60 / sliding_interval,
+        )
+
+        @errors = implementation::SlidingWindow.new(max_size: 100 * window_size)
```
kris-gaudel: What is the meaning of multiplying by 100?

lib/semian/pid_controller.rb
```diff
@@ -4,188 +4,203 @@
...
-      @last_error_rate = calculate_window_error_rate
+        @last_error_rate = calculate_error_rate
 
-      # Store error rate for historical analysis
-      store_error_rate(@last_error_rate)
+        store_error_rate(@last_error_rate)
 
-      # Reset window counters for next window
-      @current_window_requests = { success: 0, error: 0, rejected: 0 }
+        dt = @sliding_interval
```
kris-gaudel, line 72: Note: `@sliding_interval` is only going to be used sleep (`wait_for_window`) if we decide that we are going to use a P controller

lib/semian/pid_controller.rb
```diff
@@ -4,188 +4,203 @@
...
 
-      # P = (error_rate - ideal_error_rate) - rejection_rate
-      # P increases when: error_rate > ideal
-      # P decreases when: rejection_rate > 0 (feedback mechanism)
-      (current_error_rate - ideal_error_rate) - @rejection_rate
-    end
+      def calculate_error_rate
+        # Clean up old observations
+        current_timestamp = current_time
+        cutoff_time = current_timestamp - @window_size
```
kris-gaudel, line 152: I'm confused by this part. We reject observations that are older than the window, but we are sliding the window every second?

lib/semian/pid_controller.rb
kris-gaudel: I'm confused by the sliding interval if its use is limited to this (and updating `dt`)

lib/semian/pid_controller.rb
AbdulRahmanAlHamali: basically I'm allowing up to a 100 requests per second to be stored. So a 1000 requests in a 10-second time window.

It's an arbitrary limit. Just not to keep things unbound

lib/semian/pid_controller.rb
AbdulRahmanAlHamali: > Note: @sliding_interval is only going to be used sleep (wait_for_window) if we decide that we are going to use a P controller

Correct.

> I'm confused by the sliding interval if its use is limited to this (and updating dt)

The sliding interval just means that we will be rerunning the controller every second, and looking back at 10 seconds worth of data

lib/semian/pid_controller.rb
AbdulRahmanAlHamali: yes, so think of it this way: every second, we wake up the controller, the controller looks at the past 10 seconds, and removes anything older than that.

So for example, if I have observations from second 1, 2, 3, 4...10. At second 10, we look at all these observations, at second 11, we look at observations from 2 to 11, at second 12, we look at observations from 3 to 12, etc.

lib/semian/pid_controller.rb
kris-gaudel: Ahh ok that makes sense after piecing it with the example in your other comment 👍

kris-gaudel (approved):

lib/semian/pid_controller.rb
```diff
@@ -4,188 +4,203 @@
...
+        @errors.reject! { |timestamp| timestamp < cutoff_time }
+        @successes.reject! { |timestamp| timestamp < cutoff_time }
+        @rejections.reject! { |timestamp| timestamp < cutoff_time }
 
-    def calculate_window_error_rate
-      total_requests = @current_window_requests[:success] + @current_window_requests[:error]
-      return 0.0 if total_requests == 0
+        total_requests = @successes.size + @errors.size
```
anagayan, line 157: Why doesn't total requests include the rejections? I feel like there was a reason for this, but I'm blanking on it

AbdulRahmanAlHamali: you want to find the error rate in the requests that did actually make it to the backend

PR #875, opened 2025-11-17
Add defensiveness coefficient, and simplify the controller into a P-only controller

AbdulRahmanAlHamali:
A few realizations from our experiments:

1. We should not only attempt to match the error rate. Our goal is to attempt to eliminate the error rate
2. Also, when the error rate drops, our P value should not immediately drop

This PR addresses this by introducing a "defensiveness" coefficient, making the P equation:

```
(error_rate - ideal_error_rate) - 1 / defensiveness * rejection_rate
```

I set the default value of defensiveness to 5, which means that:
* The controller is willing to block 5 times as many requests as the error rate, in attempting to eliminate it (as opposed to only once as many, which we currently do)
* When the error rate is eliminated, the rejection rate will go down gradually as well.

-----
The other thing we noticed is that both the I and the D have little to no contribution to our performance. Thus, I removed them, significantly simplifying the code

fixes https://github.com/Shopify/resiliency/issues/6644 and https://github.com/Shopify/resiliency/issues/6645


Aguasvivas22 (commented):
LGTM, however there might be an issue in PR 874.

The graphs for sudden error spikes and sustained load look off and when I ran the sudden spikes test, I see an output that shows 300 windows vs what should be only 30 windows.

I haven't been able to review the other PR so not sure if that's by design or a bug, so wanted to put this out there.

No issues with this PR though but if it is an error, than we'd want to fix that and regenerate graphs here

lib/semian/proportional_controller.rb
```diff
@@ -5,31 +5,20 @@
...
-    # P = (error_rate - ideal_error_rate) - rejection_rate
+    # P = (error_rate - ideal_error_rate) - (1/defensiveness) * rejection_rate
     # Note: P increases when error_rate increases
     #       P decreases when rejection_rate increases (providing feedback)
-    class PIDController
+    class ProcessController
       attr_reader :name, :rejection_rate
 
-      def initialize(kp:, ki:, kd:, window_size:, sliding_interval:, implementation:, initial_history_duration:,
+      def initialize(defensiveness:, window_size:, sliding_interval:, implementation:,
```
anagayan, line 16: This is basically a coefficient for the rejection_rate. I'd prefer if we just treat it like that directly in the formula (without the `1/`) and illustrate what the range of coefficients will represent. I'm also not a fan of the exact naming of this variable.

As an aside, if we are weighting the rejection_rate differently, is it worth considering doing something similar for the error_rate as well.

lib/semian/pid_controller.rb
```diff
@@ -5,31 +5,20 @@
...
       attr_reader :name, :rejection_rate
 
-      def initialize(kp:, ki:, kd:, window_size:, sliding_interval:, implementation:, initial_history_duration:,
+      def initialize(defensiveness:, window_size:, sliding_interval:, implementation:,
         initial_error_rate:,
         smoother_cap_value: SimpleExponentialSmoother::DEFAULT_CAP_VALUE)
-        # PID coefficients
-        @kp = kp  # Proportional gain
-        @ki = ki  # Integral gain
-        @kd = kd  # Derivative gain
```
anagayan, line 22: I haven't done any experiments or have much basis, but given that the actual targeting equation `P` is changing I would think it might be worthwhile playing around with these values before removing I & D altogether.

If it works fine with just the P and that's good enough, then I'm fine with the simplification too

lib/semian.rb
```diff
@@ -316,14 +316,11 @@ def create_adaptive_circuit_breaker(name, **options)
     # Fixed parameters based on design document recommendations
     AdaptiveCircuitBreaker.new(
       name: name,
-      kp: 0.75, # Standard proportional gain
-      ki: 0.01, # Moderate integral gain
-      kd: 0.5, # Small derivative gain (as per design doc)
       window_size: 10, # 10-second window for rate calculation and update interval
       sliding_interval: 1, # 1-second interval for background health checks
-      initial_history_duration: 900, # 15 minutes of initial history for p90 calculation
```
anagayan, line 324: Switch to SES algorithm made this obsolete. This is a cleanup

lib/semian/pid_controller.rb
```diff
@@ -5,31 +5,20 @@
...
-    class PIDController
+    class ProcessController
       attr_reader :name, :rejection_rate
 
-      def initialize(kp:, ki:, kd:, window_size:, sliding_interval:, implementation:, initial_history_duration:,
+      def initialize(defensiveness:, window_size:, sliding_interval:, implementation:,
         initial_error_rate:,
         smoother_cap_value: SimpleExponentialSmoother::DEFAULT_CAP_VALUE)
-        # PID coefficients
-        @kp = kp  # Proportional gain
```
anagayan, line 20: Discussion detailing that we should keep Kp value here and set it to 1

anagayan, line 20: Let's keep Kp here and set it to 1 (based on discussion)

lib/semian/proportional_controller.rb
AbdulRahmanAlHamali: > As an aside, if we are weighting the rejection_rate differently, is it worth considering doing something similar for the error_rate as well.

So a coefficient for the error rate would basically allow us to block _less_ than the error rate if we want to. Which I guess is a possibility, I'm just not sure if there is value in doing that

AbdulRahmanAlHamali:
Replaced by https://github.com/Shopify/semian/pull/899

PR #887, opened 2025-11-20
Test Near target error rate

Aguasvivas22:
Test Near target error rate

Aside from showing how the classic CB can overreact with a small % of errors, We don't get much out of this test, at least not within the few minutes it's run. However, over a longer period, it can show we're not constantly increasing our rejection rate to correct a small error %


nirmitparikh8 (commented):
lgtm

PR #889, opened 2025-11-21
Add Semian support for Active Record's PostgreSQL Adapter

gmcgibbon:
Adds support for Active Record's PostgreSQL adapter.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,22 +1,22 @@
 GIT
   remote: https://github.com/rails/rails.git
-  revision: 7af1254170c852e92b293ed8ecc4e2fcb3c27a1e
+  revision: 12188144b205bcb7a67adb05d9913ea7a256d54c
   branch: main
   specs:
-    activemodel (8.1.0.beta1)
-      activesupport (= 8.1.0.beta1)
-    activerecord (8.1.0.beta1)
-      activemodel (= 8.1.0.beta1)
-      activesupport (= 8.1.0.beta1)
+    activemodel (8.2.0.alpha)
+      activesupport (= 8.2.0.alpha)
+    activerecord (8.2.0.alpha)
+      activemodel (= 8.2.0.alpha)
+      activesupport (= 8.2.0.alpha)
       timeout (>= 0.4.0)
-    activesupport (8.1.0.beta1)
+    activesupport (8.2.0.alpha)
       base64
-      benchmark (>= 0.3)
       bigdecimal
       concurrent-ruby (~> 1.0, >= 1.3.1)
       connection_pool (>= 2.2.5)
       drb
       i18n (>= 1.6, < 2)
+      json
       logger (>= 1.4.2)
       minitest (>= 5.1)
       securerandom (>= 0.3)
@@ -34,7 +34,6 @@ GEM
   specs:
     ast (2.4.3)
     base64 (0.3.0)
-    benchmark (0.4.1)
     benchmark-ips (2.14.0)
     benchmark-memory (0.2.0)
       memory_profiler (~> 1)
@@ -42,7 +41,7 @@ GEM
     byebug (12.0.0)
     coderay (1.1.3)
     concurrent-ruby (1.3.5)
-    connection_pool (2.5.4)
+    connection_pool (3.0.1)
     date (3.4.1)
     debug (1.11.0)
       irb (~> 1.10)
@@ -91,7 +90,7 @@ GEM
     logger (1.7.0)
     memory_profiler (1.1.0)
     method_source (1.1.0)
-    minitest (5.26.1)
+    minitest (5.26.2)
     mocha (2.8.0)
       ruby2_keywords (>= 0.0.5)
     mutex_m (0.3.0)
@@ -164,7 +163,7 @@ GEM
     ruby2_keywords (0.0.5)
     securerandom (0.4.1)
     stringio (3.1.7)
-    timeout (0.4.3)
+    timeout (0.4.4)
     toxiproxy (2.0.2)
     trilogy (2.9.0)
     tzinfo (2.0.6)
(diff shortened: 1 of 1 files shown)


test/adapters/activerecord_postgresql_adapter_test.rb
```diff
@@ -0,0 +1,438 @@
...
+        end
+
+        assert_nil(error.semian_identifier)
+      end
+
+      def test_resource_acquisition_for_connect
+        @adapter.connect!
+
+        Semian[:postgres_testing].acquire do
+          error = assert_raises(TrilogyAdapter::ResourceBusyError) do
```
adrianna-chang-shopify: These should all be `PostgreSQLAdapter::ResourceBusyError` right? Is it possible to dedup the test coverage for both adapters since we're generally testing the same things?

adrianna-chang-shopify (commented):
It would be nice to extract the commonalities between this new adapter and the Trilogy one to a shared `Semian::ActiveRecordAdapter` module. 

~Also, I think we're going to need to handle `QueryIntent` similar to https://github.com/Shopify/semian/pull/869~ (Sorry, just saw we'll update that separately with Nikita's existing patch)

Separately, there was a convo about retries last week ([ref](https://github.com/shop/issues-rails-infrastructure/issues/1383#issuecomment-3564197713)), it would be great to 1) figure out how Semian should actually behave with retries and 2) add proper test coverage for this.

(EDIT: I think Nikita is picking up the retry stuff separately? In which case my only feedback is to dedup shared logic between the two AR-level adapters if possible :) )

test/adapters/activerecord_postgresql_adapter_test.rb
gmcgibbon: Yes, sorry I expected the test to fail instead of actually checking references. I imagine that means that per-resource errors are just aliases and not subclasses. Either way, I'll update this 👍

lib/semian/activerecord_postgresql_adapter.rb
```diff
@@ -0,0 +1,137 @@
...
+          name = "#{host}:#{port}"
+        end
+        :"postgres_#{name}"
+      end
+      super
+    end
+
+    if ActiveRecord.version >= Gem::Version.new("8.2.a")
+      def execute_intent(intent)
+        return super if Semian::ActiveRecordTrilogyAdapter.query_allowlisted?(intent.processed_sql)
```
adrianna-chang-shopify:
```suggestion
        return super if Semian::ActiveRecordPostgreSQLAdapter.query_allowlisted?(intent.processed_sql)
```

lib/semian/activerecord_postgresql_adapter.rb
```diff
@@ -0,0 +1,137 @@
...
+        super
+      end
+    rescue ResourceBusyError, CircuitOpenError
+      false
+    end
+
+    def with_resource_timeout(temp_timeout)
+      prev_connect_timeout = @config[:connect_timeout]
+      @config.merge!(connect_timeout: temp_timeout) # Create new client with temp_timeout for read timeout
+      yield block
```
adrianna-chang-shopify:
```suggestion
      yield
```

.devcontainer/docker-compose.yml
```diff
@@ -70,3 +71,7 @@ services:
     environment:
       MYSQL_ROOT_PASSWORD: root
       MYSQL_ROOT_HOST: "%"
+
+  postgres:
+    container_name: postgres
+    image: postgres:15
```
adrianna-chang-shopify, line 77: I was having trouble getting this container to start, were you able to? Do we need to configure a username / password?

test/adapters/activerecord_postgresql_adapter_test.rb
```diff
@@ -0,0 +1,438 @@
...
+      def setup
+        super
+        @proxy = Toxiproxy[:semian_test_postgres]
+        Semian.destroy(:postgres_testing)
+
+        @configuration = {
+          adapter: "postgresql",
+          username: "root",
+          password: "root",
+          ssl: true,
```
adrianna-chang-shopify: I don't believe PostgreSQL 15 has SSL configured by default. Do we need to enable it?

test/adapters/activerecord_postgresql_adapter_test.rb
```diff
@@ -0,0 +1,438 @@
...
+        end
+
+        raw_connection = adapter.send(:raw_connection)
+
+        assert_equal(2, raw_connection.read_timeout)
+        assert_equal(2, raw_connection.write_timeout)
+      end
+
+      def test_postgresql_default_read_timeout
+        client = ::Postgresql.new(@configuration.slice(:username, :password, :ssl, :ssl_mode, :host, :port))
```
adrianna-chang-shopify: Should this be `PG`?

test/test_helper.rb
```diff
@@ -36,6 +36,11 @@
     upstream: "#{SemianConfig["mysql_host"]}:#{SemianConfig["mysql_port"]}",
     listen: "#{SemianConfig["toxiproxy_upstream_host"]}:#{SemianConfig["mysql_toxiproxy_port"]}",
   },
+  {
+    name: "semian_test_postgres",
```
eileencodes: This needs to be `semian_test_postgresql` because the code to bring back semian on toxiproxy down matches on the `ActiveRecordConnectionAdapter.db_client` which will return `postgresql` and then this won't match (ie `"semian_test_postgres".match?("postgresql")` will be false).

test/adapters/activerecord_postgresql_adapter_test.rb
gmcgibbon: Right... I'm starting to think that most of this test isn't event being ran. I'll try taking a closer look.

gmcgibbon: It wasn't. The test workflow syntax was off and it wasn't compiling the action. 🤦 Sorry for all that. Now I have tests to work through.

adrianna-chang-shopify (approved):

PR #892, opened 2025-11-25
improve experiement visualization

AbdulRahmanAlHamali:
This PR does a few things:

1. Unify the bucket size so that we can clearly compare apples to apples
2. Unify the "markers" (the white horizontal lines) to allow easy comparison
3. Decrease sizes of stuff to allow better visualization
4. Allow dropping old requests, and decrease the total number of requests generated. These two factors were responsible for very aggressive spikes when the circuit first opened that made it impossible to see other values properly


experiments/experiment_helpers.rb
```diff
@@ -166,6 +166,8 @@ def start_processing_threads
               token = @request_queue.pop
               break if token == :shutdown || @done
 
+              next if token < Time.now.to_i - 1 # Ignore requests that are more than 1 second old
```
kris-gaudel, line 169: Nice

experiments/results/main_graphs/sustained_load_adaptive.png
kris-gaudel, line 1: Removing the dots visualizes the trends better, nice

PR #896, opened 2025-11-27
Fix experiments that were not working

AbdulRahmanAlHamali:
This PR fixes several things that were not great in experiments:

1. The `one_of_many_service_degradation` was accidentally not timing out
2. The slow_query experiments needed some adjustments to actually make the slow query impactful
3. In classic circuit breaker experiments, we have been setting the error rate to 0 (as opposed to 1% for adaptive), which is not apples-to-apples. Instead, I set it back to 0.01, and fixed the error threshold to make sure the circuit breaker opens/closes correctly


experiments/experiment_helpers.rb
```diff
@@ -85,7 +81,7 @@ def setup_services
         @services = []
         mean_latency = 0.15
         # Provide barely enough threads for us to handle the expected load (using Little's law)
-        max_threads_per_service = ((@requests_per_second / @service_count) * (0.5 * mean_latency)).to_i
+        max_threads_per_service = ((@requests_per_second / @service_count) * mean_latency).to_i
```
AbdulRahmanAlHamali, line 84: This is the right Little's Law equation. I had set it at 0.5 before because other parts of the code were buggy, but then we fixed the other parts and didn't fix this one, which broke it. Now it is good

experiments/experiment_helpers.rb
```diff
@@ -633,7 +629,7 @@ def add_state_transition_markers(graph, bucketed_data, bucket_size, num_buckets)
           graph.reference_lines[:"transition_#{idx}"] = {
             index: bucket_idx,
             color: color,
-            width: 2,
+            width: 1,
```
AbdulRahmanAlHamali, line 632: just decreasing the contention of drawing state transitions

experiments/experiment_helpers.rb
```diff
@@ -99,7 +95,7 @@ def setup_services
             error_rate: @base_error_rate,
             timeout: 10,
             max_threads: @with_max_threads ? max_threads_per_service : 0,
-            queue_timeout: 0.0,
+            queue_timeout: 1.0,
```
AbdulRahmanAlHamali, line 98: this means that requests won't be dropped immediately if the service doesn't have capacity, which is more realistic

experiments/experiments/experiment_one_of_many_services_degradation.rb
```diff
@@ -11,13 +11,13 @@
...
   degradation_phases: [Semian::Experiments::DegradationPhase.new(healthy: true)] * 1 +
-                      [Semian::Experiments::DegradationPhase.new(latency: 4.95)] * 10 + # Most requests to the target service will timeout
+                      [Semian::Experiments::DegradationPhase.new(latency: 9.95)] * 10 + # Most requests to the target service will timeout
                       [Semian::Experiments::DegradationPhase.new(healthy: true)] * 3,
   phase_duration: 30,
   service_count: 10,
   semian_config: {
     success_threshold: 2,
-    error_threshold: 10,
+    error_threshold: 2,
```
AbdulRahmanAlHamali, line 20: For this experiment, we have 10 services, so I set the error threshold lower because it will be broken down among different services

experiments/experiments/experiment_one_of_many_services_degradation.rb
```diff
@@ -11,13 +11,13 @@
   experiment_name: "One of Many Services Degradation Experiment",
   resource_name: "protected_service",
   degradation_phases: [Semian::Experiments::DegradationPhase.new(healthy: true)] * 1 +
-                      [Semian::Experiments::DegradationPhase.new(latency: 4.95)] * 10 + # Most requests to the target service will timeout
+                      [Semian::Experiments::DegradationPhase.new(latency: 9.95)] * 10 + # Most requests to the target service will timeout
```
AbdulRahmanAlHamali, line 14: I had increased the timeout to 10 seconds at some point, and forgot to change that

experiments/results/main_graphs/one_of_many_services_latency_degradation.png
AbdulRahmanAlHamali, line 1: this shows clearly the protection pattern of the classic circuit breaker: when it's open, you're 100% protected, when it closes, you tank

experiments/results/main_graphs/one_of_many_services_latency_degradation_adaptive.png
AbdulRahmanAlHamali, line 1: We can clearly see here that when requests do timeout (and thus generate an error), we end up opening the circuit breaker to a certain degree, providing a level of protection

experiments/results/main_graphs/slow_query.png
AbdulRahmanAlHamali, line 1: this was rejecting all the time, even before the slow query was introduced 😅

experiments/results/main_graphs/slow_query_adaptive.png
AbdulRahmanAlHamali, line 1: this was rejecting all the time, even before the slow query was introduced 😅

experiments/results/main_graphs/sudden_error_spike_100_adaptive.png
adriangudas, line 1: wow, this is a noticeable improvement

adriangudas (approved):

PR #897, opened 2025-11-27
Remove unnecessary comments and fix smoother initalization

kris-gaudel:
This PR addresses the following issues:
- There are a lot of variables left over from previous work (e.g., pre-filling with 15min of observations) 
- Variables that are just not used at all
- Comments that are overly verbose and unnecessary
- Limits parameter (e.g., parameters for smoother) setting to `semian.rb`


lib/semian/adaptive_circuit_breaker.rb
```diff
@@ -75,8 +73,7 @@ def closed?
       @pid_controller.rejection_rate == 0
     end
 
-    # half open is not a real concept for an adaptive circuit breaker. But we need to implement it for compatibility with ProtectedResource.
```
adriangudas, line 78: (nit) I like this comment, but it's your call. I feel like it strikes the right balance between being necessary and not overly verbose.

experiments/results/main_graphs/gradual_increase_adaptive.png
adriangudas, line 1: I'm not entirely clear on why the graphs look different if this PR didn't change anything substantial? Or are we not tracking these too closely.

PR #899, opened 2025-11-27
Add Elastic Defensiveness

AbdulRahmanAlHamali:
This PR:

1. Adds an elastic defensiveness coefficient that allows us to block more than the error rate we are seeing, depending on how high that error rate is (the higher it is, the higher we block)
2. Increases Ki, this makes sure we don't drop the rejection rate too quickly when the error rate drops (especially when we're blocking 100%)
3. Changes the integral windup to a simple clamp between -10 and 10. For some reason, the saturation clamping was leading to weird results.


experiments/experiment_helpers.rb
```diff
@@ -454,6 +454,9 @@ def display_pid_controller_state
             integral_avg: metrics.sum { |m| m[:integral] } / metrics.length.to_f,
             integral_min: metrics.map { |m| m[:integral] }.min,
             integral_max: metrics.map { |m| m[:integral] }.max,
+            p_value_avg: metrics.sum { |m| m[:p_value] } / metrics.length.to_f,
```
AbdulRahmanAlHamali, line 457: We were not recording P values in the CSV, added them

experiments/results/main_graphs/one_of_many_services_latency_degradation_adaptive.png
AbdulRahmanAlHamali, line 1: much better protection

lib/semian.rb
```diff
@@ -316,9 +316,9 @@ def create_adaptive_circuit_breaker(name, **options)
     # Fixed parameters based on design document recommendations
     AdaptiveCircuitBreaker.new(
       name: name,
-      kp: 0.75, # Standard proportional gain
-      ki: 0.01, # Moderate integral gain
-      kd: 0.5, # Small derivative gain (as per design doc)
+      kp: 1.0, # Standard proportional gain
+      ki: 0.2, # Moderate integral gain
+      kd: 0.0, # Small derivative gain (as per design doc)
```
AbdulRahmanAlHamali: The Kd seems to have a bad effect in general. When we are blocking 100%, and the error rate drops to 0, P becomes very negative, while the previous P was very positive, it ends up accelerating the drop. Which is good if the incident is actually recovered, but very bad if it isn't.

Setting it to 0 seems to provide better protection

lib/semian/pid_controller.rb
```diff
@@ -87,11 +87,7 @@ def update
         # Update rejection rate (clamped between 0 and 1)
         @rejection_rate = new_rejection_rate.clamp(0.0, 1.0)
 
-        # Anti-windup: back out the integral accumulation if output was saturated
-        if new_rejection_rate != @rejection_rate
-          # Output was clamped, reverse the integral accumulation
-          @integral -= @last_p_value * dt
-        end
+        @integral = @integral.clamp(-10.0, 10.0)
```
AbdulRahmanAlHamali, line 80: I'm not really sure what was wrong with this clamping method, but looking at the integral behaviour with it, it was very weird, and never settling back to 0.

A simple clamp of -10,10 seems to work and make sure it does go back to 0 eventually

lib/semian/pid_controller.rb
```diff
@@ -143,7 +139,8 @@ def calculate_p_value(current_error_rate)
         # P = (error_rate - ideal_error_rate) - rejection_rate
         # P increases when: error_rate > ideal
         # P decreases when: rejection_rate > 0 (feedback mechanism)
-        (current_error_rate - ideal_error_rate) - @rejection_rate
+        delta_error = current_error_rate - ideal_error_rate
+        delta_error - (1 - delta_error) * @rejection_rate
```
AbdulRahmanAlHamali, line 133: This is the key change: an elastic defensiveness that stretches and expands based on how bad the error rate is

experiments/results/main_graphs/gradual_increase.png
anagayan, line 1: Why is the baseline of successes lower on the new one?

experiments/results/main_graphs/near_target_error_rate.png
anagayan, line 1: Why are these two numbers so different in the graphs?

experiments/results/main_graphs/one_of_many_services_latency_degradation.png
anagayan, line 1: Seems like a lot of oscillations. Have we tried larger Ki values to see if it helps here?

lib/semian/pid_controller.rb
```diff
@@ -130,10 +126,11 @@ def metrics
       def calculate_p_value(current_error_rate)
         ideal_error_rate = calculate_ideal_error_rate
 
-        # P = (error_rate - ideal_error_rate) - rejection_rate
+        # P = (error_rate - ideal_error_rate) - (1 - (error_rate - ideal_error_rate)) * rejection_rate
```
anagayan, line 129: Thinking about this more, can we add an error threshold before we start rejecting? As in, we allow a range of errors to happen without engaging rejections. Once it passes that threshold, we start engaging the adaptive controller and rejections.

anagayan (commented):
Edit: Moved comment to code section. Ignore this

experiments/results/main_graphs/gradual_increase.png
Aguasvivas22: pretty sure that's just normal test-to-test variability. I've noticed running all tests in parallel usually decreases the amount of rps for some tests. See adaptive for this test as well.

experiments/results/main_graphs/one_of_many_services_latency_degradation.png
Aguasvivas22: This is for classic.

experiments/results/main_graphs/near_target_error_rate.png
Aguasvivas22: same reasoning as mentioned in my other comment. Basically variability in test runs and other external factors.

I re-ran:

Image

lib/semian/pid_controller.rb
AbdulRahmanAlHamali: I think that's a good idea. Do you have a suggestion on what the threshold should be?

I'm wondering whether it should be an absolute number (e.g. 1%) or something relative to what ideal error rate is.

Another option could be to have finite resolution on our error observation. i.e. applying some rounding up/down so that 1.2% error rate just becomes 1%.

What do you think?

Aguasvivas22: Don't we essentially have that threshold since we initially set the ideal error rate at 5%? 
I do wonder if we should then clamp it to at least 1%

anagayan: I think it should be relative to the ideal error rate. Maybe like 20-25% of it or something similar, completely arbitrary but maybe something we can play around with initially

AbdulRahmanAlHamali: Sounds goood. Just to confirm: would you say your comment is specific for this "elastic defensiveness" implementation? Or more general? If general, we can get this merged and open a separate issue

anagayan: Spoke with @Aguasvivas22 on this and looks like technically we already have it in place with the P formula and subtraction. I think the only thing missing is that we may want to add a floor value so it doesn't go to `0`.

More a general comment, so ignore it on this PR and we can followup if we agree and find it useful

PR #902, opened 2025-12-02
Add option to run a specific example

ali-a-a:
It's currently not possible to run a specific example without manually modifying the `Rakefile`. But it would be helpful to have that option in cases where someone is only interested in running a single example. This PR adds an optional argument to the tasks in the `:examples` namespace to allow users to run a single example by specifying its filename.


ali-a-a:
I have signed the CLA!

PR #1042, opened 2026-05-23
Fix Semian resource allocator pairing

rwstauner:
Ruby headers redefine [strdup to ruby_strdup](https://github.com/ruby/ruby/blob/8e5cf5cecaa59a8409d0ac21ac33d5b16f2cdbbf/include/ruby/util.h#L168-L187), so res->name was allocated with Ruby's allocator even though it looked like libc strdup. Free it with xfree instead of free, and make the allocation explicit with ruby_strdup.

Also release the malloc'd strkey field and add GC regression coverage for native resource cleanup.


jhawthorn (approved):
Looks right to me!

Issue #1045, opened 2026-08-27
`Semian::Adapter`'s re-entrancy guard silently disables the circuit when the adapter is mixed into a shared object


serioushaircut:
## Summary

`Semian::Adapter` implicitly requires that the object it is mixed into is used by exactly one thread at a time — i.e. a per-connection object, like all the adapters that ship with the gem. Nothing in the README says so, and the failure is silent when the requirement is not met.

Mixed into a **shared** object (a process-wide client singleton, a memoized service
object, a pool wrapper), the re-entrancy guard turns the circuit breaker off:

1. While *any* call is in flight anywhere in the process, every concurrent call
   skips `acquire` entirely — no fast-fail on an open circuit, no failure
   accounting, no `:success` / `:circuit_open` notifications.
2. The guard's save/restore is a lost-update race that can strand it at `true`,
   which disables circuit breaking permanently for that process.

Both modes report a healthy, `closed` circuit while every call goes to a dependency
that is timing out. We hit this in a threaded Rails app whose Semian adapter was
mixed into a process-wide client singleton; it took a deliberate concurrency test to
notice, because there is nothing in the logs or the metrics to see.

The trust_battery_client gem does this ⬆️ .

Versions: semian `0.28.2` (also `0.26.6`), CRuby 3.4.10. The code below is unchanged
on `main`.

## The guard

[`lib/semian/adapter.rb#L40-L45`](https://github.com/Shopify/semian/blob/main/lib/semian/adapter.rb#L40-L45):

```ruby
def acquire_semian_resource(scope:, adapter:, &block)
  return yield if resource_already_acquired?

  semian_resource.acquire(scope: scope, adapter: adapter, resource: self) do
    mark_resource_as_acquired(&block)
  end
```

[`#L77-L87`](https://github.com/Shopify/semian/blob/main/lib/semian/adapter.rb#L77-L87):

```ruby
def resource_already_acquired?
  @resource_acquired
end

def mark_resource_as_acquired
  previous = @resource_acquired
  @resource_acquired = true
  yield
ensure
  @resource_acquired = previous
end
```

This is a correct re-entrancy guard for the built-in adapters, because they are
per-connection objects: one `Net::HTTP` / `Mysql2::Client` / `Redis` instance is
only ever used by one thread at a time, so `@resource_acquired` is effectively
per-execution-context, and it only suppresses a genuinely nested acquisition of the
same resource.

On a shared object, the same instance variable is shared by every thread.

[...]

PR #1046, opened 2026-08-27
Document that an adapter instance represents one resource

serioushaircut:
Documents the contract behind #1045: an adapter instance represents one resource, and
access through the same instance is assumed good until that session ends.

Motivation:
Trust Battery is essentially doing it "wrong" - semian adapter is in a process wide singleton for entirety of core, instead of a "session-y" object,  and we run into some weirdness. That spawned this clarification (and I follow up with Trust Battery to fix their use).

Wording follows @matthewd's review. The one genuine race in the same area — the
re-entrancy guard being stranded permanently — is #1047.

Co-authored-by: Claude Opus 5 <noreply@anthropic.com>
Orchestrated-by: ae <noreply@shopify.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,8 @@ To create a Semian adapter you must implement the following methods:
 1. [`include Semian::Adapter`][semian-adapter]. Use the helpers to wrap the
    resource. This takes care of situations such as monitoring, nested resources,
    unsupported platforms, creating the Semian resource if it doesn't already
-   exist and so on.
+   exist and so on. Include it in an object that represents a single resource,
+   typically one connection — see [Thread Safety](#thread-safety).
 2. `#semian_identifier`. This is responsible for returning a symbol that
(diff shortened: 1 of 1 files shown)


serioushaircut:
Status nudge rather than a bump: this one is docs-only (README, no behaviour change) and backs the hazard reported in #1045, so it should be cheap to review.

Two things that may help whoever picks it up:

- The reproduction in #1045 needs no monkey-patching and runs against 0.28.2. It shows an already-open circuit being bypassed by a shared adapter instance, and the guard being stranded at `true` permanently.
- If the preferred outcome is fixing the guard in the gem rather than documenting the requirement, I am happy to close this and send that patch instead. It changes behaviour for every user — circuits that never opened will start opening — which is why I did not lead with it.

<sub>🤖 AE · ✅ approved by @serioushaircut</sub>

Edouard-chin (approved):

PR #1048, opened 2026-09-03
Handle new execute_intent non-raising behaviour

andrewn617:
https://github.com/rails/rails/pull/58049 refactored `execute_intent` to no longer raise errors. Instead execution errors are stored on the query intent and can be checked later by the caller. This is to pave the way for better async execution.

However, semian relies on the raising behaviour to trip the circuit breaker. With this change to Rails, any query that fails mid execution will be considered successful by Semian. And indeed `ActiveRecordTrilogyAdapterTest#test_read_timeout_error_opens_the_circuit` fails when we bump Rails to this commit.

To solve it, in our `execute_intent` patch we can check the intent for an error and raise it within the `acquire_semian_resource` block to trip the circuit breaker. Then we can rescue it and return the result to match the new Rails behaviour.


lib/semian/activerecord_adapter.rb
```diff
@@ -58,8 +58,26 @@ def initialize(*options)
...
 
+        result = nil
+        delivered_error = nil
+
         acquire_semian_resource(adapter: semian_adapter_name, scope: :query) do
-          super
+          result = super
+
+          # In Rails 8.2 execute_intent no longer raises errors. Instead they are stored on the intent
+          # and handled later by the caller to better enable asynrhonous execution. However, we need to raise
```
adrianna-chang-shopify:
```suggestion
          # and handled later by the caller to better enable asynchronous execution. However, we need to raise
```

lib/semian/activerecord_adapter.rb
```diff
@@ -58,8 +58,26 @@ def initialize(*options)
...
+          # any errors here to trip the circuit breaker. Then we can swallow the error and return the result
+          # Rails to continue with.
+          if intent.respond_to?(:error)
+            delivered_error = intent.error
+            raise delivered_error if delivered_error
+          end
+        end
+      rescue => error
+        if error == delivered_error
```
adrianna-chang-shopify:
```suggestion
        if error.equal?(delivered_error)
```
Nit, but this ensures that it's the exact same exception object (vs two exceptions that compare as equal)

adrianna-chang-shopify (approved):

lib/semian/activerecord_adapter.rb
```diff
@@ -58,8 +58,26 @@ def initialize(*options)
...
         acquire_semian_resource(adapter: semian_adapter_name, scope: :query) do
-          super
+          result = super
+
+          # In Rails 8.2 execute_intent no longer raises errors. Instead, they are stored on the intent
+          # and handled later by the caller to better enable asynchronous execution. However, we need to raise
+          # any errors here to trip the circuit breaker. Then we can swallow the error and return the result for
+          # Rails to continue with.
+          if intent.respond_to?(:error)
+            delivered_error.equal?(intent.error)
```
Edouard-chin: That should be on line 77 where Adrianna commented

lib/semian/activerecord_adapter.rb
```diff
@@ -58,8 +58,26 @@ def initialize(*options)
...
+            raise delivered_error if delivered_error
```
Edouard-chin, line 73: I don't fully get were delivered_error gets assigned ?

Edouard-chin (commented):
It would be good to also `rm -rf` the Gemfile.lock before the bundle install steps kicks in on CI. This way we are sure bundler picks the latest Rails edge

lib/semian/activerecord_adapter.rb
andrewn617: Err I changed on the wrong line 😮‍💨 my bad. Fixed now

Edouard-chin (approved):

andrewn617:
Alright, I took a crack at improving the testing situation for this gem. Now the main gemfile and all the testing gemfiles lock to the latest released version, and I added `_edge` versions for each. That way we are running all tests against Rails 8.1 and edge.

I had to keep minitest locked to 5.6. If we bump it to 6.0 we get some test failures, so that can be done in a separate PR.

Also, this uncovered we weren't handling `PG::ConnectionBad` errors that can be raised by 8.1, since edge fixed this behaviour and we were only testing on edge. So I fixed that.

.gitignore
```diff
@@ -7,6 +7,7 @@
 .vscode/ipch
 .vscode/launch.json
 *.gem
+gemfiles/*_edge.gemfile.lock
```
adrianna-chang-shopify: I assumed we'd still check in the lockfile to make it easier for contributors, but we'd remove it in CI to ensure CI always tests against the latest edge?

test/adapters/activerecord_adapter_shared_tests.rb
```diff
@@ -343,7 +343,7 @@ def test_circuit_open_errors_do_not_trigger_the_circuit_breaker
       end
       error = adapter_resource.circuit_breaker.last_error
 
-      assert_instance_of(ActiveRecord::ConnectionNotEstablished, error)
+      assert(connection_failed_error_classes.any? { |klass| error.is_a?(klass) })
```
adrianna-chang-shopify:
```suggestion
      assert_includes(connection_failed_error_classes, error.klass)
```

adrianna-chang-shopify (approved):

Edouard-chin (approved):
@adrianna-chang-shopify Usually committing the *main* Gemfile.lock is enough.

The lockfile in `gemfiles/` are mostly CI only, and it's rare that developers need to run `BUNDLE_GEMFILE=gemfiles/Rails_8.2 bundle exec rake test`. In the event that they have a failing CI for a specific gemfile, then it's better to not have a lockfile committed so that tests run with the same dependencies as on CI (the latest one). 

Committing the Gemfile.lock remove all possible friction to contribute to a project

adrianna-chang-shopify:
>The lockfile in gemfiles/ are mostly CI only

@Edouard-chin , yep fair, in this case though, should we remove all lockfiles under `gemfiles/`? As long as the main lockfile is committed to help contributors, it would be advantageous for the CI-only ones to resolve their latest permitted dependencies right?

We can probably also drop the `rm` from CI then if we're not tracking the edge lockfiles with git.

Edouard-chin:
>  in this case though, should we remove all lockfiles under gemfiles/

@adrianna-chang-shopify Ah yes, great point. I think we should

